PR #136, opened 2015-07-21
Add support for MessageObjects and expose ChatMemberInfo where needed. 

VoiDeD:
Things of interest:

- Changing the return value behavior of `KeyValue.ReadAsBinary` so that it does not have to consume the entire stream in order to be successful.
- `ReadAsBinary` is really busted and reads one child deep compared to valve's implementation. Not sure we can do anything about that, though, as any changes will subtly break users.


SteamKit2/SteamKit2/Types/KeyValue.cs
```diff
@@ -720,7 +756,8 @@ public bool ReadAsBinary( Stream input )
                 this.Children.Add( current );
             }
 
-            return input.Position == input.Length;
+            // it's okay if we haven't reached the end of the input stream
+            return input.Position <= input.Length;
```
Netshroud: Can this expression ever be false?

VoiDeD: I believe some streams can seek past their length, but I'm not sure what kind of behavior we'll get if we attempt to read a truncated KV, for example.

PR #138, opened 2015-07-22
Fix KeyValues binary deserialization

Netshroud:
Now the deserialized element is the root element, not
the root element's single child.

Behavioural breaking change.

### DIFF
--- a/Resources/NetHookAnalyzer2/NetHookAnalyzer2/TreeNodeObjectExplorer.cs
+++ b/Resources/NetHookAnalyzer2/NetHookAnalyzer2/TreeNodeObjectExplorer.cs
@@ -139,8 +139,7 @@ namespace NetHookAnalyzer2
 			}
 			else
 			{
-				var firstChild = kv.Children[0]; // Due to bug in KeyValues parser.
-				SetValueForDisplay(null, childNodes: new[] { new TreeNodeObjectExplorer(firstChild.Name, firstChild).TreeNode });
+				SetValueForDisplay(null, childNodes: new[] { new TreeNodeObjectExplorer(kv.Name, kv).TreeNode });
 			}
 
 			TreeNode.ExpandAll();
--- a/Resources/NetHookAnalyzer2/NetHookAnalyzer2/TreeNodeObjectExplorer.cs
+++ b/Resources/NetHookAnalyzer2/NetHookAnalyzer2/TreeNodeObjectExplorer.cs
@@ -123,14 +123,7 @@ namespace NetHookAnalyzer2
 			bool didRead;
 			using (var ms = new MemoryStream(data))
 			{
-				try
-				{
-					didRead = kv.ReadAsBinary(ms);
-				}
-				catch (InvalidDataException)
-				{
-					didRead = false;
-				}
+				didRead = kv.TryReadAsBinary(ms);
 			}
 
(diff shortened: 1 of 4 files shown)


VoiDeD:
I don't really like breaking consumers, but this fix is definitely welcome.

Anything we can do to provide some warning before we go through with it? Perhaps an `ObsoleteAttribute` on `ReadAsBinary` before we finally ship the change in a later version?

Netshroud:
I don't like the idea of obsoleting, then un-obsoleting it.

I also don't like the options of making `ReadAsBinary2` or `KeyValues2` for compatibility.

We could have a compatibility flag (e.g. `KeyValues.UseBrokenBinaryParsingBehavior`) that preserves the old behaviour that we can switch and later remove, but that just defers the break in compatibility until later.

We could try to read the assembly reference of the consumer (e.g. `SteamKit,Version=1.6.3` or whatever the syntax is), but that only helps if NuGet drops a new SteamKit on top of an older consumer binary, it doesn't help anybody compiling against the newer binary.

I'm stumped for a good solution.

VoiDeD:
>I don't like the idea of obsoleting, then un-obsoleting it.

It's not obsoleting per se, just that it's the only attribute I'm familiar with that can get the compiler to emit any sort of diagnostic message.

Netshroud:
hmm, ok. So what are you suggesting then? Fix it and `[Obsolete]` for the remainder of 1.6.x, then remove the attribute in 1.7.0?

VoiDeD:
That's probably fine, as long as it has a descriptive message, or perhaps links to a wiki article similar to [how we've done some breaking changes in the past](https://github.com/SteamRE/SteamKit/wiki/JobCallback-Transition).

Netshroud:
Rebased on top of master, just to confuse things :smile: 

Changed `ReadAsBinary` to return false in all known errors, instead of throwing `InvalidDataException`. 
If you have a binary KV with a WideString I can take a shot at parsing that too.

Netshroud:
Actually, never mind about WideString: https://github.com/ValveSoftware/source-sdk-2013/blob/master/sp/src/tier1/KeyValues.cpp#L2585-L2589

psychonic:
> Actually, never mind about WideString: https://github.com/ValveSoftware/source-sdk-2013/blob/master/sp/src/tier1/KeyValues.cpp#L2585-L2589

For what it's worth, newer engine versions have this implemented and handle it fine, despite the older TF2 branch not doing so.

aegamesi:
The new ReadAsBinaryCore has incorrect behavior. In the while loop of the method, for each child encountered at the current level, the *current* name and value are continuously overwritten. The end result of this is that for each level of the KeyValue, only the *last* child is properly saved.

Netshroud:
@VoiDeD How about this?

VoiDeD:
:ship:

xPaw:
This silently broke package info in PICS.

We had `productInfo.KeyValues.Children.FirstOrDefault();` in our code to avoid the empty key, however after 1.6.4, this is no longer needed (which is correct), but there's no warning about it. So blindly updating SteamKit will break it.

Netshroud:
Damn, forgot about that one.

I've updated the release notes on GitHub to include this breaking change, but the nupkg is already out in the wild - can't do much about that.

PR #139, opened 2015-07-22
Update enums

xPaw:


VoiDeD:
:+1: unless @Netshroud has any reservations.

Netshroud:
I'm concerned about the introduction of

    YouMustRefundThisTransactionToWallet = 98;
    SendingOfAnEmailFailed = 99;
    PurchaseNotYetSettled = 100;

Those names seem like descriptions of the enum member itself, not the actual name Valve would use. Elordid tries to piece a name together as best as it can from the description, but it isn't always accurate.

As such, as soon as we find out the real names we would have to immediately apply `[Obsolete("renamed to <whatever>")]` to them (unless we get the new names before shipping 1.6.4/1.7).

I'd rather hold off introducing these three until we know their proper names. It's not urgent to include these - judging by the descriptions they appear to be for internal Valve use and probably aren't not exposed to Steamclient/SteamKit.

PR #140, opened 2015-07-23
Simplify Travis CI build process

Netshroud:
Simplified build/test/travis scripts, and removed NuGet.target/.exe/.config/whatever from everything. I can't see why they were there in the first place, but everything works fine without them - VS restores automatically, and Travis restores manually.

I've removed the retry-10-times script entirely in favour of giving Travis a second chance on the network connectivity front (and perhaps in the case of a reoccurrence we should file a bug with Travis this time).

We could probably optimise this further with a Travis container thingo, but the whole build runs in less than 3 minutes right now anyway.


VoiDeD:
:+1:

PR #141, opened 2015-07-27
Expose client_os_type and client_language in SteamUser

xPaw:
Adds `AnonymousLogOnDetails` and additionally sends `client_language` in anonymous logins, which was previously omit.

- Should `LogOnDetails` extend `AnonymousLogOnDetails`?
- Should `machine_id` be exposed too?


VoiDeD:
It's probably alright to expose these. I'm curious what the usecase is for overriding the client's language, however.

>Should machine_id be exposed too?

No, we have no plans to expose this.

PR #142, opened 2015-07-30
Add free license handler

xPaw:
This method works on free on demand packages, and is called by Steam when using `steam://install/` commands for something you don't own. @SteamDatabase is using is successfully.


VoiDeD:
LGTM.

PR #143, opened 2015-07-30
Refactor clientmsg switch statements into a dispatch table.

VoiDeD:


xPaw:
Wouldn't it be better to keep these dispatch tables static?

VoiDeD:
The handler delegates need an instance of `this` to capture, so I don't believe it'd be possible without hand rolling the delegate creation and doing some dynamic invoke which would likely offset whatever micro-optimization gains in the first place.

xPaw:
At least keeping one instance should be better than creating the directory each time HandleMsg is called, don't you think?

Netshroud:
I'd think so too - putting the dispatch table in the constructor seems more sensible than putting it in `HandleMsg`.

Netshroud:
:+1:

PR #152, opened 2015-08-26
Implement machine id support for the future.

VoiDeD:
Any day that requires code changes to logon to Steam is a spooky day, so here's some work to make sure we don't need any changes in the future when valid machine ids are required for logon.

Things to note:
* WMI can be painfully slow. Valve spins up a separate thread at startup that gathers info, ~~we currently query for info at logon time. Not ideal, but this is still a WIP.~~
* For the linux machine id, valve primarily touches `/var/lib/dbus/machine-id`, but this may not be present on some systems (and may have been why certain linux steamclients were unable to logon today). Might need to look into alternatives for anybody not using systemd/dbus.
* ~~Nothing available for OSX. Haven't had much time to dig into the steamclient binary for it, but I'd imagine there aren't many consumers who need SteamKit on OSX.~~

### DIFF
--- a/SteamKit2/SteamKit2/Steam/Handlers/SteamGameServer/SteamGameServer.cs
+++ b/SteamKit2/SteamKit2/Steam/Handlers/SteamGameServer/SteamGameServer.cs
@@ -126,7 +126,7 @@ namespace SteamKit2
 
             logon.Body.client_os_type = ( uint )Utils.GetOSType();
             logon.Body.game_server_app_id = ( int )details.AppID;
-            logon.Body.machine_id = MachineIdUtils.GenerateMachineID();
+            logon.Body.machine_id = HardwareUtils.GenerateMachineID();
 
             logon.Body.game_server_token = details.Token;
 
@@ -161,7 +161,7 @@ namespace SteamKit2
 
             logon.Body.client_os_type = ( uint )Utils.GetOSType();
             logon.Body.game_server_app_id = ( int )appId;
-            logon.Body.machine_id = MachineIdUtils.GenerateMachineID();
+            logon.Body.machine_id = HardwareUtils.GenerateMachineID();
 
             this.Client.Send( logon );
         }
--- a/SteamKit2/SteamKit2/Steam/Handlers/SteamGameServer/SteamGameServer.cs
+++ b/SteamKit2/SteamKit2/Steam/Handlers/SteamGameServer/SteamGameServer.cs
@@ -126,7 +126,7 @@ namespace SteamKit2
 
             logon.Body.client_os_type = ( uint )Utils.GetOSType();
             logon.Body.game_server_app_id = ( int )details.AppID;
-            logon.Body.machine_id = HardwareUtils.GenerateMachineID();
+            logon.Body.machine_id = HardwareUtils.GetMachineID();
 
             logon.Body.game_server_token = details.Token;
 
@@ -161,7 +161,7 @@ namespace SteamKit2
 
             logon.Body.client_os_type = ( uint )Utils.GetOSType();
             logon.Body.game_server_app_id = ( int )appId;
-            logon.Body.machine_id = HardwareUtils.GenerateMachineID();
+            logon.Body.machine_id = HardwareUtils.GetMachineID();
 
             this.Client.Send( logon );
         }
--- a/SteamKit2/SteamKit2/Util/HardwareUtils.cs
+++ b/SteamKit2/SteamKit2/Util/HardwareUtils.cs
@@ -1,0 +1,191 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Management;
+using System.Net.NetworkInformation;
+using System.Text;
+using Microsoft.Win32;
+
+namespace SteamKit2
+{
+    abstract class MachineInfoProvider
+    {
+        public static MachineInfoProvider GetProvider()
+        {
+            switch ( Environment.OSVersion.Platform )
+            {
+                case PlatformID.Win32NT:
+                case PlatformID.Win32Windows:
+                    return new WindowsInfoProvider();
+            }
+
+            return new DefaultInfoProvider();
+        }
+
+        public abstract byte[] GetMachineGuid();
+        public abstract byte[] GetMacAddress();
+        public abstract byte[] GetDiskId();
+    }
+
+    class DefaultInfoProvider : MachineInfoProvider
+    {
+        public override byte[] GetMachineGuid()
+        {
+            return Encoding.UTF8.GetBytes( "SteamKit-MachineGuid" );
+        }
+
+        public override byte[] GetMacAddress()
+        {
+            // mono seems to have a pretty solid implementation of NetworkInterface for our platforms
+            // if it turns out to be buggy we can always roll our own and poke into /sys/class/net on nix
+
+            var firstEth = NetworkInterface.GetAllNetworkInterfaces()
+                .Where( i => i.NetworkInterfaceType == NetworkInterfaceType.Ethernet )
+                .FirstOrDefault();
+
+            if ( firstEth == null )
+            {
+                // well...
+                return Encoding.UTF8.GetBytes( "SteamKit-MacAddress" );
+            }
+
+            return firstEth.GetPhysicalAddress().GetAddressBytes();
+        }
+
+        public override byte[] GetDiskId()
+        {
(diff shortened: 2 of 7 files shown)


SteamKit2/SteamKit2/Util/HardwareUtils.cs
```diff
@@ -0,0 +1,235 @@
...
+            return Encoding.UTF8.GetBytes( "SteamKit-DiskId" );
+        }
+    }
+
+    class WindowsInfoProvider : DefaultInfoProvider
+    {
+        public override byte[] GetMachineGuid()
+        {
+            RegistryKey localKey = RegistryKey
+                .OpenBaseKey( Microsoft.Win32.RegistryHive.LocalMachine, RegistryView.Registry64 )
```
Netshroud: How does `Registry64` behave on 32-bit Windows?

VoiDeD: https://msdn.microsoft.com/en-us/library/microsoft.win32.registryview%28v=vs.110%29.aspx

"If you request a 64-bit view on a 32-bit operating system, the returned keys will be in the 32-bit view."

Netshroud:
OS X notes:
* Machine Guid is `IOPlatformSerialNumber`
* MAC Address is acquired by looping through `IOEthernetInterface` services from IOKit and getting the `IOMACAddress`.
* Disk ID is acquired via the DiskArbitration framework - I didn't even know this exists. Steam requests the `kDADiskDescriptionMediaUUIDKey` property from whichever disk is mounted at `/`.

That's going to be a pain to P/Invoke, I doubt Mono has a wrapper for DiskArbitration.

SteamKit2/SteamKit2/Util/HardwareUtils.cs
```diff
@@ -0,0 +1,235 @@
...
+                // well...
+                return Encoding.UTF8.GetBytes( "SteamKit-MacAddress" );
+            }
+
+            return firstEth.GetPhysicalAddress().GetAddressBytes();
+        }
+
+        public override byte[] GetDiskId()
+        {
+            return Encoding.UTF8.GetBytes( "SteamKit-DiskId" );
```
xPaw: I think that's too generic and it should return null instead. We don't want completely different servers to match as one machineid, do we?

VoiDeD: A null hash would have the same result.

xPaw: Don't hash it, if it's null?

VoiDeD: The hash field needs to be present and populated.

xPaw: Well that's no good, I'm not keen on having a generic hash this way.

SteamKit2/SteamKit2/Steam/Handlers/SteamGameServer/SteamGameServer.cs
```diff
@@ -161,7 +161,7 @@ public void LogOnAnonymous( uint appId = 0 )
 
             logon.Body.client_os_type = ( uint )Utils.GetOSType();
             logon.Body.game_server_app_id = ( int )appId;
-            logon.Body.machine_id = MachineIdUtils.GenerateMachineID();
+            logon.Body.machine_id = HardwareUtils.GenerateMachineID();
```
xPaw: I just nethooked steamcmd, and anonymous logins do not have machine_id set. And anonymous logins worked fine last night.

xPaw: Well I just checked normal logins, and they send an empty MessageObject on windows. Someone in steamworks forum was complaining that steamcmd didn't work last night :laughing:

SteamKit2/SteamKit2/Util/HardwareUtils.cs
```diff
@@ -0,0 +1,353 @@
...
+            }
+        }
+
+        public static byte[] GenerateMachineID()
+        {
+            // the aug 25th 2015 CM update made well-formed machine MessageObjects required for logon
+            // this was flipped off shortly after the update rolled out, likely due to linux steamclients running on distros without a way to build a machineid
+            // so while a valid MO isn't currently (as of aug 25th) required, they could be in the future and we'll abide by The Valve Law now
+
+            var machineId = new MachineID();
```
xPaw: Should this function cache constructed machine id, and just return the cached object on future calls? WMI is slow, and the result isn't going to cache between calls.

VoiDeD:
Okay, we now essentially have 1:1 functionality with how steamclient generates machine ids, along with a few extras:

* We additionally check `/etc/machine-id`, some gentoo systems have this while lack `/var/lib/dbus/machine-id`.
* We also use partition UUIDs for disk id.

I believe one last issue is outstanding for Linux: what can we do about getting some disk information for anyone not mounting their root directory by UUID?

Perhaps iterate `/dev/disk/by-uuid` and grab something from there?

Netshroud:
On OS X, Valve get the UUID for whichever disk is mounted to `/`. I assume we could do the same on Linux somehow?

If it's not mounted by UUID, there's gotta be another way to get the UUID for that disk.

SteamKit2/SteamKit2/Util/HardwareUtils.cs
```diff
@@ -0,0 +1,398 @@
...
+            return Encoding.UTF8.GetBytes( Environment.MachineName + "-SteamKit" );
+        }
+
+        public override byte[] GetMacAddress()
+        {
+            // mono seems to have a pretty solid implementation of NetworkInterface for our platforms
+            // if it turns out to be buggy we can always roll our own and poke into /sys/class/net on nix
+
+            var firstEth = NetworkInterface.GetAllNetworkInterfaces()
+                .Where( i => i.NetworkInterfaceType == NetworkInterfaceType.Ethernet )
```
Netshroud: What about other network interface types?

On OS X Mono 4.0.1, almost everything is showing as `Ethernet`.
On Windows .NET Framework, I'm seeing `Ethernet` for wired connections and `Wireless80211` for WiFi.

This seems like it could fall back to `"SteamKit-MacAddress"` a bit too frequently. What adapters does Steam search for?

SteamKit2/SteamKit2/Util/HardwareUtils.cs
```diff
@@ -0,0 +1,398 @@
...
+
+            return bootOptions.Split( ' ' );
+        }
+        string[] GetDiskUUIDs()
+        {
+            string[] files;
+
+            try
+            {
+                files = Directory.GetFiles( "/dev/disk/by-uuid" );
```
Netshroud: Does this return files in any order? If so, could this change just by adding or removing a USB flash drive, for example?

xPaw: > The order of the returned file names is not guaranteed
According to MSDN.

Netshroud: MSDN has nothing on Mono :smile:

SteamKit2/SteamKit2/Util/HardwareUtils.cs
VoiDeD: >and `Wireless80211` for WiFi.

Interesting, the [MSDN docs](https://msdn.microsoft.com/en-us/library/system.net.networkinformation.networkinterface.networkinterfacetype%28v=vs.100%29.aspx) don't state that that's a possible value. I suppose it can't hurt to include wireless adapters as well.

On Windows, Steam is using [`GetAdaptersInfo`](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365917%28v=vs.85%29.aspx) and checks for adapters of type `MIB_IF_TYPE_ETHERNET` and `MIB_IF_TYPE_OTHER`.

Netshroud:
```
IF_TYPE_IEEE80211
71
An IEEE 802.11 wireless network interface.
Note  This adapter type is returned on Windows Vista and later. On Windows Server 2003 and Windows XP , an IEEE 802.11 wireless network interface returns an adapter type of MIB_IF_TYPE_ETHERNET.
```

I wonder when the machine ID code was written.

Netshroud:
LGTM :+1: 

I don't think it should be reasonably able to fall back to the `DefaultInfoProvider` strings now, unless something goes catastrophically wrong.

PR #167, opened 2015-09-25
Update NetHookAnalyzer2 to handle more SO GC message, with specialization.

psychonic:
This adds handling for 16 more Dota 2 SO object types, as well as using the specialization in two more Dota 2 SO-related GC messages.

This probably could have been done cleaner, but it does work. We could probably also split out into base types (the econ ones with low ids), and then have game-specific ones added onto that. Either way, this is progress.

### DIFF
--- a/Resources/NetHookAnalyzer2/NetHookAnalyzer2/Specializations/Dota2SOCacheSubscribedGCSpecialization.cs
+++ b/Resources/NetHookAnalyzer2/NetHookAnalyzer2/Specializations/Dota2SOCacheSubscribedGCSpecialization.cs
@@ -1,0 +1,63 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using ProtoBuf;
+using ProtoBuf.Meta;
+using SteamKit2.GC.Dota.Internal;
+using SteamKit2.GC.Internal;
+
+namespace NetHookAnalyzer2.Specializations
+{
+    class Dota2CacheSubscribedGCSpecialization : IGameCoordinatorSpecialization
+    {
+        const uint Dota2AppID = 570;
+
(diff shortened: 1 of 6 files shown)


Resources/NetHookAnalyzer2/NetHookAnalyzer2/Specializations/Dota2SOCacheSubscribedGCSpecialization.cs
```diff
@@ -0,0 +1,63 @@
...
+
+            foreach (var bucket in cacheSubscribed.objects)
+            {
+                int typeId = bucket.type_id;
+                foreach (var singleObject in bucket.object_data)
+                {
+                    var extraNode = ReadExtraObject(singleObject, typeId);
+                    if (extraNode != null)
+                    {
+                        yield return new KeyValuePair<string, object>("SO ", extraNode);
```
Netshroud: Is the space here neccesary?

Resources/NetHookAnalyzer2/NetHookAnalyzer2/Specializations/Dota2SOMultipleObjectsGCSpecialization.cs
```diff
@@ -58,14 +59,11 @@ object ReadExtraObject(CMsgSOMultipleObjects.SingleObject sharedObject)
...
 				{
-					switch (sharedObject.type_id)
-					{
-						case 2003:
-							return RuntimeTypeModel.Default.Deserialize(ms, null, typeof(CSODOTAParty));
-
-						case 2004:
-							return RuntimeTypeModel.Default.Deserialize(ms, null, typeof(CSODOTALobby));
-					}
+                    Type t;
```
Netshroud: Indentation here is off

PR #168, opened 2015-09-29
Add the ability to use a pre-connected socket.

paralin:
This enables the use of proxies and other non-standard connection methods.

---
This change is [Reviewable](https://reviewable.io/reviews/steamre/steamkit/168)


VoiDeD:
Is there a compelling use case for this?

The current implementation introduces quite a bit of redundant code, and it doesn't appear to be exposed to any callers.

PR #170, opened 2015-10-18
Add async/await support for job based messages

VoiDeD:
So here's an experiment that might be worth mainlining.

The idea here is that we introduce an `AsyncJob` object that inherits from `JobID`, and have all our job related client message handler functions return this object. Client code will continue to compile since existing code can work with the `JobID` superclass just fine.

The `AsyncJob` objects expose a `GetAwaiter()` to support awaiting directly on that object, and a `ToTask()` to get the underlying TPL task that the job itself is wrapping.

Awaiting the result of the job will directly return the callback object for that job, i.e.: awaiting a call to `GetAppOwnershipTicket` will return a `AppOwnershipTicketCallback` without any callback subscription boilerplate.

The `AsyncJob` instances are created by client message handlers, and get tied to the handlers' parent `SteamClient` instance. `SteamClient` then handles the job's completion through `PostCallback`: when a callback for that `JobID` is posted, `SteamClient` completes the job and await returns the callback.

`SteamClient` is also tasked with timing out any jobs that never receive a response from Steam. The default timeout is 1 minute, but can be managed with the `AsyncJob.Timeout` property. Any jobs that don't receive a response are canceled and can be handled through regular exception handling.

There's also an additional `AsyncJobMultiple` for the instances where Steam can reply with multiple messages to a single job based request. In this case, awaiting on this job will instead return a `ResultSet` object that contains a list of the retrieved callbacks. There's also a few cases involved with this one:

1. If Steam replies with all the messages expected for the request, the `ResultSet` is returned with all the messages and `Complete` will indicate the result set is complete.
2. If Steam replies with only a few messages (and then the timeout is reached while waiting for the rest of the set), the `ResultSet` will be returned with what the client managed to receive and `Complete` will be set to false.
3. If Steam doesn't reply to the request at all, the job is entirely canceled and the exception should be handled.

Caveats:
* Bumping up our .NET requirement to 4.5 or Mono 3.0 (which was released Oct 19, 2012).
* Some of the new unit tests require async/await support, which is only supported with xunit 2.0.0.
* xunit 2.0.0 and Travis CI's current stable build of mono have issues (see xunit/xunit#158), so we inform Travis that we need an alpha version of mono (4.2) to run our tests.


Netshroud:
> we introduce an AsyncJob object that inherits from JobID, and have all our job related client message handler functions return this object. Client code will continue to compile since existing code can work with the JobID superclass just fine.

I don't like this bit... it doesn't make sense to me for a job to be a subclass of the job id from a pure design perspective, it seems like the only justification for this is backwards compatibility.

Compilation compatibility could also be achieved through an implicit cast operator, but that would fall over for consumers that use `var` instead of declaring their receiving type, if they use `JobID` directly and don't just immediately pass it into a function that takes `JobID` as a parameter. For the 99% use case, this ought to suffice.

Otherwise, I'd rather make a breaking change and bump the major version (finally?).

VoiDeD:
We still need to keep JobIDs around for matching callbacks to requests into the far foreseeable future, even for a major version bump I wouldn't want to make a breaking change like that. Some sort of implicit cast sounds compelling, though.

VoiDeD:
After thinking about it a little, I don't think any implicit casts will work here either. Consider:

```c#
AppOwnershipTicketCallback callback = await steamApps.GetAppOwnershipTicket( 440 );
```

I don't think the compiler will be able to work out that `GetAppOwnershipTicket`'s return needs to be casted into an `AsyncJob`.

Netshroud:
I was thinking the other way around.

```
AsyncJob asyncJob = steamApps.GetAppOwnershipTicket( 440 );
AppOwnershipTicket callback = await asyncJob;
```

```
void Foo( JobID jobID ) { ... }

JobID myJobID = steamApps.GetAppOwnershipTicket( 440 );
Foo( myJobID );
```

```
void Foo( JobID jobID ) { ... }

var /* actually an AsyncJob now */ myJobID = steamApps.GetAppOwnershipTicket( 440 );
Foo( myJobID );
```

VoiDeD:
Okay, with this we now have some initial handling for heartbeats and remote failures (from #171). I've also switched it over to an implicit cast, which I _think_ retains the same amount of source code compatibility as before.

All that's really left is to implement the logic and tests for remote failure exceptions.

xPaw:
This branch has been running in production at @SteamDatabase for a couple of days now, and makes use of `await`, no issues.

Netshroud:
> "The name of an async method, by convention, ends with an "Async" suffix." - [Asynchronous Programming with Async and Await (C# and Visual Basic)](https://msdn.microsoft.com/en-us/library/hh191443.aspx)

Worth considering, at the risk of breaking backwards compatibility?

VoiDeD:
Far too much breakage in my opinion.

xPaw:
Not directly related to this PR, but perhaps look into making `CDNClient` async too?

xPaw:
There might be a problem with timing out tasks. I've ran into it twice in different scenarios, one was inside of a task that has an exception handler in `ContinueWith`, and another one where the code essentially boils down to `try { await steam.something(); } catch ...` and in both cases the task cancellation exception never reached my handlers.

```
18:40:40 [ERROR] Unhandled Exception: A task was canceled.
  at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess (System.Threading.Tasks.Task task) [0x0002d] in /tmp/buildd/mono-4.0.4.1/external/referencesource/mscorlib/system/runtime/compilerservices/TaskAwaiter.cs:193
  at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification (System.Threading.Tasks.Task task) [0x0002e] in /tmp/buildd/mono-4.0.4.1/external/referencesource/mscorlib/system/runtime/compilerservices/TaskAwaiter.cs:170
  at System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd (System.Threading.Tasks.Task task) [0x0000b] in /tmp/buildd/mono-4.0.4.1/external/referencesource/mscorlib/system/runtime/compilerservices/TaskAwaiter.cs:142
  at System.Runtime.CompilerServices.TaskAwaiter`1[SteamKit2.SteamUserStats+NumberOfPlayersCallback].GetResult () [0x00000] in /tmp/buildd/mono-4.0.4.1/external/referencesource/mscorlib/system/runtime/compilerservices/TaskAwaiter.cs:372
  at SteamDatabaseBackend.PlayersCommand+<OnCommand>c__async0.MoveNext () [0x00152] in /home/gfreeman/steamdb/deploy/IRC/Commands/Players.cs:56
--- End of stack trace from previous location where exception was thrown ---
  at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw () [0x0000b] in /tmp/buildd/mono-4.0.4.1/mcs/class/corlib/System.Runtime.ExceptionServices/ExceptionDispatchInfo.cs:61
  at System.Runtime.CompilerServices.AsyncMethodBuilderCore.<ThrowAsync>m__1 (System.Object state) [0x00000] in /tmp/buildd/mono-4.0.4.1/external/referencesource/mscorlib/system/runtime/compilerservices/AsyncMethodBuilder.cs:1022
```

This exception was logged in `AppDomain.CurrentDomain.UnhandledException`.

Await in OnCommand: https://github.com/SteamDatabase/SteamDatabaseBackend/blob/master/IRC/Commands/Players.cs#L56

OnCommand is called here, which includes try/catch: https://github.com/SteamDatabase/SteamDatabaseBackend/blob/master/IRC/CommandHandler.cs#L190-L213

cH40z-Lord:
What is the reason for the `AsyncJob<T>` class? Why is it used over `Task<T>`? When working with `async/await` one should also not forget about implementing `CancellationToken` which is the preferred way to cancel async operations. While I like the idea of having async operations available I strongly encourage to stick to existing classes unless there is a good reason not to do do.

If it's due to backwards compatibility then I'd either recommend splitting this feature into a dedicated NuGet package or following [SemVer](http://semver.org/)'s guidelines and release a new major version that explicitly allows breaking backwards compatibility changes.

VoiDeD:
Backwards compatibility is indeed the reasoning. We don't have any plans to make any significant breaking changes (such was removing our entire callback scheme for job based messages) any time in the future. Having a separate library version that follows the TPL conventions more closely is an interesting idea, but I wouldn't be happy with the maintenance burden.

SteamKit2/SteamKit2/Steam/SteamClient/SteamClient.cs
```diff
@@ -342,5 +368,14 @@ void HandleServerList( IPacketMsg packetMsg )
             PostCallback( new ServerListCallback( listMsg.Body ) );
         }
 
+        void HandleJobHeartbeat( IPacketMsg packetMsg )
+        {
+            jobManager.HeartbeatJob( packetMsg.TargetJobID );
+        }
+        void HandleJobFailed( IPacketMsg packetMsg )
```
xPaw: It might be a good idea to turn this into a callback too, so that users not using `async` can see catch their failed jobs.

VoiDeD: There's #171 regarding this, but I'm not sure if consumers really want to go through the code gymnastics of handling failures (and likely heartbeats) by having multiple callbacks.

PR #173, opened 2015-10-19
Fix default SteamID instance when parsing Steam3 ids

xPaw:
Only individual steamids default to desktop, everything else is 0.

### DIFF
--- a/SteamKit2/SteamKit2/Types/SteamID.cs
+++ b/SteamKit2/SteamKit2/Types/SteamID.cs
@@ -290,27 +290,15 @@ namespace SteamKit2
                 return false;
             type = typeString[ 0 ];
 
-            uint instance;
+            uint instance = 0;
             var instanceGroup = m.Groups[ "instance" ];
             if ( instanceGroup != null && !string.IsNullOrEmpty( instanceGroup.Value ) )
             {
                 instance = uint.Parse( instanceGroup.Value );
             }
-            else
+            else if ( type == 'U' )
             {
-                switch ( type )
-                {
-                    case 'g':
-                    case 'T':
-                    case 'c':
-                    case 'L':
-                        instance = 0;
-                        break;
-
-                    default:
-                        instance = 1;
-                        break;
-                }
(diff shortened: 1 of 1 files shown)


SteamKit2/SteamKit2/Types/SteamID.cs
```diff
@@ -290,27 +290,15 @@ public bool SetFromSteam3String( string steamId )
...
-                    case 'c':
-                    case 'L':
-                        instance = 0;
-                        break;
-
-                    default:
-                        instance = 1;
-                        break;
-                }
+                instance = 1;
```
VoiDeD: Using the `DesktopInstance` constant would be more ideal.

Netshroud:
Add a unit test.

I'm confused as to how a change to `SetFromSteam3String` fixes the instance when parsing **Steam2** IDs.

I'm also confused as to how this unit test is still passing with your changes: https://github.com/xPaw/SteamKit/blob/steamid-instance/SteamKit2/Tests/SteamIDFacts.cs#L267-L295

xPaw:
Netshroud mentioned `i` type to me, and as it turns out SteamKit throw an exception when trying to parse a steamid like `[i:1:1]`, because there is no corresponding type in `AccountTypeChars` dictionary.

I've looked around and can't find `i` type anywhere besides SteamKit. Should it be removed from regex? In `RenderSteam3` it defaults to `i`, but probably should default to `I`.

VoiDeD:
Steamclient renders `i` for SteamIDs with unknown `AccountType`s. `I` is used when the account type is explicitly set to `Invalid`.

xPaw:
Yeah @Netshroud fixed it: 2a7e9ab08c8a239e6a46854113d684b67a8a8837

xPaw:
Since the behaviour is not 100% consistent within Valve's codebase, I am going to close this pull request. Even though all parsing for gameservers defaults to universe 1, generated gameserver steamids (tf2/csgo) have universe set to 0 by default.

PR #174, opened 2015-10-21
ProtobufDumper improvements

xPaw:
* Removes unsafe code and replaces it with MemoryStream.
* At least 4x speed up on `steamclient.dylib`
* Defers all protos until the end (to fix duplicate protos)

### DIFF
--- a/Resources/ProtobufDumper/ProtobufDumper/ImageFile.cs
+++ b/Resources/ProtobufDumper/ProtobufDumper/ImageFile.cs
@@ -59,6 +59,8 @@ namespace ProtobufDumper
         private Dictionary<string, int> enumLookupCount;
         private List<string> deferredEnumTokens;
 
+        private Regex ProtoFileNameRegex;
+
         public ImageFile(string fileName, string output = null)
         {
             this.FileName = fileName;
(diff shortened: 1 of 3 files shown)


xPaw:
Proof that this PR works: https://github.com/SteamDatabase/SteamTracking/commit/67f6065ff843c8784aca12d6346a49fc1877d324

VoiDeD:
Any regressions with running this on windows binaries or is that still functional?

Resources/ProtobufDumper/ProtobufDumper/ImageFile.cs
```diff
@@ -410,8 +405,25 @@ private void FinalPassWriteFiles()
...
                     }
                 }
-                
-                File.WriteAllText(outputFile, proto.buffer.ToString());
+
+                var buffer = proto.buffer.ToString();
+
+                if (alreadyWrittenProtos.ContainsKey(outputFile))
+                {
+                    if (alreadyWrittenProtos[outputFile] == buffer)
```
VoiDeD: This is one heck of a string comparison, but it's probably fine.

Resources/ProtobufDumper/ProtobufDumper/ImageFile.cs
```diff
@@ -668,7 +680,25 @@ private string GetValueForProp(dynamic propInfo, object options, bool field, out
                         string value = GetValueForProp(fieldInfo, deserialized, true, out name);
 
                         if (value != null)
-                            options_kv.Add(name, value);
+                        {
+                            if (options_kv.ContainsKey(name))
+                            {
+                                if (options_kv[name] == value)
+                                {
+                                    Console.WriteLine("Found a duplicate option '{0}' with same value, ignoring...", name);
```
VoiDeD: I'm curious if the serialized proto descriptors actually include duplicate options in their byte steams, or if this is a bug elsewhere in protobuf dumper.

xPaw:
Yep, Windows dumping still works.

xPaw:
With my changes, ProtobufDumper is at about 4 times faster when run on `steamclient.dylib` file.

* master: ~3800ms
* back-to-the-future: ~930ms

VoiDeD:
LGTM if it works.

Netshroud:
LGTM too. No known regressions and an impressive performance boost, but I want final confirmation from AzuiSleet before merging. I'm getting mixed messages from IRC:

```
[05:28:13] <AzuiSleet> safeguard is a no-go. I'll take a look at what deferredProtos needs to do
...
[05:53:38] <AzuiSleet> alright, it looks good, closer to 4 passes though
```

(safeguard is still in there)

All OK, @azuisleet?

Netshroud:
For the record:
```
<AzuiSleet> safeguard is OK because the previous code simply looped until 0
```

Merging...

PR #180, opened 2015-10-23
Generate GC base separately for each game

xPaw:
I'm not going to regenerate them because it destroys them due to line endings *(honk honk, someone doesn't want to normalize them)*


VoiDeD:
I don't think I mind line endings on dumped/generated files at this point since SteamDB maintains a better record of history than we do.

xPaw:
Ok, I've regenerated everything and removed no longer existing files, and also the non-generated EMsg enums for CSGO and TF2 that were super outdated.

`MsgGCHdrProtoBuf` in `gamecoordinator.steamd` uses Dota's `CMsgProtoBufHeader`, I've kept it that way, but should it stay that way?

Netshroud:
I'd assume the `CMsgProtoBufHeader` would be consistent across all GCs and the CMs, since it's in `steammessages_base.proto` and it appears to be a common structural definition.

If so, `MsgGCHdrProtoBuf` could, and perhaps should, just use `SteamKit2.Internal.CMsgProtoBufHeader`. Otherwise we could make it generic and the consumer would specify their header proto on a per-message / per-GC basis, but that seems like a lot of hassle for no value.

xPaw:
Dota's `CMsgProtoBufHeader` might not be compatible because it has 2 fields that steamclient doesn't and they seem directly related to GC stuff:
```
	optional .GCProtoBufMsgSrc gc_msg_src = 200 [default = GCProtoBufMsgSrc_Unspecified];
	optional uint32 gc_dir_index_source = 201;
```

Also, while I'm cleaning up proto stuff, is there any point in having protos in the repo that aren't used for anything? (portal, dod, etc).

VoiDeD:
I'm not sure if relying on a specific game's header is the right approach either.

One idea that crossed my mind is maintaining a copy of our own .proto with a copy of whatever header looks most up to date, and simply using that. This allows us to control our own field changes.

If some day games diverge enough to have differing header fields, we can look into allowing consumers to specify what header they want.

VoiDeD:
LGTM unless @Netshroud has any other ideas.

Issue #182, opened 2015-10-26
KeyValue class, inherit IDictionary<string, KeyValue>


WildCard65:
It would make KeyValue have more functionality if it inherits the interface Dictionary<> uses, it already implements 1 of IDictionary<>'s properties which is the this[TKey key] { get; set; } property.

PR #183, opened 2015-10-26
Extended the KeyValue class

WildCard65:
This PR's purpose is to extend the KeyValue class with useful interfaces based around how it works. Also this closes: https://github.com/SteamRE/SteamKit/issues/182

### DIFF
--- a/SteamKit2/SteamKit2/Types/KeyValue.cs
+++ b/SteamKit2/SteamKit2/Types/KeyValue.cs
@@ -4,6 +4,7 @@
  */
 
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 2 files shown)


WildCard65:
I have no idea what I did to the tests dll

SteamKit2/SteamKit2/Types/KeyValue.cs
```diff
@@ -262,6 +315,51 @@ public KeyValue( string name = null, string value = null )
         /// </summary>
         public List<KeyValue> Children { get; private set; }
 
+        /// <summary>
+        /// This gets the all the children's keys of this <see cref="KeyValue"/>.
+        /// </summary>
+        public ICollection<string> Keys
```
VoiDeD: This doesn't make sense at all. A KeyValue doen't have a collection of keys.

WildCard65: It's there due to IDictionary<>

VoiDeD:
This interface implementation doesn't make logical sense. There's no single string -> KeyValue mapping that exists as a key can have multiple KeyValues as children.

WildCard65:
The implementation is just basically follows how this[string key] works

Netshroud:
I don't see the value here whatsoever.

1. Why do you need `KeyValue` to implement `IConvertible`? Does it even work? There are no unit tests covering this.

2. Why do you need `KeyValue` to implement `IEquatable`? What does it provide over `object.Equals(object)`? This has inadequate test coverage.

3. Why do you need `KeyValue` to implement `IComparable`? In what case does ordering by both name and value make any sense? This has inadequate test coverage.

4. Why do you need `KeyValue` to implement `IDictionary`? The nature of `KeyValue` means that bolting on `IDictionary` conformance, as you've done, makes no sense. A dictionary has keys and values, whereas a `KeyValue` has a name, and either a value or children. It's a dodgy Valve construct that doesn't directly map to any standard collection types. The usage (e.g. `var theRealValue = myDictionaryThatIsActuallyKeyValues[ key ].Value`) is also unexpected for consumers of the generic interface.

WildCard65:
The IConvertible implementation wraps around String's implementation of the interface. I may expand the single test covering the Uint16 conversion into covering all the methods.

The Object.Equals only compares references by default unless an override of Equals(object obj) exists. IEquatable provides typed overrides instead of the base obj type. There are tests that cover this.

The IComparible thing is for OrderBy(). At least 2 tests exist to test it.

The IDictionary thing is to easily wrap a KeyValue's Children into KeyValuePair<> that follows the logic of how you implemented the public KeyValue this[ string key ] property(which is a property of IDictionary)

VoiDeD:
You haven't answered _why_ you need KeyValue to implement interfaces that don't align with its data. So far it sounds like you want to implement them just because you want to implement them.

WildCard65:
To me, it makes sense for keyvalue to have it to iterate it's children
dictionary like

Netshroud:
I didn't say "no tests", I said "inadequate test coverage". You haven't tested most of the code branches, including possible exception branches (what does `new KeyValue("bob", "george").ToDateTime()` do?).

And still, as VoiDeD said, you haven't provided any compelling case for this exhaustive interface compliance - not for one of them, let alone four.

If you just want to be able to iterate over a `KeyValue`'s children, `KeyValue.Children` is iterable. It also doesn't justify the other three interfaces.

WildCard65:
I never was good at explaining things

Netshroud:
Unfortunately, communication skills are (sometimes overlooked) prerequisite to software development.

I can't justify merging this, and the future maintenance burden that comes along with all of this, without a compelling reason. "Just because" isn't good enough.

VoiDeD:
Going to reject this as the interface implementations don't fit the data.

PR #198, opened 2015-11-06
Use C# 6.0 expression bodied properties

stil:
Since we're already using VS2015 to compile, using this C# 6.0 feature
would make codebase slightly lighter and more readable.


SteamKit2/SteamKit2/Base/ClientMsg.cs
```diff
@@ -28,14 +28,15 @@ internal ClientMsgProtobuf( int payloadReserve = 0 )
...
+        public override bool IsProto => true;
+
         /// <summary>
         /// Gets the network message type of this client message.
         /// </summary>
         /// <value>
         /// The network message type.
         /// </value>
-        public override EMsg MsgType { get { return Header.Msg; } }
+        public override EMsg MsgType => Header.Msg;
```
VoiDeD: I think an expression body change like this modifies the property to be both get and set, something we don't want here. Ditto for a bunch of get only properties.

Netshroud: Expression-bodied properties are read-only, there's no setter.

VoiDeD: Okay I see, I had also seen the syntax that involved `{ get; } = ...` and thought that was the method to correctly make a read only property.

WildCard65: No the { get; } = thing is the same as the doing [DefaultValue()]

Netshroud: This is patently false.

`{ get; } = ...` instructs the compiler to emit an assignment to the backing field in the class's constructor (or static constructor for a static property).

`DefaultValueAttribute` does nothing, a framework has to explicitly search for it for it to be of use.

WildCard65: good to know then

VoiDeD:
Alright so I think the bottom line is that it's probably not harmful to update this syntax, but we'd be changing syntax (and making our blame history slightly more convoluted) just for the sake of changing syntax, rather than fixing any underlying issues.

PR #204, opened 2015-11-15
Create SteamKit logger to log packets in nethook format

xPaw:
This is initial implementation to gather feedback. What do you think?

### DIFF
--- a/Samples/3.DebugLog/Program.cs
+++ b/Samples/3.DebugLog/Program.cs
@@ -76,6 +76,12 @@ namespace Sample3_DebugLog
 
             // create our steamclient instance
             steamClient = new SteamClient();
+
+            // uncomment this if you'd like to dump raw sent and received packets
+            // that can be opened for analysis in NetHookAnalyzer
+            // NOTE: dumps may contain sensitive data (such as your Steam password)
+            //steamClient.DebugNetworkListener = new NetHookNetworkListener();
+
             // create the callback manager which will route callbacks to function calls
             manager = new CallbackManager( steamClient );
 
--- a/SteamKit2/SteamKit2/Steam/CMClient.cs
+++ b/SteamKit2/SteamKit2/Steam/CMClient.cs
@@ -83,6 +83,11 @@ namespace SteamKit2.Internal
         /// </value>
         public TimeSpan ConnectionTimeout { get; set; }
 
+        /// <summary>
+        /// Gets or sets the network listening interface. Use this for debugging only.
+        /// For your convenience, you can use <see cref="NetHookNetworkListener"/> class.
+        /// </summary>
+        public IDebugNetworkListener DebugNetworkListener { get; set; }
 
         Connection connection;
         bool encryptionSetup;
@@ -194,6 +199,14 @@ namespace SteamKit2.Internal
 
             DebugLog.WriteLine( "CMClient", "Sent -> EMsg: {0} (Proto: {1})", msg.MsgType, msg.IsProto );
 
+            try
+            {
+                DebugNetworkListener?.OnOutgoingNetworkMessage(msg.MsgType, msg.Serialize());
+            }
+            catch ( Exception e )
+            {
+                DebugLog.WriteLine( "CMClient", "DebugNetworkListener threw an exception: {0}", e );
+            }
 
             // we'll swallow any network failures here because they will be thrown later
             // on the network thread, and that will lead to a disconnect callback
@@ -241,6 +254,19 @@ namespace SteamKit2.Internal
             }
 
             DebugLog.WriteLine( "CMClient", "<- Recv'd EMsg: {0} ({1}) (Proto: {2})", packetMsg.MsgType, ( int )packetMsg.MsgType, packetMsg.IsProto );
+
+            // Multi message gets logged down the line after it's decompressed
+            if ( packetMsg.MsgType != EMsg.Multi )
+            {
+                try
+                {
+                    DebugNetworkListener?.OnIncomingNetworkMessage( packetMsg.MsgType, packetMsg.GetData() );
+                }
+                catch ( Exception e )
+                {
+                    DebugLog.WriteLine( "CMClient", "DebugNetworkListener threw an exception: {0}", e );
+                }
+            }
 
             // ensure that during channel setup, no other messages are processed
             if ( ( !encryptionSetup && pendingNetFilterEncryption == null && packetMsg.MsgType != EMsg.ChannelEncryptRequest ) ||
--- a/SteamKit2/SteamKit2/Util/DebugNetworkListener.cs
+++ b/SteamKit2/SteamKit2/Util/DebugNetworkListener.cs
@@ -1,0 +1,117 @@
+﻿/*
+ * This file is subject to the terms and conditions defined in
+ * file 'license.txt', which is part of this source code package.
+ */
+
+
+
+using System;
+using System.IO;
+using System.Threading;
+
+namespace SteamKit2
+{
+    /// <summary>
+    /// This is a debug utility, do not use it to implement your business logic.
+    /// 
+    /// This interface is used for logging network messages sent to and received from the Steam server that the client is connected to.
+    /// </summary>
+    public interface IDebugNetworkListener
+    {
+        /// <summary>
+        /// Called when a packet is received from the Steam server.
+        /// </summary>
+        /// <param name="msgType">Network message type of this packet message.</param>
+        /// <param name="data">Raw packet data that was received.</param>
+        void OnIncomingNetworkMessage( EMsg msgType, byte[] data );
+
+        /// <summary>
+        /// Called when a packet is about to be sent to the Steam server.
+        /// </summary>
+        /// <param name="msgType">Network message type of this packet message.</param>
+        /// <param name="data">Raw packet data that will be sent.</param>
+        void OnOutgoingNetworkMessage( EMsg msgType, byte[] data );
+    }
+
+    /// <summary>
+    /// Dump any network messages sent to and received from the Steam server that the client is connected to.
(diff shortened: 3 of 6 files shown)


VoiDeD:
Tossing some thoughts along with some from the IRC discussion:

1. I don't want to lock this down to one specific file format (filename, directory, etc), but rather allow consumers implement something like an `INetworkListener` interface, and let them implement the logic for where dumped traffic should go.

2. I don't think we should expose this from the constructor, it might give consumers the idea that it's a first class feature of `SteamClient`, rather than what it really is: a debugging convenience. So on that topic, something like adding a `SetNetworkListener` on `CMClient` would likely be more ideal.

3. Also don't want to make it convenient for consumers to do something like:
```c#
public void OnIncomingMessage( EMsg msgType, byte[] data )
{
    if ( msgType == EMsg.ClientLogOnResponse )
    {
        // business logic...
    }
}
```
So what if we made this a DEBUG build only feature?

xPaw:
I agree with first two, however I think the nethook implementation should be kept as well for ease of use with nethook analyzer.

SteamKit2/SteamKit2/Steam/CMClient.cs
```diff
@@ -189,6 +192,10 @@ public void Send( IClientMsg msg )
 
             DebugLog.WriteLine( "CMClient", "Sent -> EMsg: {0} (Proto: {1})", msg.MsgType, msg.IsProto );
 
+            if ( netHookLogger != null )
+            {
+                netHookLogger.LogOutgoingNetMessage( msg.MsgType, msg.Serialize() );
```
Netshroud: Use C# 6 null-conditional operator. :)

SteamKit2/SteamKit2/Util/NetHookLogger.cs
```diff
@@ -0,0 +1,90 @@
...
+        private uint MessageNumber;
+        private string LogDirectory;
+
+        /// <summary>
+        /// Will create a folder in path "%assembly%/nethook/%currenttime%/"
+        /// </summary>
+        public NetHookLogger()
+        {
+            LogDirectory = Path.Combine(
+                Path.GetDirectoryName( System.Reflection.Assembly.GetExecutingAssembly().Location ),
```
Netshroud: Use `CodeBase`, not `Location`.

Netshroud:
I agree with @VoiDeD, but I do think a NetHook-compatible implementation would be useful - both for myself, as well as a reference implementation for that interface.

I agree with point 3, but think this would also be useful in Release builds (e.g. to monitor and report certain unhandled message types such as VAC). I'm not sure how else to prevent if-else handling by consumers, though.

Samples/3.DebugLog/Program.cs
```diff
@@ -75,7 +75,9 @@ static void Main( string[] args )
             pass = args[ 1 ];
 
             // create our steamclient instance
-            steamClient = new SteamClient();
+            steamClient = new SteamClient( System.Net.Sockets.ProtocolType.Tcp );
```
VoiDeD: Nit: Don't need to modify the constructor call here.

SteamKit2/SteamKit2/Util/DebugNetworkListener.cs
```diff
@@ -0,0 +1,115 @@
...
+        /// <param name="msgType">Network message type of this packet message.</param>
+        /// <param name="data">Raw packet data that will be sent.</param>
+        void OnOutgoingNetMessage( EMsg msgType, byte[] data );
+    }
+
+    /// <summary>
+    /// Dump any network messages sent to and received from the Steam server that the client is connected to.
+    /// These messages are dumped to file, and can be analyzed further with NetHookAnalyzer, a hex editor, or your own purpose-built tools.
+    /// </summary>
+    public class NetHookLogger : IDebugNetworkListener
```
VoiDeD: What about calling it `NetHookNetworkListener`?

Samples/3.DebugLog/Program.cs
xPaw: Unsure why it got changed, will fix.

VoiDeD:
This is probably fine to introduce unless @Netshroud has any other reservations.

The only last consideration is any security exceptions from `NetHookNetworkListener` attempting to write to a directory without permissions.

Samples/3.DebugLog/Program.cs
```diff
@@ -76,6 +76,8 @@ static void Main( string[] args )
 
             // create our steamclient instance
             steamClient = new SteamClient();
+            steamClient.DebugNetworkListener = new NetHookNetworkListener();
```
Netshroud: Is this sample really the place for this? Users might have their passwords stored locally in dumps and not realise it.

SteamKit2/SteamKit2/Util/DebugNetworkListener.cs
```diff
@@ -0,0 +1,115 @@
...
+    /// This interface is used for logging network messages sent to and received from the Steam server that the client is connected to.
+    /// </summary>
+    public interface IDebugNetworkListener
+    {
+        /// <summary>
+        /// Called when a packet is received from the Steam server.
+        /// </summary>
+        /// <param name="msgType">Network message type of this packet message.</param>
+        /// <param name="data">Raw packet data that was received.</param>
+        void OnIncomingNetMessage( EMsg msgType, byte[] data );
```
Netshroud: I don't like the 'Net' shorthand, but meh.

Samples/3.DebugLog/Program.cs
xPaw: Perhaps commenting it out by default would be better?

SteamKit2/SteamKit2/Util/DebugNetworkListener.cs
```diff
@@ -0,0 +1,115 @@
...
+
+            LogDirectory = path;
+        }
+
+        /// <summary>
+        /// Called when a packet is received from the Steam server.
+        /// </summary>
+        /// <param name="msgType">Network message type of this packet message.</param>
+        /// <param name="data">Raw packet data that was received.</param>
+        public void OnIncomingNetMessage( EMsg msgType, byte[] data )
```
Netshroud: Seems like a good candidate for an expression-bodied method, but again meh.

SteamKit2/Tests/NetworkListenerFacts.cs
```diff
@@ -0,0 +1,42 @@
+﻿using Xunit;
+using SteamKit2;
+using SteamKit2.Internal;
+
+namespace Tests
+{
+    class TestNetworkListener : IDebugNetworkListener
+    {
+        public EMsg NextExpectedSentMessage;
```
Netshroud: This is backwards from how I've usually seen tests like this - store the last one in the stub, assert from the test itself. Here you have no control from the test itself about where this runs.

SteamKit2/Tests/NetworkListenerFacts.cs
```diff
@@ -0,0 +1,42 @@
...
+        public void OnIncomingNetMessage(EMsg msgType, byte[] data)
+        {
+            //
+        }
+
+        public void OnOutgoingNetMessage(EMsg msgType, byte[] data)
+        {
+            var expected = NextExpectedSentMessage;
+
+            NextExpectedSentMessage = 0;
```
Netshroud: This doesn't require a cast?

SteamKit2/Tests/NetworkListenerFacts.cs
```diff
@@ -0,0 +1,42 @@
...
+
+            Assert.Equal(expected, msgType);
```
Netshroud: Assert data too - maybe just a simple test, such as not null and length > 0

SteamKit2/Tests/NetworkListenerFacts.cs
xPaw: Doesn't appear to.

Samples/3.DebugLog/Program.cs
Netshroud: I'd probably make it a separate sample with warnings everywhere, or leave it unsampled. It seems fairly simple enough to add into any pre-existing sample.

SteamKit2/SteamKit2/Steam/CMClient.cs
```diff
@@ -194,6 +199,7 @@ public void Send( IClientMsg msg )
 
             DebugLog.WriteLine( "CMClient", "Sent -> EMsg: {0} (Proto: {1})", msg.MsgType, msg.IsProto );
 
+            DebugNetworkListener?.OnOutgoingNetMessage( msg.MsgType, msg.Serialize() );
```
Netshroud: What happens if the debug listener throws an exception? Same for incoming below.

xPaw: DebugLog can throw too, can't it?

SteamKit2/SteamKit2/Util/DebugNetworkListener.cs
```diff
@@ -0,0 +1,115 @@
...
+        /// Will create a folder in path "%assembly%/nethook/%currenttime%/"
+        /// </summary>
+        public NetHookNetworkListener()
+        {
+            var uri = new Uri( System.Reflection.Assembly.GetExecutingAssembly().CodeBase );
+
+            LogDirectory = Path.Combine(
+                Path.GetDirectoryName( uri.LocalPath ),
+                "nethook",
+                DateUtils.DateTimeToUnixTime( DateTime.Now ).ToString()
```
Netshroud: Seems a bit racy - two concurrently created `SteamClient`s with separate listeners could both try claim the same directory.

SteamKit2/SteamKit2/Steam/CMClient.cs
Netshroud: Yep, but that doesn't make it right.

Netshroud:
:+1:

Issue #181, opened 2015-10-26
ServerList should handle SteamUser.LogOff


Netshroud:
Contrary to our xmldoc, `SteamUser.LogOff` gets the CM to close the connection, which we treat as a bad connection. We should have some way to expect a server-side disconnection.

VoiDeD:
Only slightly related, but I vote to update our documentation here.

PR #205, opened 2015-11-16
Mark LogOff-triggered disconnections as user-initiated

Netshroud:
Had to do a few hacky things I don't particularly like. Any suggestions to improve this?

Ideally, this probably needs a full-blown state machine.

### DIFF
--- a/SteamKit2/SteamKit2/Steam/CMClient.cs
+++ b/SteamKit2/SteamKit2/Steam/CMClient.cs
@@ -158,7 +158,7 @@ namespace SteamKit2.Internal
         /// The <see cref="IPEndPoint"/> of the CM server to connect to.
         /// If <c>null</c>, SteamKit will randomly select a CM server from its internal list.
         /// </param>
-        public virtual void Connect( IPEndPoint cmServer = null  )
+        public void Connect( IPEndPoint cmServer = null  )
         {
             this.Disconnect();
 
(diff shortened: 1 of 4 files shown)


Netshroud:
Re-did this a lot simpler to resolve #181.

SteamKit2/SteamKit2/Steam/CMClient.cs
```diff
@@ -83,6 +83,13 @@ public IPAddress LocalIP
         /// </value>
         public TimeSpan ConnectionTimeout { get; set; }
 
+        /// <summary>
+        /// Gets or sets whether or not this <see cref="CMClient"/> should imminently expect the server to close the connection.
+        /// If this is true when the connection is closed, the <see cref="SteamClient.DisconnectedCallback"/>'s <see cref="SteamClient.DisconnectedCallback.UserInitiated"/> property
+        /// will be set to <c>true</c>.
+        /// </summary>
+        public bool ExpectDisconnection { get; set; }
```
VoiDeD: Don't think this needs to be entirely public, consumers might be able to misuse it.

Netshroud: I was thinking that, but I don't want our own `ClientMsgHandlers` to be able to do things that consumer-built ones cannot.

VoiDeD: CMClient could likely handle that logic based on the emsg being passed to `Send`.

PR #211, opened 2015-12-07
Update and regenerate steamclient protobufs

DoctorMcKay:


DoctorMcKay:
`supports_rate_limit_response` presumably tells the CM whether or not the client can get back a `RateLimitExceeded` eresult in response to logon (currently a logon rate-limit comes back as `InvalidPassword`).

This doesn't appear to work (yet) though.

Netshroud:
Considering that `protocol_version` is in the same message, I would hope Valve haven't created a flag for what should be a protocol version bump.

DoctorMcKay:
This *is* Valve that we're talking about.

Netshroud:
Yep. I'd hope that above a certain protocol version, the flag can be toggled on or off. Below that protocol version, it's treated as always off.

Then again, Valve.

PR #218, opened 2015-12-13
Maintain Server List Ordering

### LINKED ISSUE #178: CMServerList should respect servers order
When specifying a cellid, Steam returns servers that are ordered (geographically?) starting with closest at the top.

CMServerList should use that order when weighting servers.

Netshroud:
Resolves #178.

### DIFF
--- a/SteamKit2/SteamKit2/Networking/Steam3/CMServerList.cs
+++ b/SteamKit2/SteamKit2/Networking/Steam3/CMServerList.cs
@@ -127,9 +127,9 @@ namespace SteamKit2
             lock ( listLock )
             {
                 var distinctEndPoints = listToMerge.Distinct().ToArray();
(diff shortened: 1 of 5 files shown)


SteamKit2/SteamKit2/Networking/Steam3/CMServerList.cs
```diff
@@ -122,6 +115,36 @@ public bool TryAddRange( IEnumerable<IPEndPoint> endPoints )
...
+                }
+                
+                for ( var i = 0; i < distinctEndPoints.Length; i++ )
+                {
+                    AddCore( distinctEndPoints[ i ], i );
+                }
+            }
+        }
+
         void Add( IPEndPoint endPoint )
```
xPaw: Nothing besides `UseInbuiltList` calls this method, I'd say just remove it.

PR #232, opened 2016-01-20
[WIP] KeyValue: Handle (ignore) conditionals

Netshroud:
As pointed out by @xPaw, we crash when reading `tf_english.txt`.

I narrowed this down to our handling of conditionals - it's completely broken. We count the conditional as a key, then read the next key as a value, etc. until our keys and values are swapped.

Therefore, an empty **value** after a conditional will trigger the exception that fails a parse on an empty **key**.

I don't think this is necessarily the best way to handle them, but at least it doesn't crash :smile: 

Do not merge (yet), I'm only posting this for discussion. How do we want to handle conditionals, and is this the best way?

I think I can see some implementation flaws in this code (i.e. the green part of the diff) but it's getting late and I need to go to bed so I'm stopping for tonight. I know it's not perfect.

### DIFF
--- a/SteamKit2/SteamKit2/Types/KeyValue.cs
+++ b/SteamKit2/SteamKit2/Types/KeyValue.cs
@@ -25,7 +25,7 @@ namespace SteamKit2
             : base( input )
         {
             bool wasQuoted;
-            bool wasConditional;
+            string condition;
 
             KeyValue currentKey = kv;
 
@@ -33,7 +33,7 @@ namespace SteamKit2
             {
                 // bool bAccepted = true;
 
-                string s = ReadToken( out wasQuoted, out wasConditional );
+                string s = ReadToken( out wasQuoted, out condition );
 
                 if ( string.IsNullOrEmpty( s ) )
                     break;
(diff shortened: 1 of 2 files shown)


xPaw:
Ignoring them is good enough for now, to be honest.

SteamKit2/SteamKit2/Types/KeyValue.cs
```diff
@@ -710,6 +719,13 @@ private void RecursiveSaveTextToFile( Stream stream, int indentLevel = 0 )
             WriteString( stream, "}\n" );
         }
 
+        static string FormatValueForWriting( KeyValue kv )
+        {
+            return kv.AsString()
+                .Replace("\r\n", @"\n")
+                .Replace("\n", @"\n");
```
xPaw: Is this a mistake? You're replacing `\n` with `\n`.

Netshroud: This replaces `<newline>` with `\n`. The `@` means an unescaped string :)

xPaw: Ah, yeah, that makes sense.

stil:
I made a simple fix for my personal needs. I might add pull request if you think it's useful. It could be sufficient until we have full support for conditionals.

It correctly loads up `csgo_english.txt` with all 11.7k tokens.

stil/SteamKit@7fcc2418a835c4fba865b8621286211026cd6770

yaakov-h:
Closing in favour of new draft PR.

PR #270, opened 2016-05-01
Add extra two conversions

JustArchi:
I like to keep memory usage of my SK2-based projects low.

It's pretty rare but sometimes key values returned by API call can be stored in 0-255 range or 0-65535. And it's even guaranteed according to Valve documentation.

It should be fine with you I guess :+1:. I kept naming convention just in case we'll want to add ```sbyte``` or ```short``` later.

### DIFF
--- a/SteamKit2/Tests/KeyValueFacts.cs
+++ b/SteamKit2/Tests/KeyValueFacts.cs
@@ -501,27 +501,27 @@ namespace Tests
         }
 
         [Fact]
-		public void KeyValuesUnsignedByteConversion()
+        public void KeyValuesUnsignedByteConversion()
         {
             byte expectedValue = 37;
 
-			var kv = new KeyValue("key", "37");
-			Assert.Equal(kv.AsUnsignedByte(), expectedValue);
+            var kv = new KeyValue("key", "37");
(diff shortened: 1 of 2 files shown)


Netshroud:
Sorry I missed this one, slipped under the radar.

Code itself looks good. Can you please add unit tests for this?

JustArchi:
Done I think - I'm not completely familiar with xunit yet, so sorry in advance if I didn't do something correctly :+1: 

Let me know if there is anything else you'd like to see.

SteamKit2/Tests/KeyValueFacts.cs
```diff
@@ -500,6 +500,30 @@ public void KeyValuesSavesTextToStream()
             Assert.Equal( expected, text );
         }
 
+        [Fact]
+        public void KeyValuesUnsignedByteConversion()
+        {
+            byte expectedValue = 37;
+
+            var kv = new KeyValue("key", "37");
+            Assert.Equal(kv.AsUnsignedByte(), expectedValue);
```
Netshroud: XUnit syntax is `Assert.Equal(expected, actual)`, not `Assert.Equal(actual, expected)`. Same on L512, L521 and L524.

JustArchi:
I'm sorry, done, thanks! :+1:

Issue #247, opened 2016-02-23
Builtin server list is out of date


rossengeorgiev:
Nearly half the servers on the list are not responding

```
nc: connect to 208.64.200.201 port 27017 (tcp) failed: Connection timed out
nc: connect to 208.64.200.201 port 27018 (tcp) failed: Connection timed out
nc: connect to 208.64.200.201 port 27019 (tcp) failed: Connection timed out
nc: connect to 208.64.200.201 port 27020 (tcp) failed: Connection timed out
nc: connect to 208.64.200.202 port 27017 (tcp) failed: Connection timed out
nc: connect to 208.64.200.202 port 27018 (tcp) failed: Connection timed out
nc: connect to 208.64.200.202 port 27019 (tcp) failed: Connection timed out
nc: connect to 208.64.200.203 port 27017 (tcp) failed: Connection timed out
nc: connect to 208.64.200.203 port 27018 (tcp) failed: Connection timed out
nc: connect to 208.64.200.203 port 27019 (tcp) failed: Connection timed out
nc: connect to 208.64.200.204 port 27017 (tcp) failed: Connection timed out
nc: connect to 208.64.200.204 port 27018 (tcp) failed: Connection timed out
nc: connect to 208.64.200.204 port 27019 (tcp) failed: Connection timed out
nc: connect to 208.64.200.205 port 27017 (tcp) failed: Connection timed out
nc: connect to 208.64.200.205 port 27018 (tcp) failed: Connection timed out
nc: connect to 208.64.200.205 port 27019 (tcp) failed: Connection timed out
Connection to 208.78.164.9 27017 port [tcp/*] succeeded!
Connection to 208.78.164.9 27018 port [tcp/*] succeeded!
Connection to 208.78.164.9 27019 port [tcp/*] succeeded!
Connection to 208.78.164.10 27017 port [tcp/*] succeeded!
Connection to 208.78.164.10 27018 port [tcp/*] succeeded!
Connection to 208.78.164.10 27019 port [tcp/*] succeeded!
Connection to 208.78.164.11 27017 port [tcp/*] succeeded!
Connection to 208.78.164.11 27018 port [tcp/*] succeeded!
Connection to 208.78.164.11 27019 port [tcp/*] succeeded!
Connection to 208.78.164.12 27017 port [tcp/*] succeeded!
Connection to 208.78.164.12 27018 port [tcp/*] succeeded!
Connection to 208.78.164.12 27019 port [tcp/*] succeeded!
[...]

PR #275, opened 2016-05-24
Removed dead CM IPs from builtin list

rossengeorgiev:
See #247


VoiDeD:
This will likely need to go hand in hand with [persistent-server-list](/SteamRE/SteamKit/tree/persistent-server-list).

Netshroud:
This looks to be superseded by #293.

PR #278, opened 2016-05-27
Correct LoginID documentation

DoctorMcKay:
Steam only cares about duplicate obfustucated_private_ips from the same
public IP, which makes sense given that it's to prevent duplicate
client connections from the same machine.


JustArchi:
:+1:

PR #285, opened 2016-06-05
Added KeyValue.AsDateTime()

ObsidianMinor:
Useful for converting from Unix time, which Steam uses in many places it seems.

### DIFF
--- a/SteamKit2/SteamKit2/Types/KeyValue.cs
+++ b/SteamKit2/SteamKit2/Types/KeyValue.cs
@@ -413,6 +413,16 @@ namespace SteamKit2
         }
 
         /// <summary>
+        /// Returns the value of this instance as a DateTime object.
+        /// Example: 1427581143 As DateTime creates 3/28/2015 12:00:00 AM
+        /// </summary>
+        /// <returns>The value of this instance as a DateTime object</returns>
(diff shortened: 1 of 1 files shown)


Netshroud:
Is Steam guaranteed to always use UTC?

I'd rather we use [DateTimeOffset.FromUnixTimeSeconds ](https://msdn.microsoft.com/en-us/library/system.datetimeoffset.fromunixtimeseconds(v=vs.110).aspx) but it's only available in .NET 4.6.

JustArchi:
Also it might be good idea to add optional parameter that defaults to ```DateTime.MinValue``` in case of conversion failure.

I think it's better than checking if DateTime of the year is 1970 (or equal to unix 0).

Netshroud:
Also, we already have UNIX-time logic in SteamKit. Re-use [DateUtils. DateTimeFromUnixTime](https://github.com/SteamRE/SteamKit/blob/32c1a6f3794ba61554a3a38c94e339068b2633d3/SteamKit2/SteamKit2/Util/Utils.cs#L201) rather than creating a second, parallel implementation.

ObsidianMinor:
Alright, I didn't notice that we had UNIX-time logic in SteamKit, I'll use that instead.

PR #288, opened 2016-06-09
Update protobuf download link

JustArchi:
The project has been moved to GitHub - while old link works for now and redirects into new place properly, it might stop working anytime and we should track current location instead.


VoiDeD:
Thanks!

PR #289, opened 2016-06-11
Added Java package information to the .proto files

dpeger:
In order to generate Java source based on the protocol buffers with protoc (or https://github.com/google/protobuf-gradle-plugin) additional Java options are needed.

### DIFF
--- a/Resources/Protobufs/csgo/base_gcmessages.proto
+++ b/Resources/Protobufs/csgo/base_gcmessages.proto
@@ -3,6 +3,9 @@ import "steammessages.proto";
 option optimize_for = SPEED;
 option cc_generic_services = false;
 
+option java_package = "org.opensteamworks.steamkit.base.proto.csgo";
+option java_multiple_files = true;
+
 enum EGCBaseMsg {
 	k_EMsgGCSystemMessage = 4001;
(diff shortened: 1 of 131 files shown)


Netshroud:
These protobufs are reverse-engineered to match as closely as possible the original protobufs that Valve fed into the Protobuf Compiler.

We will not be accepting manual modifications to these files.

dpeger:
So what would be the right place to add these modifications as non-manual steps?

Netshroud:
Either ask Valve to add them, or add a step to your own build pipeline to preprocess the proto files and insert those lines.

dpr-rec:
> ask Valve to add them

LOL, nice one. Did you guys receive much help from Valve implementing SteamKit?

However. Thanks for reviewing...

Netshroud:
I figure it's always worth a shot at the very least... I once asked Valve to make an internal interface public.

That didn't go so well.

dpr-rec:
And the protos have been provided by Valve?

Netshroud:
Yes... in the form of binary data inside of Steam/Source/Games/etc.

https://github.com/SteamRE/SteamKit/tree/master/Resources/ProtobufDumper/ProtobufDumper

PR #307, opened 2016-07-22
Tell Steam that we support the RateLimitExceeded logon response

DoctorMcKay:


Codingale:
What is the rate limit by the way I've apparently hit it.. just trying to login... woops.

PR #313, opened 2016-08-14
Extend default EClientPersonaStateFlag info

JustArchi:
Default value that steam client uses right now is equal to ```3154 = PlayerName, Presence, LastSeen, ClanTag, Facebook```. While I'm sure we don't really need ```ClanTag``` or ```Facebook``` info, ```LastSeen``` can be actually pretty useful, for example in my case when I'd like to remove friends that didn't log in since X time. This would allow me as well as every interested person to skip asking for that extra info in another request.

Since this request is default in steam client, and it results in generation of barely 2 extra fields (```LastLogOn``` and ```LastLogOff```), I think this should be added to default info.

### DIFF
--- a/SteamKit2/SteamKit2/Steam/Handlers/SteamFriends/SteamFriends.cs
+++ b/SteamKit2/SteamKit2/Steam/Handlers/SteamFriends/SteamFriends.cs
@@ -495,7 +495,8 @@ namespace SteamKit2
         // the default details to request in most situations
         const EClientPersonaStateFlag defaultInfoRequest =
             EClientPersonaStateFlag.PlayerName | EClientPersonaStateFlag.Presence |
-            EClientPersonaStateFlag.SourceID | EClientPersonaStateFlag.GameExtraInfo;
+            EClientPersonaStateFlag.SourceID | EClientPersonaStateFlag.GameExtraInfo |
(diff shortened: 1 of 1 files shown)


JustArchi:
Bump, I want to see this one in :+1:

JustArchi:
Monthly bump.

Netshroud:
Still not sure about this, it's a bunch of extra bandwidth that most clients won't make use of.

Maybe we should make the default flags configurable. @VoiDeD?

JustArchi:
Making default flags configurable sounds very good, but I don't think that asking for 2 fields in unix timestamp, so barely 64 bits more, would be that big bandwidth hit either. I thought of this being a good addition, as that info is also included in default Steam Client flags, and to be honest I was deeply shocked to see lack of that info in my default persona callbacks, I actually had to check what default flags are in SK2 as I was nearly sure that I'm doing something wrong.

So while both solutions make me happy, I'd still opt for accepting this change due to being Steam client standard + optionally making default flags configurable, so everyone can ask for whatever they want 👍. That's only my opinion though, you decide here.

VoiDeD:
Making this configurable could definitely be the long term solution here. I'm alright with adding `LastSeen`, this gives us more parity with the client.

PR #335, opened 2016-10-16
Add GSOCache types

Prosto-Sanja:
Added class for handling GSOCache(subscribe, desroy, etc) message types.
Needed for GC communication.

### DIFF
--- a/SteamKit2/SteamKit2/Base/Generated/GC/Dota/CacheSubscriptionTypes.cs
+++ b/SteamKit2/SteamKit2/Base/Generated/GC/Dota/CacheSubscriptionTypes.cs
@@ -1,0 +1,98 @@
+﻿namespace SteamKit2.GC.Dota.Internal
+{
+    /// <summary>
+    /// Cache types
+    /// </summary>
+    internal enum CacheSubscritionTypes
+    {
+        /// <summary>
+        /// An economy item.
+        /// </summary>
+        EconItem = 1,
+
+        /// <summary>
+        /// An econ item recipe.
+        /// </summary>
+        ItemRecipe = 5,
+
+        /// <summary>
+        /// Game account client for Econ.
+        /// </summary>
+        EconGameAccountClient = 7,
+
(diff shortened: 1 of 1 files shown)


psychonic:
This looks like a near-exact copy of [CacheTypes.cs](https://github.com/paralin/Dota2/blob/master/src/Dota2/Base/Data/CacheTypes.cs) from paralin's Dota 2 SteamKit extension. Is there a reason that you're not just using that library for something more Dota-specific?

Prosto-Sanja:
That library is dota specific and i understand it clearly. it helps to make communication with GC easier.

But this file should be in Steamkit project, as it is needed for barebones operation of already existing dota protobuffs, which are here.

rossengeorgiev:
SO types different on game to game bases. Granted they are mostly the same, but it can't be a single generic enum unfortunately. Just look at:

https://github.com/SteamRE/SteamKit/blob/master/Resources/NetHookAnalyzer2/NetHookAnalyzer2/Specializations/Dota2/Dota2SOHelper.cs
https://github.com/SteamRE/SteamKit/blob/master/Resources/NetHookAnalyzer2/NetHookAnalyzer2/Specializations/TF2/TF2SOHelper.cs
https://github.com/SteamRE/SteamKit/blob/master/Resources/NetHookAnalyzer2/NetHookAnalyzer2/Specializations/CSGO/CSGOSOHelper.cs

Prosto-Sanja:
But, maybe then we will include these files to steamkit project? As, again, they are needed for operating with GC. My First suggestion was based on dota, as I work primarily with it. Or are you completely against SOTypes in SteamKit?

psychonic:
As-is, we don't maintain any manually-reversed, game-specific data/classes in SteamKit itself. (NHA is more of an add-on). The only game-specific things in SK are the GC protobuf classes, which are programmatically dumped and generated.

Prosto-Sanja:
Ok, so i think you can close this request

PR #343, opened 2016-12-15
Update protobuf-net to v2.1.0

paralin:
Hi guys,

I really need this update asap because Protobuf.net does not support dotnetstandard until v2.1.0, and I cannot install both protobuf-net v2.1.0 and Steamkit2 in its current version at the same time.

I've bumped the dependencies and would greatly appreciate it if you could spin a release with this update.

Thanks!


paralin:
Un-bumped xunit, updated samples, should work

Samples/3.DebugLog/Sample3_DebugLog.csproj
```diff
@@ -37,6 +37,9 @@
     <Prefer32Bit>false</Prefer32Bit>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="protobuf-net, Version=2.1.0.0, Culture=neutral, PublicKeyToken=257b51d87d2e4d67, processorArchitecture=MSIL">
```
Netshroud: Why add a reference to a project that previously did not need it?

Samples/1.Logon/Sample1_Logon.csproj
```diff
@@ -37,6 +37,9 @@
     <Prefer32Bit>false</Prefer32Bit>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="protobuf-net, Version=2.1.0.0, Culture=neutral, PublicKeyToken=257b51d87d2e4d67, processorArchitecture=MSIL">
```
Netshroud: Why add a reference to a project that previously did not need it?

Samples/4.Friends/Sample4_Friends.csproj
```diff
@@ -39,6 +39,9 @@
     <Prefer32Bit>false</Prefer32Bit>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="protobuf-net, Version=2.1.0.0, Culture=neutral, PublicKeyToken=257b51d87d2e4d67, processorArchitecture=MSIL">
```
Netshroud: Same here, etc...

Netshroud:
What's the purpose of this? You've stated that:

>  Protobuf.net does not support dotnetstandard until v2.1.0, and I cannot install both protobuf-net v2.1.0 and Steamkit2 in its current version at the same time.

however, this just leaves me confused.

SteamKit2 does not support .NET Standard either, and likely will not until .NET Standard 2.0 early next year.

In the meantime, you won't be able to use SteamKit on .NET Core. We reference a lot of things that do not exist in Core, e.g. `Thread`, `SerializableAttribute`, `IsolatedStorage`, `Stream.Dispose()` and so on.

If you just want to use protobuf-net 2.1.0 in your own project on .NET Framework, rather add an Assembly Binding Redirect.

If you want to use SteamKit on .NET Core, you're going to be SOL anyway. I'm not sure what this pull request achieves.

JustArchi:
> If you want to use SteamKit on .NET Core, you're going to be SOL anyway. I'm not sure what this pull request achieves.

It makes SK2 one step closer to .NET Core, and updates dependencies to newer version, which might bring various visible and invisible improvements. We should be asking why **not** update, instead of why should we.

Netshroud:
@JustArchi I agree in principle. This PR is still good to have anyway... but I don't think it will have the desired effect.

JustArchi:
Totally agreed with you there @Netshroud, SK2 is not that easy library to bring up with .NET Core dependencies, and it's definitely not a good idea to try make it work with .NET Core at all cost. Release of V2 should make things much easier to port, and this is probably when SK2 should slowly try to make a move 👍

paralin:
@Netshroud whatever man. I realize it won't work with .Net standard now,
but keep your dependencies up to date!

No regens are required. Protogen has a newer version but I don't think it's
worth the effort to update it yet.

Netshroud:
Is there any value in updating half a dependency?

paralin:
????

You're updating a full dependency! protobuf-net.

Netshroud:
Protogen being the other 'half'.

Netshroud:
Turns out there is no newer version of protogen. 😅

Netshroud:
Closing as superceded by #375.

PR #354, opened 2017-01-12
Move duplicated dispatching code of ClientMsgHandlers to base class

stil:
I needed to write my own `ClientMsgHandler`, but I think it's not neccessary to override `HandleMsg` method every time. Usually it's enough to follow the the approach that uses Dictionary for action mapping.

I moved redundant code to abstract `ClientMsgMappingHandler` class. I think it allows for easier creating extension Handlers for communication not supported in SteamKit.

### DIFF
--- a/SteamKit2/SteamKit2/Steam/Handlers/ClientMsgMappingHandler.cs
+++ b/SteamKit2/SteamKit2/Steam/Handlers/ClientMsgMappingHandler.cs
@@ -1,0 +1,41 @@
+﻿/*
+ * This file is subject to the terms and conditions defined in
+ * file 'license.txt', which is part of this source code package.
+ */
+
+
+using System;
+using System.Collections.Generic;
+
+namespace SteamKit2
+{
+    /// <summary>
+    /// This class implements the base requirements every message handler should inherit from.
+    /// Provides simple message dispatching logic based on dispatch map.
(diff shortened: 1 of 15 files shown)


yaakov-h:
Originally I didn't like this proposal, but I'm not so sure now.

Will think it over and profile it, unless anyone else goes "no we need this merge it now".

Personally, I'd rather migrate everything back to switch statements - less memory overhead.

xPaw:
This seems like a good public API to have, since consumers have to implement their own mappings. But perhaps it to hide the implementation detail of `DispatchMap` and have a helper method to add/remove callbacks.

JustArchi:
I'm in with yaakov-h at this one, I don't really think adding dictionary there is that much beneficial over switch statements. I'd need to override `HandleMsg()` anyway for my own use case, as I'm doing there a bit more than just dispatching to appropriate function. Sure, dispatch map doesn't sound like a bad idea and in theory works better than switch statements in the open, but we need to implement handlers for every message anyway, which compared to one more function handling the type of message doesn't seem that much worse.

yaakov-h:
Closing as this hasn't been particularly compelling over the last two and a half years. (😅)

And two and a half years later I'd still rather use switch statements... or at least a static dictionary so the cost is not per handler instance.

Issue #324, opened 2016-09-22
Make default EClientPersonaStateFlag configurable


Netshroud:
See #313.

VoiDeD:
@yaakov-h Now that we've got a `SteamConfiguration`, this one should be fairly easy. 😄

yaakov-h:
Yep. This is next on my list. 😄

PR #363, opened 2017-02-11
Allow overriding webapi root

### LINKED ISSUE #352: Allow overriding WebAPI endpoint
If overwritten, it would also contain the protocol (making the secure option do nothing).

Use cases include forcing http for easier proxying, or using another domain (like community.steam-api.com).

xPaw:
Fixes #352.

### DIFF
--- a/SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs
+++ b/SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs
@@ -210,6 +210,14 @@ namespace SteamKit2
                 RegexOptions.Compiled | RegexOptions.IgnoreCase
             );
 
+            /// <summary>
+            /// Gets or sets the URI which is used to execute WebAPI requests.
+            /// When overriden, "secure" option is ignored. This also affects internal SteamKit requests.
+            /// </summary>
+            /// <value>
+            /// Endpoint URI which is used to execute WebAPI requests. For example, "http://steam-api.local/proxy".
+            /// </value>
+            public static string HttpEndpoint { get; set; }
 
             internal AsyncInterface( string iface, string apiKey )
             {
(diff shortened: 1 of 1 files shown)


SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs
```diff
@@ -210,6 +210,14 @@ public sealed class AsyncInterface : DynamicObject, IDisposable
...
             );
 
+            /// <summary>
+            /// Gets or sets the URI which is used to execute WebAPI requests.
+            /// When overriden, "secure" option is ignored. This also affects internal SteamKit requests.
+            /// </summary>
+            /// <value>
+            /// Endpoint URI which is used to execute WebAPI requests. For example, "http://steam-api.local/proxy".
+            /// </value>
+            public static string HttpEndpoint { get; set; }
```
xPaw: It's static to affect internal SK calls too.

VoiDeD: Would rather have it be non-static, don't want to introduce any potential for race conditions here.

xPaw: How do you propose allowing to override the endpoint for calls that SK executes? (e.g. SteamDirectory)

VoiDeD: Hm, what are your thoughts on adding properties to `SmartCMServerList` and having it pass it through to the `SteamDirectory` calls?

Are there any other places within SK that you need to override the endpoint? I believe `SteamDirectory` is likely the only outlier, but I might be wrong.

VoiDeD:
While we're here in WebAPI code, do you feel there's any downside to making `secure` default to `true`? It seemingly defaults to `false` but I don't recall why.

JustArchi:
I'm using ```secure: true``` in all my calls since a long time and I didn't spot any issues both Windows/Mono, just saying if it can help you make a choice. Also there are endpoints that care about protocol being used, I spotted ```ISteamUserAuth``` that does that for example.

Netshroud:
I see a few possible ways to do this:

1. The way that has been done, above, with a global static. I don't like this approach.
2. A similar approach using a different global mechanism, e.g. a `Switch` from `System.Diagnostics`, or `ConfigurationManager.AppSettings`.
3. A similar approach, but an immutable one - once set, you cannot change it.
4. Create some kind of immutable configuration object that gets passed into `WebAPI`. Code that consumes `WebAPI` directly can pass one in, otherwise you can pass one into `SteamClient` and it will propagate. Adding a configuration object would also let us take care of #324 at the same time.

What do you all think?

PR #364, opened 2017-02-12
Enabled automatic deployment from AppVeyor

Netshroud:
With these changes, AppVeyor will automatically publish a new release on GitHub and deploy to NuGet.

To use this, you need to push a tag on `master` that is a supported NuGet version number, e.g. `1.9.0` or `1.9.0-alpha`.

It also requires changes to appveyor.yml to set the new version number (both `version` and `assembly_version`), and you still have to manually update changes.txt and copy it into the GitHub release description.

This does still simplify the process so that anyone (with push privileges) can cut a new release with little effort.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,4 @@
-version: 1.0.{build}-Alpha
+version: 1.8.1.{build}
 os: Visual Studio 2015
 
 cache:
@@ -8,7 +8,9 @@ cache:
 assembly_info:
   patch: true
   file: AssemblyInfo.cs
-  assembly_informational_version: "{version} - CI (AppVeyor)"
+  assembly_version: "1.8.1"
+  assembly_file_version: "{version}"
+  assembly_informational_version: "{version} - CI (AppVeyor, branch: {branch})"
 
 configuration:
   - Debug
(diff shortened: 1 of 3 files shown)


VoiDeD:
This looks good to me. Can we commit this so that this merge only introduces automatic deployment, and then a separate commit pushes out 1.8.1? Just so the git log is more clear.

appveyor.yml
```diff
@@ -1,10 +1,17 @@
...
 os: Visual Studio 2015
 
 cache:
   - '%LocalAppData%\NuGet\Cache'
   - '%Temp%\nethook2-dependencies -> Resources\NetHook2\SetupDependencies.ps1'
 
+assembly_info:
+  patch: true
+  file: AssemblyInfo.cs
+  assembly_version: "1.8.1"
```
VoiDeD: If my understanding is correct and this patches our AssemblyInfo.cs, could we write a comment there that it's automatically handled by CI for deployments.

Netshroud: Sounds sensible. I'll do that once Visual Studio finishes installing.

xPaw: Why aren't you using `{version}` here?

xPaw:
🎉

JustArchi:
Thanks a lot for the idea, I did exactly the same with my ASF project - this simplifies things a lot! 🎉

Netshroud:
@JustArchi I stole the idea from xPaw in the first place... good ideas are contagious. 😄

PR #366, opened 2017-02-25
[NHA2] Add support for showing difference between set fields and implied fields for proto msgs.

psychonic:
For some protobuf message, there is a difference between how Valve handles them, particularly for some GC messages, for whether a field is set at all versus a field set to its default value.

NetHookAnalyzer2 makes no distinction here, always showing all fields it knows about as set, and showing the default value if they are not. This changes that to instead all like this.

Default - Fields that aren't set are hidden:
Default

Show All - All fields are shown, with their effective values, but still having distinction between what was set and what was not.
ShowAll

This does require altering our protogen options to generate a second property for each field, a bool that gets named "<field name>Specified". It's possible that the extra field could be a breaking change for third-parties, but it seems unlikely. It would have caused much extra clutter in NetHookAnalyzer, but not after part of the included changes here.

### DIFF
--- a/Resources/NetHookAnalyzer2/NetHookAnalyzer2/MainForm.Designer.cs
+++ b/Resources/NetHookAnalyzer2/NetHookAnalyzer2/MainForm.Designer.cs
@@ -192,6 +192,8 @@
 			// showAllCheckBox
 			// 
 			this.showAllCheckBox.AutoSize = true;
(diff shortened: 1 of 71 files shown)


psychonic:
The new checkbox is now checked (showing all) by default per @VoiDeD

PR #375, opened 2017-03-14
Add .NET Core support

### LINKED ISSUE #299: .NET Core Support
TODO: Investigate value and feasibility.

Looks like protobuf-net 2.1.x(-alpha) supports .NET Core.

Netshroud:
This PR:
* Converts csproj files to .NET Core SDK
* Bumps .NET Framework requirement to .NET 4.6
* Adds .NET Standard 1.3 output
* Updates protobuf-net dependency to 2.1 (supercedes #343).
* Converts Samples to dual-target .NET 4.6, .NET Core 1.1
* Converts tests to dual-target .NET 4.6, .NET Core 1.1. All tests pass on Windows (.NET, Core), macOS (Core) and Linux (Core).
* Replace use of WMI with P/Invoke calls to `DeviceIoControl` as `System.Management` is not available in the new SDK, even for .NET Framework. This also works on .NET Core on Windows.
* Updates `CDNClient` and `WebAPI` to use `System.Net.Http.HttpClient` instead of `HttpWebRequest`/`System.Net.WebClient`.
* Changes API surface of `CDNClient` to be asynchronous with tasks.
* Updates AppVeyor and Travis builds to use .NET Core. Travis only builds .NET Standard 1.3, AppVeyor builds both .NET 4.6 and .NET Standard 1.3, and tests on both.
* Updates NetHook2 to build with VS2017, latest Windows SDK ("Anniversary Edition").

Resolves #299.


VoiDeD:
@psychonic @azuisleet `CDNClient` is now all async, any potential impact here on DepotDownloader or is this fine?

@Netshroud Since we've converting all our csprojs to the new format, is there any way to retain VS2015 support? Not sure if this is really a viable thing to do, but 2017 did essentially _just_ come out, and I'm wondering if we could have a way to maintain support for 2015 at least for a while as people may still be transitioning.

Netshroud:
DepotDownloader would at minimum need to be adjusted to consume the async APIs, could potentially be converted to .NET Core as well... 😈 

@VoiDeD The .NET Core SDK RTM is only available with VS2017. There are preview builds for VS2015 but I have no idea how well they work, and I do believe the standalone bits (`dotnet.exe` and so on) are available separately for Windows so you could use the command line (like on macOS/Linux) but then lose IDE support.

On the bright side, VS2017 Community is free and installs seamlessly side-by-side with VS2015.

Regarding NetHook2, I believe the new binary will need a new version of the Microsoft C++ Runtime Redistributable Bits And Bobs. It Works On My Machine with the SDK installed, I haven't tried it on a clean machine / in a VM yet.

shravan2x:
Commit 645b91a should follow https://github.com/dotnet/corefx/issues/3324.

What does Steam use this ID for? Can Steam work without a disk ID? Can we just generate a random one and store it somewhere?

Netshroud:
@shravan2x in theory any value will do, however I don't want to break compatibility with existing consumers. Steam must see a machine the same way after a SteamKit upgrade, otherwise we risk our consumers being locked out of trade etc. due to Steam Guard restrictions.

JustArchi:
How is that P/Invoke call working on Mono then? There is no way we can port `kernel32.dll` call to Linux.

Netshroud:
@JustArchi technically, Microsoft already ported kernel32.dll to Linux for SQL Server, but let's not go there.

On Mono [we go a different way](https://github.com/SteamRE/SteamKit/blob/netcore-for-realsies/SteamKit2/SteamKit2/Util/HardwareUtils.cs#L47) and don't hit this P/Invoke call.

Netshroud:
I just need to finalize DepotDownloader, then I'll merge this in.

xPaw:
What about making this a major version update and drop all the obsolete and deprecated stuff? Seems like a good time for a cleanup.

Netshroud:
I was considering that for a 1.9.0 release - all except `TradeProposedCallback.OtherName`, I think, since that was only obsoleted after 1.8.2.

xPaw:
You can make a 1.8.3 release with the deprecation before merging this branch.

shravan2x:
Is think adding .NET Core support is worthy of 2.0. I don't see any other major changes around the corner and wasting netcore on 1.9 seems sad.

Netshroud:
If we followed semantic versioning we'd be on version 15.y.z or thereabouts.

JustArchi:
Nobody said that 1.10 must result in 2.0. I'd also say that 2.0 is good for .NET core support, but I can also totally understand it being "just" 1.9, because after all code-wise nothing is changing that much, so it doesn't bring anything new for people that are targetting .NET 4.5+.

Still, versioning is something misc, there is no need to discuss about that - just let developer decide himself how important his changes are, in this case @Netshroud  🙂 .

(And thanks a lot for this, in case I forgot to show my gratitude)

JustArchi:
Not sure if somebody is interested in it, but I've moved one of my projects (ArchiBoT) to .NET core 1.1 and latest alpha 3 of SK2 works excellent with it (apart from #383). I couldn't find any other issue, and my bot does make use of a lot of SK2 issues, so it's a good testing subject 🙂 .

Interesting fact: I couldn't establish connection with Steam when using not-yet-released .NET core 2.0. It could be because of the net core itself being in still somehow unstable development stage, as I got NRE somewhere in internal SK2 code, so no need to sound an alarm, especially because it's not released yet. I'll keep an eye on it and see if things improve over time with new .NET core builds getting released.

@Netshroud do you plan on moving to .NET core 2.0 (standard in your case, as SK2 is a library) when its released? I'm pretty sure it will be breaking change, but I also realize there are a few things that .NET core doesn't implement yet in current standard (1.6), while it is wanted in SK2 - I think isolated storage and generating device ID was in it. I'm asking because I'm not sure if you're going to jump on new version and deprecate current one the moment its released, or you're going to somehow support old one for time being.

Netshroud:
@JustArchi I'd like to move a couple APIs back to how they were pre-this-PR, which requires .NET Standard 2.0.

Not sure yet what we'll do, staying as-is will be fine if we're stable, otherwise multi-targeting is always an option.

If we do switch, we'll probably have a single multi-targeted migratory release.

JustArchi:
BTW, that connection problem I was talking about turned out to be something entirely different, it looks like SK2 is perfectly compatible with 2.0 at first sight, although I didn't test everything.

@Netshroud Could you release another 2.0 alpha with https://github.com/SteamRE/SteamKit/commit/1c3757a0144bb2ace0e1b446f29b9bf0883a9815 fix in? It's important, as it would allow me to further test 2.0 on wider range of target machines, perhaps finding out more issues like #383 from which SK2 could benefit in long-run before it hits stable release 👍

PR #385, opened 2017-04-15
Update readme.md

JustArchi:
Mono 4.8 required for using SK 2.0+, I couldn't connect with 4.6 or lower.
Mono 5.0 required for compilation - it supports C# 7.0.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -39,11 +39,11 @@ SteamKit2 is released under the [LGPL-2.1 license](http://www.tldrlegal.com/lice
 
 In order to use SteamKit2 at runtime, the following dependencies are required:
 
-  - .NET 4.6, .NET Core 1.0, or [Mono ≥4.0](http://mono-project.com)
+  - .NET 4.6, .NET Core 1.0, or [Mono ≥4.8](http://mono-project.com)
 
 To compile SteamKit2, the following is required:
 
-  - C# 7.0 compiler &mdash; [Visual Studio 2017](https://www.visualstudio.com/vs/whatsnew/) or [.NET Core 1.1](https://www.microsoft.com/net/core)
+  - C# 7.0 compiler &mdash; [Visual Studio 2017](https://www.visualstudio.com/vs/whatsnew/) or [.NET Core 1.1](https://www.microsoft.com/net/core) or [Mono ≥5.0](http://mono-project.com)
   - [protobuf-net](http://code.google.com/p/protobuf-net/) ([NuGet package](http://nuget.org/packages/protobuf-net))
 
 Note: If you're using the NuGet package, the protobuf-net dependency _should_ be resolved for you. See the [Installation Guide](https://github.com/SteamRE/SteamKit/wiki/Installation) for more information.


Netshroud:
Mono 5.0 is not released yet.

What happened on 4.6? My understanding is that 3.x some time added .NET 4.6 runtime support, but I can't find the exact version. 4.0 provided the ability to build C# 6.0 (which shipped alongside .NET 4.6), so Mono used that as a way to signifify .NET 4.6 compatibility.

There's nothing in the 4.8 release notes to indicate any additional runtime / target framework support.

JustArchi:
> Mono 5.0 is not released yet.

I know, but it will ship with roslyn compiler, which is capable of compiling C# 7.0 perfectly fine. So I think that including future release in what is capable of compiling SK2 would not be a huge issue - especially because roslyn compiler is being used also in .NET and .NET core now.

> What happened on 4.6?

I'm too lazy to dig into outdated versions since I don't use them neither recommend to people, but I got many confirmed reports that SK in version 2.0+ doesn't work on Mono 4.6 or lower - first version of Mono that works fine with SK 2.0+ is Mono 4.8.

Because you plan on releasing version 2.0+ sooner or later, I think it'd be a good idea to update README to match a Mono version that is confirmed to be **working** with SK2, and not the one that doesn't work, or works only partially until you hit runtime bug.

The exact issue of SK2 not working properly on Mono older than 4.8 can be noticed e.g. in https://github.com/JustArchi/ArchiSteamFarm/issues/529. It was not possible to establish a connection on Mono 4.6 when using SK2 in version 2.0+. Older SK2 in version 1.8.3 worked just fine. I don't think that it's worth development effort to see what has changed in SK2/Mono to cause incompatibility, **especially** because latest stable Mono in version 4.8 works just fine with all SK2 versions, so it'd be a good idea to state that officially - we don't want somebody with Mono 4.2 to struggle with not being capable to connect to Steam only because he downloaded SK2 in version 2.0+ and he tries to run it on Mono older than 4.8 - SK2 states currently that any Mono 4.0 or above should work just fine, which is not true.

Mono in general changes dramatically, I started from requiring version 4.2 in my ASF as initial one, then I had to bump it to 4.4, then to 4.6, and now I'm at 4.8. First bump was caused by by TLS/SSL issues when using HTTPS protocol, I can't even find the related bug report, second bump was caused by **[this](https://bugzilla.xamarin.com/show_bug.cgi?id=41701)**, last bump was caused by SK2 not working properly on <= 4.6 (the reason I'm sending this PR). There is a lot of **[other bugs](https://bugzilla.xamarin.com/show_bug.cgi?id=42606)** that I reported across last several months, I don't really want to dig too deep into the bugzilla, but I can sadly state that there is just too many critical fixes required for various third-party libraries and assemblies that running outdated Mono is being directly asking for problems.

And I do believe that this change could simply save hours of struggling for newcomers trying to make use of SK2. In fact, I'm not even doing this PR for myself, one of my friends tried to run SK2 in version 2.0+ with Mono 4.2 and it didn't work, so he asked me for advice, and I immediately recognized my old ASF issue, then checked out what SK2 requires/recommends, then made a PR to correct it.

Personally I think that it's much better to answer "why Mono 4.8 is required", pointing to appropriate issue, rather than making somebody waste several hours and then answering "why Mono 4.2 doesn't work". I didn't want to bump required/recommended Mono version for my own project either, but I had to, Mono is not as stable as people think it is.

Netshroud:
I still have to get a CI test matrix for netcore and mono before 2.0 release, currently we only have .NET Core on Linux and .NET Core + .NET Framework on Windows.

I expect most consumers would use one of two types of Mono:
* Latest stable
* The version that ships with their OS (i.e. available by default in their package manager).

If you require a higher version than we do then I'm fine with that, but someone else might still need 4.5 or 4.6 or whatever, considering that [major versions of Ubuntu are tied to specific minor releases of Mono](http://packages.ubuntu.com/search?keywords=mono-complete), for example.

We probably need a project that logs into Steam, does some basic operation(s) and disconnects - similar to our Samples - that we can run on the CI servers to ensure that functionally, everything works to a bare minimum.

JustArchi:
> If you require a higher version than we do then I'm fine with that, but someone else might still need 4.5 or 4.6 or whatever, considering that major versions of Ubuntu are tied to specific minor releases of Mono, for example.

Except when he finds out after 2+ hours of struggling that the Mono version shipped with his Ubuntu doesn't work with SK2, because it's in version 4.6 that SK2 in version 2.0+ doesn't work with, and nobody told him about that - this has happened to my friend (hence, this PR).

I could change wording to "recommended", or we could put up a note somewhere else which Mono is confirmed to be working, but I'm not sure if that's what we want - I thought that bumping *required* version of Mono, to a first version that is confirmed to be *working* would be a good idea, since technically it's required if you want to use anything that the library offers, logging in is just a basic example.

Yes, adding Mono to CI build matrix and some bare functionality tests would be a best long-term solution here, but it doesn't conflict with a small note in the readme, to avoid people trouble. After all readme should be helpful - I really doubt that proposed change would do any harm to anybody - it's a mere indicator that is supposed to save time.

Netshroud:
And there are likely also critical bugs in 4.8 and 5.0 and 5.1 and so on that will mean in another 6-12 months we can't recommend those either.

I don't mean to sound dismissive of Mono and this is personal opinion only, but I don't think it's really our concern that Mono is a buggy pile of crap year in, year out. It works well enough to do some stuff, but I always hit strange edge cases that make it unworkable, and it's not just me - you have it above, and I know SteamDB have had issues with HTTP and threading too.

That's one of the biggest reasons I've been watching .NET Core and driving support for it here - it may be new, but it works a hell of a lot better in my experience, since day one.

I'm somewhat tempted to mark Mono support as "eh it probably works", but there's no way I can drop it for "you're on your own" due to backwards compatibility, which is what I'd really like to do.

JustArchi:
Closing the pr as I'm not interested in it anymore, I switched to .NET core and quality of my life improved drastically 😆.

PR #390, opened 2017-05-11
Remove dependencies on System.Diagnostics.Process and System.Threading.Thread

stackia:
System.Diagnostics.Process and System.Threading.Thread don't exist on UWP.
After these two modifications, I have tested SK2 now works with UWP.

### DIFF
--- a/SteamKit2/SteamKit2/Networking/Steam3/TcpConnection.cs
+++ b/SteamKit2/SteamKit2/Networking/Steam3/TcpConnection.cs
@@ -20,7 +20,7 @@ namespace SteamKit2
         private IPEndPoint destination;
         private Socket socket;
         private INetFilterEncryption netFilter;
-        private Thread netThread;
+        private Task netTask;
         private NetworkStream netStream;
         private BinaryReader netReader;
         private BinaryWriter netWriter;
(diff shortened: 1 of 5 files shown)


Netshroud:
I'm hesitant to make such a change to threading - I looked at this exact change a couple times in initial ports, but it turns out that it was unneeded.

Officially, `System.Threading` is supported on .NET Standard 1.3 and higher (exactly what we target), but from what I've read regarding UWP it only actually works on .NET Standard 1.4 and higher.

The JobID change I'm not thrilled about either - particularly, complete removal of a unit test.

What happens when you run this on UWP, and which version of UWP are you targeting?

stackia:
I'm targeting the latest UWP 15063.

netstandard does have `System.Threading` namespace but doesn't have `System.Threading.Thread` until 2.0. Actually `System.Threading.Thread` isn't available in UWP now.

`System.Diagnostics.Process` nuget package doesn't work with UWP, although it can be installed via nuget.
Sorry for the removal of the unit test cause I don't know how to write a replacement without `System.Diagnostics.Process`.

BTW, netstandard 2.0 is coming soon and UWP is going to adapt to netstandard 2.0 later this year. Both `System.Threading.Thread` and `System.Diagnostics.Process` exist in netstandard 2.0, so it may not be that urgent to fix this problem now 😃.

Netshroud:
Yeah, if UWP vNext complies with `netstandard2.0` then we can probably just retarget that... or you might be able to use the current build of SteamKit on UWP vNext without even recompiling. 😄

stackia:
I will close this PR. Anyone who want to use SK2 in their UWP app can reference the temporary solution provided here.

Issue #400, opened 2017-06-17
NRE in new SK2 2.0 networking code


JustArchi:
```
2017-06-16 15:36:05|mono-sgen-1935|FATAL|ASF|UnobservedTaskExceptionHandler() System.AggregateException: A Task's exception(s) were not observed either by Waiting on the Task or accessing its Exception property. As a result, the unobserved exception was rethrown by the finalizer thread. ---> System.NullReferenceException: Object reference not set to an instance of an object
  at SteamKit2.TcpConnection.TryConnect (System.Object sender) [0x00094] in <f740b1bfddfa4ddb95d88fffe3231fb2>:0 
  at SteamKit2.TcpConnection+<>c__DisplayClass19_0.<Connect>b__0 (System.Threading.Tasks.Task`1[TResult] t) [0x000b2] in <f740b1bfddfa4ddb95d88fffe3231fb2>:0 
  at System.Threading.Tasks.ContinuationTaskFromResultTask`1[TAntecedentResult].InnerInvoke () [0x00024] in <73ee1b14c2fa4d61b481096ff3d8d6d7>:0 
  at System.Threading.Tasks.Task.Execute () [0x00010] in <73ee1b14c2fa4d61b481096ff3d8d6d7>:0 
   --- End of inner exception stack trace ---
---> (Inner Exception #0) System.NullReferenceException: Object reference not set to an instance of an object
  at SteamKit2.TcpConnection.TryConnect (System.Object sender) [0x00094] in <f740b1bfddfa4ddb95d88fffe3231fb2>:0 
  at SteamKit2.TcpConnection+<>c__DisplayClass19_0.<Connect>b__0 (System.Threading.Tasks.Task`1[TResult] t) [0x000b2] in <f740b1bfddfa4ddb95d88fffe3231fb2>:0 
  at System.Threading.Tasks.ContinuationTaskFromResultTask`1[TAntecedentResult].InnerInvoke () [0x00024] in <73ee1b14c2fa4d61b481096ff3d8d6d7>:0 
  at System.Threading.Tasks.Task.Execute () [0x00010] in <73ee1b14c2fa4d61b481096ff3d8d6d7>:0 <---
```

[...]

PR #402, opened 2017-06-22
[WIP] Started refactoring TCPConnection

azuisleet:
Here's the justification for these changes:

The goal here is to simplify some of the rather ugly locking. There are two necessary locks: NetLock and ReleaseLock. 

Anything that can concurrently access a socket or stream needs to hold NetLock so that they don't disappear underneath them. The call to CleanupSocket locks and releases these objects. NetLock is held by GetLocalIP which inspects the socket and Send which competes for the ability to write to the socket.

The Disposable objects like the ManualResetEvent and CancellationToken have a slightly different lifetime. If something is actively waiting on them, it acquires ReleaseLock such that the final PostRelease call won't dispose them until this lock is given up. It's only important that these aren't Disposed underneath because:

The responsibility of signalling the CancellationToken and waiting for the DisconnectedCallback to be dispatch is given to the winning thread that calls Disconnect and wins the Interlocked.Exchange.

This should remove any ambiguity around the lifetime of objects re #400, as well as ensuring that if you call Disconnect while Connected, the DisconnectedCallback is on the queue (need to find my notes on why this was not the case). 

Concurrent Connect calls still have a connectLock so that a bunch of connections aren't opened only to be immediately destroyed.

After some thought, I'm also including connectLock in Disconnect so that you can't call Disconnect (exchanging with null) and then calling Connect without first letting the owning Disconnect call complete. Even if these calls are out of order globally, we should still enforce the order of callbacks.

### DIFF
--- a/SteamKit2/SteamKit2/Networking/Steam3/TcpConnection.cs
+++ b/SteamKit2/SteamKit2/Networking/Steam3/TcpConnection.cs
@@ -15,166 +15,223 @@ namespace SteamKit2
 {
     class TcpConnection : Connection
     {
+        internal class ConnectionState
+        {
+            public IPEndPoint Destination { get; private set; }
+            public Socket Socket { get; private set; }
+            public INetFilterEncryption NetFilter { get; private set; }
+            public Thread NetThread { get; private set; }
+            public NetworkStream NetStream { get; private set; }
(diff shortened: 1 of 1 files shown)


JustArchi:
Just wanted to say thank you for taking your time to work on this, I appreciate it a lot 💓.

PR #405, opened 2017-06-24
Add websocket client support

Netshroud:
Prototype branch, do not merge [yet]. Discuss here.

---

1. I'm not really liking the current approach to bypassing the traditional crypto handshake in order to get the `ConnectedCallback` posted. The best option I can think of is that we may have to redesign the `Connection` suite of classes.

Currently we have:

`Connection` --> `<Tcp/Udp>Connection` --> `CMClient` -> `SteamClient`

* `CMClient` is responsible for setting up the connection, including encryption, and the primitive state machine surrounding the crypto handshake. This includes the `EUniverse` of the connection.
* `SteamClient` is responsible for posting the `ConnectedCallback` and `DisconnectedCallback`.

Since WebSocket has no crypto handshake, this mechanism needs to be bypassed. I'm thinking we could possibly get around it by splitting it up like so:

* `IConnection` --> `<Tcp/UdpConnection>` --> `EnvelopeEncryptedConnection` --> `CMClient` --> `SteamClient`
* `IConnection` --> `WebSocketConnection` --> `CMClient` --> `SteamClient`

whereby all of the traditional  encryption logic moves to `EnvelopeEncryptedConnection`.

The rough outline looks like this:

```
class EnvelopeEncryptedConnection : IConnection
{
    public EnvelopeEncryptedConnection(IConnection connection)
    {
        innerConnection = connection;
        innerConnection.OnRecieveRawMsg += DecryptAndForwardMsg;
    }

    public void SendRaw(byte[] data) => innerConnection.Send(Encrypt(data));

    public void DecryptAndForwardMsg(byte[] data)
    {
        if (IsCryptoInitialized()) {
            HandleIncomingHandshakeMsg(data);
        } else {
            this.OnRecieveRawMsg?.Invoke(Decrypt(data));
        }
    }

    // Other methods are proxied to innerConnection
}
```

2. I'm also not sure how we should handle `EUniverse` for websocket, since the server does not tell us what universe we are. It looks like the Steam client knows the universe of the server before it connects... we could do some changes so that consumers can specify a universe, or we can just hardcode `EUniverse.Public`. In practice this only really matters if anyone uses this library inside of Valve... do they?

3. I'm thinking maybe make ProtocolType a flags enum, then return the best server that matches one or more bits, so `CMConnectionType.Socket`translates to `ProtocolType.Tcp | ProtocolType.Udp`. Then we'd need to create a Connection based on the server type, rather than picking a server based on the connection type.

TODO:
- [x] Make it work
- [x] Make it fast
- [ ] Make it pretty
  - [x] Refactor connection/encryption
    - [x] Move encryption logic from `CMClient`/`SteamClient` to `EnvelopeEncryptedConnection`
    - [x] Don't require RSA/AES crypto handshake in order to appear connected
    - [x] Unify ConnectionType and ProtocolType
    - [x] Create/select `IConnection` based on consumer-chosen `ProtocolTypes`
  - [ ] Sort out EUniverse / SteamClient construction
    - [x] Refactor settings into configuration object
    - [ ] Make configuration immutable after consuming
  - [ ] anything else?


ObsidianMinor:
> I'm also not sure how we should handle EUniverse for websocket, since the server does not tell us what universe we are

I handled this in [my implementation](https://github.com/ObsidianMinor/SteamStandard/blob/master/src/Network/Steam.Net/SteamNetworkConfig.cs#L40) by adding a "DefaultUniverse" option in a client config. You could implement it by adding an EUniverse ctor parameter to CMClient and SteamClient that defaults to EUniverse.Public.

VoiDeD:
Pretty ideal that 2.0 is still an alpha, gives us a chance to do any refactoring needed for websockets so this work can be a part of 2.0 as well.

>I'm not really liking the current approach to bypassing the traditional crypto handshake in order to get the ConnectedCallback posted. The best option I can think of is that we may have to redesign the Connection suite of classes.

I actually like the idea of moving out the encryption handshake out of CMClient and into a `EnvelopeEncryptedConnection`. It definitely feels like the handshake should be more of a transport-level concept, rather than logic the CMClient itself needs to perform. Unfortunate that the handshake is performed with `Msg<T>` classes that the `CMClient` is better at working with, but this might be the right approach.

>In practice this only really matters if anyone uses this library inside of Valve... do they?

_Potentially?_ In the past we used to make some assumptions about which universe we were connected to, and I recall hearing that a certain Valve needed to make some tweaks to the library to make it work. Not a super great argument for designing this, but is there really no way to know which universe we've connected to for a websocket CM? If not, I still think it would be better to allow the ability to handle non-public universes.

>I'm thinking maybe make ProtocolType a flags enum, then return the best server that matches one or more bits, so CMConnectionType.Sockettranslates to ProtocolType.Tcp | ProtocolType.Udp. Then we'd need to create a Connection based on the server type, rather than picking a server based on the connection type.

Picking a server based on server type definitely sounds more ideal, but we should also retain the ability to select either tcp or udp connections for non-websocket CMs.

Netshroud:
I've split the connection classes up as described... the encrypted connection has to call `CMClient.GetPacketMsg`, but perhaps that method itself should be moved off of `CMClient`.

If we allow consumers to define the universe (with a default of `EUniverse.Public`), that should also tie in with #352 - particularly, we'd want to connect to a different Dev/Beta Web API in order to retrieve the Dev/Beta CM list.

I'm thinking perhaps we should have a configuration object that gets passed into the `SteamClient` constructor. This would let us select protocol types, universe, WebAPI root, and possibly also be the point where the server list provider and/or debug network listener get defined.

yaakov-h:
OK, I think I've fixed the last of the thread safety issues.

Comments and critique welcome. If there are no further issues I plan to merge this late next week.

I'll defer the immutable configuration until I can figure out a clean way of doing it - I'm not sure I like using the builder pattern here, and want some time to pick at alternatives.

SteamKit2/SteamKit2/Steam/CMClient.cs
```diff
@@ -332,6 +330,24 @@ protected virtual void OnClientDisconnected( bool userInitiated )
             }
         }
 
+        IConnection CreateConnection( ProtocolTypes protocol )
+        {
+            if ( ( protocol & ProtocolTypes.WebSocket ) > 0 )
```
VoiDeD: `HasFlag` for these checks as well.

SteamKit2/SteamKit2/Steam/SteamClient/SteamConfiguration.cs
```diff
@@ -0,0 +1,85 @@
...
+
+using System;
+using System.Threading;
+using SteamKit2.Discovery;
+
+namespace SteamKit2
+{
+    /// <summary>
+    /// Configuration object to use.
+    /// This object should not be mutated after it is passed to one or more <see cref="SteamClient"/> objects.
```
VoiDeD: >I'll defer the immutable configuration until I can figure out a clean way of doing it
Guess it helps if I read. :)

SteamKit2/SteamKit2/Networking/Steam3/WebSocketConnection.cs
```diff
@@ -0,0 +1,226 @@
...
+        public void Disconnect()
+            => DisconnectCore(userInitiated: true, specificContext: null);
+
+        public IPAddress GetLocalIP() => IPAddress.None;
+
+        public void Send(byte[] data)
+        {
+            try
+            {
+                currentContext?.SendAsync(data).GetAwaiter().GetResult();
```
yaakov-h: This doesn't do anything with deadlocks, it just throws the actual exception instead of AggregateException - i.e. it saves me having to catch it, then rethrow the inner exception with `ExceptionDispatchInfo`.

SteamKit2/SteamKit2/Steam/Discovery/SmartCMServerList.cs
```diff
@@ -264,7 +233,8 @@ private IPEndPoint GetNextServerCandidateInternal()
                 ResetOldScores();
 
                 var serverInfo = servers
-                    .Select( (s, index) => new { EndPoint = s.EndPoint, IsBad = s.LastBadConnectionDateTimeUtc.HasValue, Index = index } )
+                    .Where( s => (s.Record.ProtocolTypes & supportedProtocolTypes) > 0 )
```
yaakov-h: `HasFlag` is a slower check at runtime, at least on .NET Framework. How do you feel about an extension method, so we have both clarity and speed?

VoiDeD: Is this a hot path? An extension method is probably fine, not sure if it makes much of a perf difference between that and `HasFlag`, and if it's worth it. :)

JustArchi: You're already using LINQ, `HasFlag` has negligible performance hit. Sure there is type check and boxing/unboxing (everything is casted to `ulong` IIRC), but this is still negligible compared to huge advantage of clean code. This is just my personal opinion though.

yaakov-h: Both?

yaakov-h: But yeah. Not overly concerned about perf on this one. I'm more concerned about CMClient and server list.

Samples/1.Logon/Program.cs
```diff
@@ -76,14 +76,6 @@ static void Main( string[] args )
 
         static void OnConnected( SteamClient.ConnectedCallback callback )
         {
-            if ( callback.Result != EResult.OK )
```
xPaw: Why did you remove these?

SteamKit2/SteamKit2/Steam/Discovery/SmartCMServerList.cs
```diff
@@ -274,59 +244,61 @@ private IPEndPoint GetNextServerCandidateInternal()
...
-            return GetNextServerCandidateInternal();
+            return GetNextServerCandidateInternal( supportedProtocolTypes );
         }
 
         /// <summary>
         /// Gets the <see cref="System.Net.IPEndPoint"/>s of all servers in the server list.
         /// </summary>
         /// <returns>An <see cref="T:System.Net.IPEndPoint[]"/> array contains the <see cref="System.Net.IPEndPoint"/>s of the servers in the list</returns>
-        public IPEndPoint[] GetAllEndPoints()
+        public CMServerRecord[] GetAllEndPoints()
```
xPaw: Would `IList` be a better fit here?

JustArchi: If we go with interfaces then `ICollection` would be much better fit, endpoints not necessarily have to be stored in a list, `HashSet` also works perfectly with them.

Samples/1.Logon/Program.cs
yaakov-h: Yup. We only ever triggered the callback with EResult.OK, so it’s effectively redundant.

SteamKit2/SteamKit2/Steam/Discovery/SmartCMServerList.cs
yaakov-h: IReadOnlyList would be ideal, but ICollection gets optimised in LINQ.

I don’t see any issue with Array here, as it’s a copy of the internal data, not a mutable handle.

PR #407, opened 2017-06-25
Move from multi-target to only .NET Standard 2.0

Netshroud:
PR for discussion / issue-tracker stats.

Do not merge until .NET Core 2.0 / .NET Standard 2.0 and appropriate SDK is released.


JustArchi:
> Unable to resolve 'Microsoft.NETCore.App (>= 2.0.0)' for '.NETCoreApp,Version=v2.0'.

> /usr/share/dotnet/sdk/1.0.1/NuGet.targets(97,5): error : Unable to resolve 'Microsoft.NETCore.App (>= 2.0.0)' for '.NETCoreApp,Version=v2.0'. [/home/travis/build/SteamRE/SteamKit/SteamKit2/SteamKit2.sln]

https://github.com/JustArchi/ArchiSteamFarm/blob/netcore/appveyor.yml#L8

https://github.com/JustArchi/ArchiSteamFarm/blob/netcore/.travis.yml#L20

Just in case you'd want to make CI work with this PR before final release of 2.0 👍

SteamKit2/SteamKit2/Util/Utils.cs
```diff
@@ -168,49 +150,13 @@ public static EOSType GetOSType()
...
-                if ( buffer != IntPtr.Zero )
-                    Marshal.FreeHGlobal( buffer );
-            }
-
-            return false;
         }
 
-        [DllImport ("libc")]
-        static extern int uname (IntPtr buf);
+        public static bool IsMacOS()
```
xPaw: This doesn't appear to be needed outside of this file, so maybe get rid of this function completely?

yaakov-h:
So now that SDK 2.0 and VS15.3 is out, what do we want to do with this?

Options:

1. Add .NET Standard 2.0 into the current lineup. Target Frameworks:
    - .NET Standard 1.3
    - .NET Standard 2.0
    - .NET Framework 4.6

2. Support .NET Standard only. Target Frameworks:
    - .NET Standard 1.3
    - .NET Standard 2.0

3. Support .NET Standard 2.0 only. Target Frameworks:
    - .NET Standard 2.0

4. Something else

I'm thinking we should do either (2) or (3) - we can get away with (3) since we're doing major breaking changes to SteamKit 2.0 anyway.

@VoiDeD @azuisleet @psychonic

xPaw:
If you're going to drop .NET framework support, there's little gain from not going straight to .net standard 2.0.

yaakov-h:
If we keep .NET Standard 1.3 then we still keep .NET Core 1.0/1.1, UWP pre-"Fall Creator's Update", and older versions of Mono and Xamarin.

Or we could just push everyone and everything to latest, since UWP vNext is slated for October... unless some poor sod is actually trying to use this with Windows 10 Mobile.

JustArchi:
.NET Standard is probably the most API-rich that is very close to original API of .NET Framework 4.6 that SK2 originally planned to target, so I say let's go for 2.0.

VoiDeD:
I'm on board with just targetting .NET Standard 2.0,

appveyor.yml
```diff
@@ -1,5 +1,5 @@
 version: 2.0.0.{build}
-os: Visual Studio 2017 Preview
+os: Visual Studio 2017
```
JustArchi: .NET Core SDK included in Visual Studio 2017 might not always be the one you want use. If instead you'd like to mimic travis behaviour and use some specific fixed version, feel free to take a look at **[my appveyor config](https://github.com/JustArchi/ArchiSteamFarm/blob/master/appveyor.yml#L19-L35)**, since it's very easy to install fixed SDK version and use it during build 👍.

yaakov-h: Typically an application should build with the latest SDK, but target for the lowest runtime it supports. AppVeyor currently ship SDK 2.0 on their VS2017 images, which suits me fine.

When there's a new SDK we can look at if we need to pin it - when 2.1 rolls around we can worry about it then. Hopefully they don't break compatibility 😄

yaakov-h:
I'll give it another day for further feedback, if there are no concerns then I'll merge.

appveyor.yml
JustArchi: I think it's different for .NET Standard library then, `netcoreapp2.0` is very picky as SDK typically includes corefx libraries only for given version, so if they decide to bump it, even `2.1` can be fatal for me. This is what happens when you try to build `netcoreapp1.1` with 2.0 SDK:

```
It was not possible to find any compatible framework version
The specified framework 'Microsoft.NETCore.App', version '1.1.2' was not found.
  - Check application dependencies and target a framework version installed at:
      /
  - Alternatively, install the framework version '1.1.2'.
```

This is why in my case I prefer to pin it to 2.0, so suddenly the above won't happen with 2.1 SDK trying to build `netcoreapp2.0` 🙂.

yaakov-h: That would be because you don't have the 1.x runtime installed... I'm still not clear on what is supposed to be bundled with the SDK and what should be shipped separately. Hopefully that becomes well-defined by 2.1.

JustArchi: > still not clear on what is supposed to be bundled with the SDK and what should be shipped separately.

Exactly, the only reason I decided to pin it, so CI won't break randomly at some point 😀. Still, probably unneeded for .NET standard library, but I decided to mention in case you'd be interested 👍.

yaakov-h: Our unit tests are netcoreapp.... 😈 

I'm OK in this instance with seeing how it breaks in CI first, rather than only finding out when someone upgrades VS/SDK.

PR #428, opened 2017-08-11
Completely remove UDP protocol support

xPaw:
YOLO. SteamKit has always defaulted to TCP, and I'm sure UDP doesn't seem much use in SK2.

Valve themselves are planning on phasing it out in favour of websocket protocol.

This PR just dumbly removes the UDP stuff, some of the connection stuff can probably be cleaned up further.


JustArchi:
Do we really need to remove this just now? I mean, it'd make total sense if Valve indeed removed UDP support from CMs, but removing completely valid and still working protocol to use based on removal promise sounds silly to me.

As for practical usage UDP can still be considered as a nice alternative for networks that are specifically blocking TCP connections to Steam CMs but not UDP ones. Completely irrelevant to consider this as an argument though.

yaakov-h:
I don't think we should remove UDP outright until well after Valve do.

xPaw:
I'll close this, feel free to open whenever the UDP removal point comes (if by that time this PR doesn't need a massive rebase).

Issue #438, opened 2017-08-31
Sometimes NetHook2 dumps no logs


ThorConzales:
I built a tool that relies on NetHook2 injection to provide data for it: https://github.com/ThorConzales/AoE2HDLobbyCompanion. It has been reported and I have witnessed it too that sometimes no logs are being intercepted and dumped once the injection is complete. It seems to happen randomly and mostly restarting steam client helps, sometimes multiple restarts are required. No unusual logs are being displayed in the console once it happens. Is this a known issue?

azuisleet:
This may be related to the use of WebSocket CM connections

ThorConzales:
Do you also know why and when Steam decides to use WS connection? Currently it seems to be pretty random. More importantly will you guy support WebSocket interception in the future?

azuisleet:
WebSocket is currently done at random when connecting. I believe there are some command line options like -udpforce or -tcp that you can use to steer it.

I don't see any reason we wouldn't also intercept the functions for WebSocket in NetHook, so it's likely going to be supported in the future.

ThorConzales:
Great, will eagerly wait for the WS interception. Based on the that documentation https://support.steampowered.com/kb_article.php?ref=5623-QOSV-5250 there is -tcp flag. However I'm concerned how much of a performance impact it will yield once it is downgraded from UDP in terms of latency. Have you guys played with this flag, are there any performance problems to be concerned about?

asherkin:
`-udpforce` is a recent addition and documented for game servers, so almost certainly also exists for the client.

JustArchi:
SK2 has been working in TCP mode by default since ages if not the first version. Valve's UDP mode is actually more like "tcp over udp" because it still has e.g. packets retransmission.

I doubt it'll result in any major difference.

DoctorMcKay:
There is zero performance difference. And voice chat is independent of the CM connection protocol.

VoiDeD:
#432 for some more discussion of how we might add websocket support for NetHook.

VoiDeD:
@ThorConzales please try the latest version of NetHook.

ThorConzales:
Tested 10 times and worked flawlessly. Did also release it now in my tool, will see if anyone complains. Thanks!

Issue #432, opened 2017-08-22
Create WebSocket version of NetHook


yaakov-h:
Should be doable at just a protocol level with something like FiddlerCore, though licensing looks sketchy.

VoiDeD:
Mirroring some notes from IRC, in case we want to expand NetHook to just hook more methods without fiddler/proxies/etc.

For sending: `CWebSocketConnection::BBuildAndAsyncSendFrame(EWebSocketOpCode,uchar const*,int)` might be a good starting point. This will be called after the `CWebSocketConnection` has fully assembled the data that will be in a frame, but before it's encoded into an actual web socket frame (and before it gets piped through the `COpenSSLConnection`). Data is passed right through the params, so we've got everything we need here.

Receiving is a bit more fun... `CWebSocketConnection` handles incoming data through `CWebSocketConnection::DecodePktAndDispatch(uchar *,uint)`. This method will handle decoding the frame and dumping it into a `m_sFrameCur` member, which is at the very least a struct with the first member being a `void *m_pubPayload`.

Once `CWebSocketConnection::DecodePktAndDispatch` is done decoding, it will call into `CWebSocketConnection::DispatchCurrentFrame(void)`, where it will create a `CNetPacket` from the data in `m_sFrameCur.m_pubPayload`.

Hooking `CWebSocketConnection::DispatchCurrentFrame` is probably the best option, but we'd need to maintain an offset to `m_sFrameCur`, which wouldn't be super fun going forward.

Another option might be hooking `CWebSocketConnection::RecvPkt(CNetPacket *)` , which will be called by `DispatchCurrentFrame`. Maintaining the structure of `CNetPacket` would likely be much easier.

VoiDeD:
Implemented via #439.

PR #439, opened 2017-09-02
Fix dumping of net messages sent over websocket connections.

VoiDeD:
Also dropping the hook on SymmetricDecrypt since the new network hooking for websocket hooks `CCMInterface::RecvPkt` that will also handle UDP and TCP connections.

In the future we can look into removing the `SymmetricEncrypt` hook and hooking a more relevant method on `CCMInterface` or `CCMConnection`, since the encryption method ends up getting called for more than just encrypting net messages.

Should solve #438 and #432 .

### DIFF
--- a/Resources/NetHook2/NetHook2/net.cpp
+++ b/Resources/NetHook2/NetHook2/net.cpp
@@ -97,7 +97,7 @@ bool CNet::BBuildAndAsyncSendFrame(void *webSocketConnection, void *unused, EWeb
 {
 	if (eWebSocketOpCode == k_eWebSocketOpCode_Binary)
 	{
-		g_pLogger->LogNetMessage(k_eNetOutgoing, const_cast<uint8 *>(pubData), cubData);
+		g_pLogger->LogNetMessage(k_eNetOutgoing, pubData, cubData);
 	}
 	else
 	{
--- a/Resources/NetHook2/NetHook2/net.cpp
+++ b/Resources/NetHook2/NetHook2/net.cpp
@@ -1,0 +1,120 @@
+
+#define WIN32_LEAN_AND_MEAN
+#include <windows.h>
+
+
+#include "net.h"
+
+#include "logger.h"
+#include "csimplescan.h"
+
+
+namespace NetHook
+{
+
+
+BBuildAndAsyncSendFrameFn BBuildAndAsyncSendFrame_Orig = NULL;
+RecvPktFn RecvPkt_Orig = NULL;
+
+CNet::CNet()
+	: m_RecvPktDetour(NULL),
+	  m_BuildDetour(NULL)
+{
+	CSimpleScan steamClientScan("steamclient.dll");
+
+	BBuildAndAsyncSendFrameFn pBuildFunc = NULL;
+	bool bFoundBuildFunc = steamClientScan.FindFunction(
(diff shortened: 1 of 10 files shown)


Resources/NetHook2/NetHook2/net.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+	if (m_BuildDetour)
+	{
+		m_BuildDetour->Detach();
+		delete m_BuildDetour;
+	}
+}
+
+
+bool CNet::BBuildAndAsyncSendFrame(void *webSocketConnection, void *unused, EWebSocketOpCode eWebSocketOpCode, const uint8 *pubData, uint32 cubData)
```
yaakov-h: Is `unused` actually unused, or just unknown?

Resources/NetHook2/NetHook2/net.cpp
```diff
@@ -0,0 +1,120 @@
...
+{
+	if (eWebSocketOpCode == k_eWebSocketOpCode_Binary)
+	{
+		g_pLogger->LogNetMessage(k_eNetOutgoing, const_cast<uint8 *>(pubData), cubData);
```
yaakov-h: I'm not so comfortable with removing the `const` modifier. Can we make `LogNetMessage` take a const buffer instead?

Resources/NetHook2/NetHook2/net.cpp
```diff
@@ -0,0 +1,120 @@
...
+	{
+		g_pLogger->LogConsole("Sending websocket frame with opcode %d (%s), ignoring\n",
+			eWebSocketOpCode, EWebSocketOpCodeToName(eWebSocketOpCode)
+		);
+	}
+
+	return (*BBuildAndAsyncSendFrame_Orig)(webSocketConnection, unused, eWebSocketOpCode, pubData, cubData);
+}
+
+void CNet::RecvPkt(void *cmConnection, void *unused, CNetPacket *pPacket)
```
yaakov-h: Same here regarding `unused`

Resources/NetHook2/NetHook2/steam/net.h
```diff
@@ -0,0 +1,71 @@
...
+
+#include "steamtypes.h"
+
+
+typedef uint32 HCONNECTION;
+
+
+enum EWebSocketOpCode
+{
+	k_eWebSocketOpCode_Continuation	= 0x00,
```
yaakov-h: Nit: 0x00 isn't vertically aligned with the others here.

yaakov-h:
Works beautifully. 👍

Resources/NetHook2/NetHook2/net.cpp
VoiDeD: `fastcall` is just a quick and dirty way to hook `thiscall` methods on windows since it passes the first 2 parameters through `ecx` and `edx`, and we really only care about the `this` pointer being passed through `ecx`. `unused` is just the value in `edx`.

yaakov-h: Nasty, but clever.

Resources/NetHook2/NetHook2/steam/net.h
VoiDeD: It's aligned with a tab character in VS, I blame github.

JustArchi:
I love how hackish and clever this is at the same time 👍

PR #440, opened 2017-09-03
Use ServerRecord in CMListCallback

### LINKED ISSUE #434: CMListCallback should use ServerRecord
As half of the code has been changed to accept `ServerRecord`, this callback should use them too.

xPaw:
Fixes #434


yaakov-h:
Should also include `CMsgClientCMList.cm_websocket_addresses`.

xPaw:
@yaakov-h Like that?

yaakov-h:
I wasn't sure whether we should do that, or mirror the proto and have two properties.

No big deal, though, this approach should work just fine.

PR #441, opened 2017-09-04
Check arguments before attempting to use them

### LINKED ISSUE #412: Check arguments before doing stuff
While looking at #268 it seems that we almost never throw `Argument[Null|OutOfRange]Exception` if a consumer passes invalid inputs (e.g. null SteamID).

We should add null checks and range checks to our public API surface.

yaakov-h:
There are a lot of places in SK2 where we can be passed null values and either attempt to operate on them immediately, or store them and fail at a later time.

This PR adds null argument checking to the API surface, so that consumer misuse is exposed immediately at the original call-site, with an explanation of what is wrong.

PR for tracking and discussion at this point in time.

Fixes #412.

Todo:
- [x] Base
- [X] Networking
- [X] Steam
- [x] Types
- [x] Util

### DIFF
--- a/SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs
+++ b/SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs
@@ -232,13 +232,19 @@ namespace SteamKit2
             async Task<KeyValue> CallAsyncCore( HttpMethod method, string func, int version = 1, Dictionary<string, string> args = null )
             {
                 if ( method == null )
+                {
                     throw new ArgumentNullException( nameof(method) );
+                }
 
                 if ( func == null )
+                {
                     throw new ArgumentNullException( nameof(func) );
+                }
 
                 if ( args == null )
+                {
                     args = new Dictionary<string, string>();
+                }
 
 
                 var urlBuilder = new StringBuilder();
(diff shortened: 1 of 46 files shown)


SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs
```diff
@@ -443,6 +449,16 @@ public override bool TryInvokeMember( InvokeMemberBinder binder, object[] args,
         /// <returns>A dynamic <see cref="Interface"/> object to interact with the Web API.</returns>
         public static Interface GetInterface( Uri baseAddress, string iface, string apiKey = "" )
         {
+            if ( baseAddress == null )
```
xPaw: Shouldn't these be done in `AsyncInterface` constructor instead?

SteamKit2/SteamKit2/Types/Manifest.cs
```diff
@@ -115,6 +115,11 @@ private Steam3Manifest()
 
         public Steam3Manifest(byte[] data)
         {
+            if (data ==  null)
```
xPaw: spacing

SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs
yaakov-h: Could be, but then param names could get out of sync.

Issue #442, opened 2017-09-05
Player name is changed to [unassigned] on successful login


GabeHirakawa:
I've ran into an issue where whenever I successfully log into steam the user account that I logged into's name is changed to: [unassigned]. I've tried looking around to see what could possibly find the issue and the only thing I can find is referenced here: https://github.com/Jessecar96/SteamKit2/blob/master/SteamKit2/changes.txt

> Local user is now given the persona name [unassigned] before SteamUser.AccountInfoCallback comes in.

However, this is from 2012 and is also not the official repo so I'm not sure how relevant it is. I'm kind of stuck and feel like I'm missing something pretty obvious. Any help with this would be appreciated.

JustArchi:
https://github.com/JustArchi/ArchiSteamFarm/blob/master/ArchiSteamFarm/Bot.cs#L1429-L1431

You're welcome 🙂.

GabeHirakawa:
@JustArchi So that did seem to the nickname issue, but now it doesn't seem to be putting me online. Any idea?

JustArchi:
If you put an example code that'd be much better than asking my crystal ball, I'm trying to keep my mana for later.

GabeHirakawa:
@JustArchi My bad, I forgot to copy and paste /facepalm
```
        private static async void OnAccountInfo(SteamUser.AccountInfoCallback callback)
        {
            string nickname = _steamFriends.GetPersonaName();
            for (byte i = 0; (string.IsNullOrEmpty(nickname) || nickname.Equals("[unassigned]")); i++) {
                await Task.Delay(1000).ConfigureAwait(false);
                nickname = _steamFriends.GetPersonaName();
            }

            if (string.IsNullOrEmpty(nickname) || nickname.Equals("[unassigned]")) {
                return;
            }

            try {
                await _steamFriends.SetPersonaState(EPersonaState.Online);
            } catch (Exception e) {
                Console.WriteLine(e);
            }
        }
```

yaakov-h:
Can you post a reproducible example?

I have a suspicion you're trying to set your persona state too early.

GabeHirakawa:
Here we go, this should work better for all of us. https://github.com/GabeHirakawa/SteamFriendUtility/blob/master/SteamFriendUtility/Program.cs

JustArchi:
[...]

PR #443, opened 2017-09-05
Never permit of using [unassigned] value in SetPersonaState() + deterministic name

JustArchi:
Please review the changes carefully, this can cause serious regressions if I didn't catch some common-use scenario.

Rationale: #442

### DIFF
--- a/SteamKit2/SteamKit2/Steam/Handlers/SteamFriends/SteamFriends.cs
+++ b/SteamKit2/SteamKit2/Steam/Handlers/SteamFriends/SteamFriends.cs
@@ -57,7 +57,8 @@ namespace SteamKit2
 
 
         /// <summary>
-        /// Gets the local user's persona name. Can be null before user's initialization.
+        /// Gets the local user's persona name. Will be null before user's initialization.
+        /// User initialization is typically happening during <see cref="AccountInfoCallback"/> callback.
         /// </summary>
         /// <returns>The name.</returns>
         public string GetPersonaName()
@@ -144,7 +145,8 @@ namespace SteamKit2
         }
 
         /// <summary>
-        /// Gets the persona name of a friend. Can be null before friend's initialization.
+        /// Gets the persona name of a friend. Will be null before friend's initialization.
+        /// Friend initialization is typically happening during <see cref="PersonaStateCallback"/> callback.
         /// </summary>
         /// <param name="steamId">The steam id.</param>
         /// <returns>The name.</returns>
@@ -226,7 +228,8 @@ namespace SteamKit2
         }
 
         /// <summary>
-        /// Gets the name of a clan. Can be null before clan's initialization.
+        /// Gets the name of a clan. Will be null before clan's initialization.
+        /// Clan initialization is typically happening during <see cref="PersonaStateCallback"/> callback.
         /// </summary>
         /// <param name="steamId">The clan SteamID.</param>
         /// <returns>The name.</returns>
--- a/SteamKit2/SteamKit2/Steam/Handlers/SteamFriends/SteamFriends.cs
+++ b/SteamKit2/SteamKit2/Steam/Handlers/SteamFriends/SteamFriends.cs
@@ -104,18 +104,12 @@ namespace SteamKit2
         /// <returns>The Job ID of the request. This can be used to find the appropriate <see cref="PersonaChangeCallback"/>.</returns>
         public AsyncJob<PersonaChangeCallback> SetPersonaState( EPersonaState state )
         {
-            if (cache.LocalUser.Name == null) {
-                throw new ArgumentNullException(nameof(cache.LocalUser.Name),
-                    "The function was called before name of the local user could be registered.");
-            }
-
             cache.LocalUser.PersonaState = state;
 
             var stateMsg = new ClientMsgProtobuf<CMsgClientChangeStatus>( EMsg.ClientChangeStatus );
             stateMsg.SourceJobID = Client.GetNextJobID();
 
             stateMsg.Body.persona_state = ( uint )state;
-            stateMsg.Body.player_name = cache.LocalUser.Name;
 
             this.Client.Send( stateMsg );
 
--- a/SteamKit2/SteamKit2/Steam/Handlers/SteamFriends/SteamFriends.cs
+++ b/SteamKit2/SteamKit2/Steam/Handlers/SteamFriends/SteamFriends.cs
@@ -57,8 +57,8 @@ namespace SteamKit2
 
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


JustArchi:
The biggest advantage of this is fixing two outstanding issues I have with current implementation:
1. It's not possible to tell a difference between pre-initialized name field and user who put `[unassigned]` as his name in particular.
2. If `SetPersonaState()` will be called before name initialization, as seen in #442 and in my own projects that I've patched, name of the user is changed to `[unassigned]`, which is unwanted. I think exception in this case is justified - user should ensure that nickname is initialized. If needed we could improve it by adding extra documentation and/or requiring from user to put valid `name` argument instead of using our cached data.

If I didn't miss any usec ase of those two fields, this is a breaking change as default value of uninitialized user changed, but I think it's justified if we're able to fix both points above while not generating any major regressions in internal SK2 code. While I can verify fixes, I hope you can help me with SK2 internals knowledge. It looks good to me, but I don't know private SK2 code that good.

yaakov-h:
I was thinking more along the lines of omitting `stateMsg.Body.player_name` and only sending `stateMsg.Body.persona_state`.

We have `SetPersonaName` and `SetPersonaState` as two separate functions, but each implicitly does the same role as the other.

I'm assuming though that Steam can handle only one of the two fields being set... I'll have to have a look at how Steam handles it and at what point in the application lifecycle it calls `SetPersonaState( k_EPersonaState_Online )` or similar.

JustArchi:
If you confirm that we can omit it then we can get rid of that ugly exception and setting that name variable in the request itself. This is even more satisfying than current solution, let me know about your results.

yaakov-h:
Steam's behaviour is as I remember it:

1. `SetPersonaState` and `SetPersonaName` both update client-side records, similar to SteamKit. They then call `CCMInterface::SendFriendsStatusToServer(bool, unsigned long long)`
1. `CCMInterface::SendFriendsStatusToServer` appears to have 3 different codepaths that send `CMsgClientChangeStatus`. I think this is new? I can't quite tell what the difference is between each of them.

For a more practical approach, rather than an analytic one, I just tested with SteamKit and if you send `CMsgClientChangeStatus` without `player_name`, the CMs preserve the old name.

On the flip side, if you omit `persona_state`, Steam interprets it as the default value, i.e. 0, i.e. `k_EPersonaStateOffline`.

JustArchi:
Now the only question is if something explicitly depends on default value of account's name. I bet not in SK2 internals, since otherwise it should be marked with `const` and properly referenced in some other place.

I love how we could get rid of that exception and just avoid setting name entirely 👍.

yaakov-h:
This means now that `GetPersonaName`, `GetFriendPersonaName` and `GetClanName` can all now return null.

SteamKit2/SteamKit2/Steam/Handlers/SteamFriends/SteamFriends.cs
```diff
@@ -57,7 +57,7 @@ internal SteamFriends()
 
 
         /// <summary>
-        /// Gets the local user's persona name.
+        /// Gets the local user's persona name. Can be null before user's initialization.
```
VoiDeD: I think we need to do a better job of defining exactly when the local user's persona name will be available.

VoiDeD:
I like the idea of excluding the local name in `SetPersonaState`, that looks fine. But I'd really like to get a handle on when the local user's name is available and document that a little better before we merge this.

SteamKit2/SteamKit2/Steam/Handlers/SteamFriends/SteamFriends.cs
JustArchi: Feel free to update PR with a better explanation if appropriate - GitHub recently added a way for repo maintainers to update PRs if contributor enables such option 🙂.

yaakov-h:
The local user's name is available after `AccountInfoCallback`, but I think part of the problem is that:

1. The local user's name is only available after `SteamFriends` handles the account info msg
1. Custom `ClientMsgHandler`s may also want to handle the account info msg and call code which depends on `SteamFriends`
1. The `AccountInfoCallback` is posted by `SteamUser`, not `SteamFriends.
1. The order in which `ClientMsgHandler`s are run is undefined because we are enumerating through a dictionary's values, not an ordered collection.

Therefore, code which handles the account info msg or code which handles the account info callback could have the local user name set, or could not. It's one big fat race condition.

For code which handles subsequent messages/callbacks, the local user name should always be available.

JustArchi:
It's also important to remember that Steam might not always send callbacks in appropriate order, so we might receive `AccountInfoCallback` at later stage, not making it possible to say "`AccountInfoCallback` will always arrive before XXX callback".

It's fine to say that local user's name will be available after receiving `AccountInfoCallback`, but we can't make any more assumptions in terms of callback order. I think it could be possible to iterate over dictionary in a way that SK2 handlers could be evaluated firstly, we could even add a `bool IsPriority` and do `.OrderByDescending(handler => handler.IsPriority ? 1 : 0)`.

yaakov-h:
Or just use a `System.Collections.Specialized.OrderedDictionary`.

JustArchi:
How about now? It looks enough documented to me.

yaakov-h:
Could do with some grammatical tweaks,:

* "user initialisation", not "user's initialisation".
* Should not be left ambiguous, e.g. "is performed prior to `AccountInfoCallback` rather than "typically happens during `AccountInfoCallback`". Although in this case we can't be honest as per above, since `AccountInfoCallback` could be triggered before `SteamFriends` finishes... perhaps document how it *should* work, then fix the behaviour.

I'd also remove the initialisation descriptions from Friend and Clan. We go from knowing nothing about them to knowing the exist, and Steam provides us with the name at that point in time (though it's not guaranteed to do so - protocol could change). The old xmldoc is probably sufficient for these two, and we can use the release notes to announce that they could potentially be null in a severe edge case.

JustArchi:
I think this should work now.

Apart from docs suggestions I changed internal implementation of handlers to ordered dictionary, so SK2 handlers are always first (in our fixed order), and then user handlers are executed last.

I also changed order of SteamUser -> SteamFriends to SteamFriends -> SteamUser during initialization, I stopped for a second wondering if old order was supposed to mean anything, but then, previous dictionary had unpredictable order anyway, so it should not. I checked if all tests were passing after doing this change, looks good to me.

JustArchi:
As of now I think this PR is in good state to be merged, unless somebody has something else to add.

We could in the future improve handlers and their order, but even current solution is good enough and doesn't result in worse overall health condition than previous one, in fact, it fixes things the way they should be fixed to begin with.

yaakov-h:
I'm happy with this. @VoiDeD?

PR #445, opened 2017-09-06
Allow setting webapi key in configuration

### LINKED ISSUE #431: Add webapi key configuration option
Changing webapi base url to `https://partner.steam-api.com` always requires a key, but the current situation is that GetCMList will be called without a key internally and the request will always fail.

xPaw:
Fixes #431

### DIFF
--- a/SteamKit2/SteamKit2/Steam/WebAPI/SteamConfigurationWebAPIExtensions.cs
+++ b/SteamKit2/SteamKit2/Steam/WebAPI/SteamConfigurationWebAPIExtensions.cs
@@ -1,0 +1,42 @@
+﻿using System;
+
+namespace SteamKit2
+{
+    /// <summary>
+    /// Provides helper extensions to make WebAPI interfaces from existing SteamConfiguration.
+    /// </summary>
+    public static class SteamConfigurationWebAPIExtensions
+    {
+        /// <summary>
+        /// Retrieves a dynamic handler capable of interacting with the specified interface on the Web API.
+        /// </summary>
+        /// <param name="config">The configuration to use for this Web API interface.</param>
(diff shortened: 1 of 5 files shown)


JustArchi:
I'm wondering if it wouldn't make sense to also use it in all API requests made from the `WebApi.GetInterface()` when user didn't specify API key. This way we could define one key in configuration and omit it when getting interface for api requests.

xPaw:
@JustArchi Yes i'm making an extension to SteamConfiguration, you'll see it in next commit.

SteamKit2/SteamKit2/Steam/WebAPI/SteamConfigurationWebAPIExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+    {
+        /// <summary>
+        /// Retrieves a dynamic handler capable of interacting with the specified interface on the Web API.
+        /// </summary>
+        /// <param name="config">The configuration to use for this Web API interface.</param>
+        /// <param name="iface">The interface to retrieve a handler for.</param>
+        /// <returns>A dynamic <see cref="WebAPI.Interface"/> object to interact with the Web API.</returns>
+        public static WebAPI.Interface GetWebAPIInterface(this SteamConfiguration config, string iface)
+        {
+            if (config == null)
```
JustArchi: Is it even possible for this to be `null` if it's extension method?

yaakov-h: Yes. You can call an extension method on `null`, or manually call this function since it's a public static function.

PR #447, opened 2017-09-06
Expose HTTP timeout on WebAPI

### LINKED ISSUE #381: WebAPI.GetAsyncInterface() is missing Timeout property
As initially stated in #379:

Async interface returned by `WebAPI.GetAsyncInterface()` doesn't have `Timeout` property like its non-async version. Since it uses `HttpClient` under the hood, and that one can have custom `Timeout` property, I think it should be added.

In my code instead of:

```
HttpClient = new HttpClient();
```

I'm using:

```
HttpClient = new HttpClient {
	Timeout = TimeSpan.FromMilliseconds(60 * 1000) // Or anything user provides as int, I used FromMilliseconds to keep compatibility with non-async interface
};
```

Is there anything problematic regarding this in .NET core? I'm looking for a possible reason why @Netshroud decided to omit it as stated **[here](https://github.com/SteamRE/SteamKit/issues/379#issuecomment-290259479)**.

https://github.com/SteamRE/SteamKit/blob/f41261be563d6b999be3e46102c77e4768074e76/SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs#L220

Thanks!

yaakov-h:
- Use `HttpClient.Timeout` as the implementation of our timeout mechanism.
- This will throw a `TaskCanceledException` (subclass of `OperationCanceledException`) on timeout now, instead of `TimeoutException`.
- Add `Timeout` property to `AsyncInterface`
- Change `Timeout` property from `int` to `TimeSpan`

Fixes #381.

@JustArchi: mind reviewing?

### DIFF
--- a/SteamKit2/Tests/WebAPIFacts.cs
+++ b/SteamKit2/Tests/WebAPIFacts.cs
@@ -11,7 +11,15 @@ namespace Tests
         {
             var iface = WebAPI.GetInterface( new Uri("https://whatever/"), "ISteamWhatever" );
 
-            Assert.Equal( iface.Timeout, 1000 * 100 );
+            Assert.Equal( iface.Timeout, TimeSpan.FromSeconds( 100 ) );
+        }
+
+        [Fact]
+        public void WebAPIAsyncHasDefaultTimeout()
+        {
+            var iface = WebAPI.GetAsyncInterface( new Uri("https://whatever/"), "ISteamWhatever" );
+
+            Assert.Equal( iface.Timeout, TimeSpan.FromSeconds( 100 ) );
         }
 
         [Fact]
--- a/Samples/6.WebAPI/Program.cs
+++ b/Samples/6.WebAPI/Program.cs
@@ -64,9 +64,9 @@ namespace Sample6_WebAPI
                 {
                     kvNews = steamNews.GetNewsForApp002( appid: 730, maxlength: 100, count: 5 );
                 }
-                catch ( WebException ex )
+                catch ( Exception ex )
                 {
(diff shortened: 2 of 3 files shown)


SteamKit2/Tests/WebAPIFacts.cs
```diff
@@ -11,7 +11,15 @@ public void WebAPIHasDefaultTimeout()
...
-            Assert.Equal( iface.Timeout, 1000 * 100 );
+            Assert.Equal( iface.Timeout, TimeSpan.FromSeconds( 100 ) );
+        }
+
+        [Fact]
+        public void WebAPIAsyncHasDefaultTimeout()
+        {
+            var iface = WebAPI.GetAsyncInterface( new Uri("https://whatever/"), "ISteamWhatever" );
+
+            Assert.Equal( iface.Timeout, TimeSpan.FromSeconds( 100 ) );
```
JustArchi: Maybe we should make it less hardcoded, such as `internal const int defaultTimeout = 100` in `WebAPI`? Just an idea though. If we ever want to change timeout it'd be better for tests to match it too, unless we intentionally don't want to change default timeout.

yaakov-h: Personal opinion only, unit testing should not depend upon the same constants when that's the thing it's testing. If you want to change tested functionality, you should be required to be make at minimum two changes - one to functional code and one to test code.

JustArchi: This is fine, it's a matter of whether we want to test that our default timeout is set, or whether we want to test that and in addition to that test it also for specific hardcoded value. It's a matter of what really test should be doing, in your case I assume second, which is totally OK.

PR #461, opened 2017-10-01
Handle exception when creating connection.

### LINKED ISSUE #455: Does websocket fail properly when not being able to connect?
One of my users can't establish a connection when using websocket protocol, while totally being able to do so through TCP protocol.

By taking a look at ASF debug log, I noticed the following:

```
2017-09-26 01:53:15|ArchiSteamFarm-5028|INFO|bot|Start() Starting...
2017-09-26 01:53:15|ArchiSteamFarm-5028|INFO|bot|Connect() Connecting...
2017-09-26 01:53:15|ArchiSteamFarm-5028|DEBUG|ASF|WriteLine() ServerList | Resolving server list
2017-09-26 01:53:15|ArchiSteamFarm-5028|DEBUG|ASF|WriteLine() ServerList | Resolved 160 servers
2017-09-26 01:53:15|ArchiSteamFarm-5028|DEBUG|ASF|WriteLine() ServerList | Next server candidiate: Unspecified/CM01-LUX.cm.steampowered.com:27021 (WebSocket)
2017-09-26 01:55:15|ArchiSteamFarm-5028|WARN|bot|InitPermanentConnectionFailure() Failed to disconnect the client. Abandoning this bot instance!
2017-09-26 01:55:15|ArchiSteamFarm-5028|INFO|bot|Stop() Stopping...
2017-09-26 01:55:15|ArchiSteamFarm-5028|INFO|bot|Start() Starting...
2017-09-26 01:55:15|ArchiSteamFarm-5028|INFO|bot|Connect() Connecting...
2017-09-26 01:55:15|ArchiSteamFarm-5028|DEBUG|ASF|WriteLine() ServerList | Next server candidiate: Unspecified/CM01-LUX.cm.steampowered.com:27021 (WebSocket)
```

It seems that websocket does not fail in at least 2 minutes of timeout. You can notice my workaround code that tries to detect fuckups like these (stalled connections) and destroys/recreates the bot from the ground up, but then it tries to connect to exactly the same server and loop goes forever.

Is this normal? Is websocket timeout configurable? Maybe you'd want to dig deeper into this? I don't know if this should be classified as a bug, it looks like a bug to me since websocket doesn't react in any way for 2 minutes since attempting to connect. This does not happen when SK2 fails to connect via TCP or UDP - `OnDisconnected()` always arrives.

yaakov-h:
Fixes #455.

@JustArchi Want to do a quick review, considering you raised the issue?

This doesn't prevent users from getting themselves into a toxic situation (e.g. setting `ProtocolTypes` to WebSocket-only on Windows 7) but does trigger a disconnected callback, same as when the connection fails for any other reason.

### DIFF
--- a/SteamKit2/SteamKit2/Steam/CMClient.cs
+++ b/SteamKit2/SteamKit2/Steam/CMClient.cs
@@ -138,8 +138,8 @@ namespace SteamKit2.Internal
 
             var cancellation = new CancellationTokenSource();
             var token = cancellation.Token;
-            var oldCancellation = Interlocked.Exchange(ref connectionCancellation, cancellation);
-            Debug.Assert(oldCancellation == null);
+            var oldCancellation = Interlocked.Exchange( ref connectionCancellation, cancellation );
+            Debug.Assert( oldCancellation == null );
 
             ExpectDisconnection = false;
 
@@ -154,7 +154,7 @@ namespace SteamKit2.Internal
                 recordTask = Task.FromResult( cmServer );
             }
 
-            connectionSetupTask = recordTask.ContinueWith(t =>
+            connectionSetupTask = recordTask.ContinueWith( t =>
             {
                 if ( token.IsCancellationRequested )
                 {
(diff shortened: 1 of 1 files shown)


JustArchi:
What I like the most in this is the fact that eventually bot wil succeed connecting using TCP-based server when he's using not-only-websocket protocols, as CM list should be smart enough to push faulted server on the bottom. This works much better than current situation of stalled connection where you don't know what happened without taking a closer look - I think it's a good solution.

Issue #450, opened 2017-09-07
Making it possible for using custom HttpClient in WebAPI?


JustArchi:
This is a random idea I thought about for a second. Since new async WebAPI interface uses HttpClient under the hood, how about making it possible to re-use already initialized and ready HttpClient that is being used by the consumer?

Microsoft recommends to re-use HttpClient as much as possible, going as far as using a single `static readonly` instance across entire program's lifetime. I don't think this is appropriate for our case, as it still takes resources and there is no need to have SK2-dedicated HttpClient if there are no requests happening, but being able to pass `HttpClient` to `GetAsyncInterface()` would be a really nice thing to do in order to re-use existing sockets and entire client initialization. Of course, in such case the client can't be disposed as consumer probably expects it to be used further.

I'm mainly looking at my own use case here, where I already have `HttpClient` being initialized and used for sending requests, so I can re-use it in its current state in order to save memory and overhead of creating another one. Creating a new HttpClient for a single request most of the time is wasted effort, and it can even become **[a serious problem](https://aspnetmonsters.com/2016/08/2016-08-27-httpclientwrong/)** in some cases.

Thoughts? I even wanted to send a PR with this right away, but noticed that it'd require more changes (such as base URL rewrite), so I decided to create issue first.

yaakov-h:
This is definitely something to consider.

I saw this same behaviour at work - each `HttpClient` meant a new set of SSL/TLS connections. We moved to a singleton instance and performance drastically improved.

I'm quite tempted to (post-2.0) revisit the entire WebAPI... well, API.

I'm not sure I like the idea of allowing custom `HttpClient`s, since a custom `HttpMessageHandler` could screw with things in subtle ways, then we get blamed for it - but you should definitely be permitted to somehow share an `HttpClient` between multiple `Interface` or `AsyncInterface` objects. We could probably hack that together today by adding it to the `SteamConfiguration` object...

Alternatively, we could make WebAPI a set of extensions to `HttpClient`/`HttpRequestMessage`, and drop the custom dynamic object invocations.

VoiDeD:
`SteamConfiguration` sounds like it may grow to be unwieldy in the future. Wonder if we'll want to have sub-configuration objects or something in the future.

yaakov-h:
That sounds like a good idea.

With `SteamConfiguration` I tried to extract various common settings and static state. I didn't realise there was quite so much of it 😅 

We could possibly subdivide the config with #415.

[...]

PR #468, opened 2017-10-08
Make SteamConfiguration immutable

### LINKED ISSUE #415: Make SteamConfiguration immutable, somehow
Left-over from #405.

A few options:

1. The Builder Pattern - i.e. `new SteamConfigurationBuilder().WithBlah(...).Build()`
2. The `System.Windows.Freezable` pattern, i.e. `configuration.Freeze()`
3. Truly immutable, every set method returns a copy of the object i.e. `SteamConfiguration SetBlah(Blah value)`

Anything else?

yaakov-h:
Fixes #415.

This is my first stab at it.

I figured we should make the way to construct a `SteamConfiguration` instance easily discoverable on the type itself, so `SteamConfiguration.Create` is the only way for external code and the canonical way for internal code to create a configuration object.

Internal code can also use `SteamConfiguration.CreateDefault` as a reasonable default where a consumer has not passed in a configuration.

The factory method approach here is somewhat similar to ASP.NET WebAPI's `GlobalConfiguration.Configure` method.

Splitting HTTP/WebAPI has been left for later as per #450.

Thoughts, feedback and [constructive] criticism welcome.

### DIFF
--- a/Samples/7.ServerList/Program.cs
+++ b/Samples/7.ServerList/Program.cs
@@ -35,8 +35,6 @@ namespace Sample7_ServerList
             user = args[ 0 ];
             pass = args[ 1 ];
 
-            var configuration = new SteamConfiguration();
-
             var cellid = 0u;
 
             // if we've previously connected and saved our cellid, load it.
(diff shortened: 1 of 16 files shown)


Samples/7.ServerList/Program.cs
```diff
@@ -45,15 +43,17 @@ static void Main( string[] args )
                 if ( !uint.TryParse( File.ReadAllText( "cellid.txt"), out cellid ) )
                 {
                     Console.WriteLine( "Error parsing cellid from cellid.txt. Continuing with cellid 0." );
+                    cellid = 0;
```
JustArchi: `cellid` will be `default(uint)` in this case which will be `0` anyway, but it might be good to explicitly state this for code readability.

yaakov-h: Actually it’ll be whatever `uint.TryParse` sets it to, which is documented as “zero if the conversion failed”, but either way I just added it in for readability.

PR #470, opened 2017-10-14
Prefer stable release of Microsoft.Win32.Registry

JustArchi:
Side note: should we keep bumping `protobuf-net` to latest released version after confirming that it works fine, or we're keeping it at `2.1.0` as a minimum requirement?


yaakov-h:
Pity it's not in .NET Standard. 😞

https://apisof.net/catalog/Microsoft.Win32.RegistryKey.OpenBaseKey(RegistryHive,RegistryView)

I don't think there's anything new in protobuf-net that we need, and the ProtoGen story is very very odd at the moment... at least last time I checked (2.3.0).

PR #498, opened 2017-12-12
fix(samples): unable to build samples solution

leonard-thieu:
Building the Samples solution in Visual Studio 2017 results in errors like

`Unable to find project information for 'S:\Projects\_toofz\SteamKit\SteamKit2\SteamKit2\SteamKit2.csproj'. The project file may be invalid or missing targets required for restore.`

The path is correct but the build still fails. Adding the `SteamKit2` project to the solution fixes it.


yaakov-h:
Thanks!

PR #500, opened 2017-12-18
De-async-ify SetPersonaName and SetPersonaState

### LINKED ISSUE #491: SteamFriends.SetPersonaState() task never returns?
Consider following code in `async OnLoggedOn()`:

```
try {
	await SteamFriends.SetPersonaState(EPersonaState.Online);
} catch (Exception e) {
	Console.WriteLine(e);
}
```

`SetPersonaState()` returns awaitable `AsyncJob`, but compared to other async jobs I'm awaiting, this one seems to never return. I suspect this is because in fact there is no Steam Network response in `SetPersonaState()`, so despite of request succeeding, `TaskCanceledException` is always raised in 10 seconds since request being sent.

I reproduced this more than dozen of times already, even adding `await Task.Delay(10000).ConfigureAwait(false)` before call just to ensure it's not some packet disappearing when callback manager is busy handling other things immediately after logging in, no luck. Perhaps it was already fixed in #481 and I just coincidentally reproduced it just now?

Thing is, I reproduced it in console app, and my understanding is that there should be no issue like #481 in my case, so maybe it's just false logic of `SetPersonaState()` returning `AsyncJob` when it should just return `void` instead.

Thank you in advance.

yaakov-h:
`CMsgClientChangeStatus` is no longer a job message.

Fixes #491.

### DIFF
--- a/SteamKit2/SteamKit2/Steam/Handlers/SteamFriends/SteamFriends.cs
+++ b/SteamKit2/SteamKit2/Steam/Handlers/SteamFriends/SteamFriends.cs
@@ -67,25 +67,19 @@ namespace SteamKit2
         }
         /// <summary>
         /// Sets the local user's persona name and broadcasts it over the network.
-        /// Results are returned in a <see cref="PersonaChangeCallback"/> callback.
-        /// The returned <see cref="AsyncJob{T}"/> can also be awaited to retrieve the callback result.
         /// </summary>
         /// <param name="name">The name.</param>
-        /// <returns>The Job ID of the request. This can be used to find the appropriate <see cref="PersonaChangeCallback"/>.</returns>
-        public AsyncJob<PersonaChangeCallback> SetPersonaName( string name )
+        public void SetPersonaName( string name )
         {
             // cache the local name right away, so that early calls to SetPersonaState don't reset the set name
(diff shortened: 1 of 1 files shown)


VoiDeD:
Do you think we're missing a client protocol version bump somewhere? Maybe
the job functionality moved behind a newer protocol version on the server
(because Valve).

yaakov-h:
NetHook on the client (as of [22 days ago](https://github.com/SteamRE/SteamKit/issues/491#issuecomment-347088756)) shows it without a job id.

JustArchi:
The issue itself is much older, at least several good months passed before users reported issue with this for the first time, let me grab some initial date...

```
2017-10-02 17:23:30|dotnet-6090|WARN|HucKer|OnAccountInfo() System.Threading.Tasks.TaskCanceledException: A task was canceled.
at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task) at ArchiSteamFarm.Bot.<OnAccountInfo>d__133.MoveNext() in C:\projects\archisteamfarm\ArchiSteamFarm\Bot.cs:line 1460
```

At least October this year.

Issue #260, opened 2016-03-29
[Bug] SteamClient in UDP mode


JustArchi:
Noticed below crash from one of my ASF users:

```
29/03/2016 23:36:04 [!] EXCEPTION: UnhandledExceptionHandler() <Main> Unable to decrypt incoming packet
29/03/2016 23:36:06 [!] StackTrace:
   em SteamKit2.NetFilterEncryptionWithHMAC.ProcessIncoming(Byte[] data)
   em SteamKit2.UdpConnection.DispatchMessage()
   em SteamKit2.UdpConnection.ReceiveData(UdpPacket packet)
   em SteamKit2.UdpConnection.ReceivePacket(UdpPacket packet)
   em SteamKit2.UdpConnection.NetLoop()
   em System.Threading.ThreadHelper.ThreadStart_Context(Object state)
   em System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   em System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   em System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
   em System.Threading.ThreadHelper.ThreadStart()
29/03/2016 23:36:06 [!] EXCEPTION: UnhandledExceptionHandler() <Main> O comprimento dos dados a desencriptar é inválido (Translation: The length of the data to decrypt is invalid).
29/03/2016 23:36:06 [!] StackTrace:
   em System.Security.Cryptography.RijndaelManagedTransform.TransformFinalBlock(Byte[] inputBuffer, Int32 inputOffset, Int32 inputCount)
   em System.Security.Cryptography.CryptoStream.FlushFinalBlock()
   em System.Security.Cryptography.CryptoStream.Dispose(Boolean disposing)
   em System.IO.Stream.Close()
   em System.IO.Stream.Dispose()
   em SteamKit2.CryptoHelper.SymmetricDecrypt(Byte[] input, Byte[] key, Byte[]& iv)
   em SteamKit2.CryptoHelper.SymmetricDecryptHMACIV(Byte[] input, Byte[] key, Byte[] hmacSecret)
   em SteamKit2.NetFilterEncryptionWithHMAC.ProcessIncoming(Byte[] data)
```

(This is complete stacktrace, my logger logs exception and every inner exception recursively)

[...]

PR #501, opened 2017-12-18
Handle CMClient.GetPacketMsg returning null.

yaakov-h:
Possible fix for #260.

### DIFF
--- a/SteamKit2/SteamKit2/Networking/Steam3/EnvelopeEncryptedConnection.cs
+++ b/SteamKit2/SteamKit2/Networking/Steam3/EnvelopeEncryptedConnection.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Diagnostics;
 using System.Net;
-using System.Threading.Tasks;
 using SteamKit2.Internal;
 
 namespace SteamKit2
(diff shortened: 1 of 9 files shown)


yaakov-h:
Need to double-check for deadlocks before I can merge this.

yaakov-h:
Two further issues here:

1. `DisconnectedCallback.UserInitiated` is `true`. We might need to change the internal connection API to `bool Disconnect(bool userInitiated)`, as we currently assume that any disconnect attempt from  outside of the `IConnection` is user-initiated.
2. `UdpConnection` doesn't shut down properly. I tested this by making `CMClient` return null for `ClientLogOnResponse`, and we send a `Disconnect` packet that Steam doesn't appear to ever acknowledge. `TcpConnection` and `WebSocketConnection` behave properly.

azuisleet:
UdpConnection would try to resend forever even when a Disconnect was requested. Number 2 is fixed with  2dc0f29. I'm a little unsure about the direct usage of a new thread to call Disconnect, as the latest commit 14d883b should have the fixed the deadlock there.

yaakov-h:
The threading issue is that we are waiting for a task to finish while still within that task.

Perhaps OnClientMsgReceived should be triggered away from the connection thread/task?

yaakov-h:
I think that should do the trick.

JustArchi:
I can't even express how much I appreciate all the work you're doing ❤️

PR #513, opened 2018-01-08
Fix build failure due to Codecov

### LINKED ISSUE #512: AppVeyor Debug builds failing with Powershell/Codecov errors
```
codecov.exe : The filename, directory name, or volume label syntax is incorrect.
At line:7 char:5
+     & codecov -f "SteamKit2-code-coverage.xml" -X gcov
+     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (The filename, d...x is incorrect.:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
 
<output from codecov>

Command executed with exception: The filename, directory name, or volume label syntax is incorrect.
```

No idea why it thinks it can't launch a exe when it then appears to run.

Perhaps we should just remove code coverage - it hasn't worked since we moved to .NET Standard.

leonard-thieu:
Fixes #512 

The Python version of Codecov is causing the build to fail despite succeeding. It's replaced with the .NET version which doesn't have this issue.

This also fixes support for generating code coverage results. OpenCover cannot generate code coverage results for portable PDBs (https://github.com/OpenCover/opencover/issues/610) so the Debug build now produces Windows PDBs. In addition, OpenCover requires `mscorlib.dll` to be named `mscorlib` but that's not the case for .NET Standard (https://github.com/OpenCover/opencover/issues/595). The `-oldStyle` switch is used to work around this.

### DIFF
--- a/SteamKit2/SteamKit2/SteamKit2.csproj
+++ b/SteamKit2/SteamKit2/SteamKit2.csproj
@@ -17,7 +17,7 @@
     <DocumentationFile>bin\$(Configuration)\$(TargetFramework)\SteamKit2.xml</DocumentationFile>
   </PropertyGroup>
 
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+  <PropertyGroup Condition="'$(Configuration)'=='Debug'">
     <DebugType>full</DebugType>
     <DebugSymbols>true</DebugSymbols>
   </PropertyGroup>
--- a/SteamKit2/SteamKit2/SteamKit2.csproj
+++ b/SteamKit2/SteamKit2/SteamKit2.csproj
@@ -16,6 +16,11 @@
   <PropertyGroup>
     <DocumentationFile>bin\$(Configuration)\$(TargetFramework)\SteamKit2.xml</DocumentationFile>
   </PropertyGroup>
+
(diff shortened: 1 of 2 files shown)


SteamKit2/SteamKit2/SteamKit2.csproj
```diff
@@ -16,6 +16,11 @@
   <PropertyGroup>
     <DocumentationFile>bin\$(Configuration)\$(TargetFramework)\SteamKit2.xml</DocumentationFile>
   </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
```
yaakov-h: Perhaps this should just be `Condition="'$(Configuration)'=='Debug'"`. I know Visual Studio generates it in this form, but platform is unneccesary here.

yaakov-h:
Do you know _why_ the Python version fails and succeeds simultaneously?

I'd rather switch to the .NET version regardless - keeps the tooling etc. consistent.

leonard-thieu:
Not sure why the Python version fails the way it does. It's really odd.

yaakov-h:
Thanks!

PR #522, opened 2018-02-02
WebAPI: Do not render null/empty parameters

JustArchi:
Consider following code:

```c#
using (dynamic iEconService = WebAPI.GetAsyncInterface(IEconService, steamApiKey)) {
	iEconService.Timeout = WebBrowser.Timeout;

	response = await iEconService.GetTradeHoldDurations(
		secure: true,
		steamid_target: steamID,
		trade_offer_access_token: tradeToken
	);
}
```

`tradeToken` comes from function argument where it's optional argument set to `null` by default. This API endpoint can be called with `trade_offer_access_token`, or without.

My guess would be that if I provide argument with `null` value then it won't be encoded into resulting URL at all. Instead, I got NRE directly in SK2. This has to be fixed, depending how we want to handle this scenario:

- Just ignore `null` and empty arguments. They do not add anything at all, only pollute URL (at best), or cause NRE (at worst, when `null` is passed). This is my PR.
- Just ignore `null` arguments. Like above, but allow empty ones for some reason. This would be non-breaking behaviour, even though I can't see how encoding empty arguments in the URL can affect anything apart from how request looks like, so I don't consider this PR breaking in the first place, unless I'm wrong.
- Throw appropriate `ArgumentException` on null arguments. IMHO this makes things worse, since I'll need to either split my above function into conditional `if`, or use `trade_offer_access_token: tradeToken ?? ""` which is silly. This is current behaviour, but with (unintentional) NRE instead.

### DIFF
--- a/SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs
+++ b/SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs
@@ -269,7 +269,7 @@ namespace SteamKit2
                 }
 
                 // append any args
-                paramBuilder.Append( string.Join( "&", args.Select( kvp =>
+                paramBuilder.Append( string.Join( "&", args.Where( kvp => !string.IsNullOrEmpty( kvp.Value ) ).Select( kvp =>
                 {
                     // TODO: the WebAPI is a special snowflake that needs to appropriately handle url encoding
                     // this is in contrast to the steam3 content server APIs which use an entirely different scheme of encoding
--- a/SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs
+++ b/SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs
@@ -269,7 +269,7 @@ namespace SteamKit2
                 }
 
                 // append any args
-                paramBuilder.Append( string.Join( "&", args.Where( kvp => !string.IsNullOrEmpty( kvp.Value ) ).Select( kvp =>
+                paramBuilder.Append( string.Join( "&", args.Select( kvp =>
                 {
                     // TODO: the WebAPI is a special snowflake that needs to appropriately handle url encoding
                     // this is in contrast to the steam3 content server APIs which use an entirely different scheme of encoding
@@ -374,6 +374,12 @@ namespace SteamKit2
                     string argName = binder.CallInfo.ArgumentNames[ x ];
                     object argValue = args[ x ];
 
+                    if ( argValue == null )
+                    {
+                        // null value is not valid for any of below methods
+                        continue;
+                    }
+
                     // method is a reserved param for selecting the http request method
                     if ( argName.Equals( "method", StringComparison.OrdinalIgnoreCase ) )
                     {
@@ -408,8 +414,15 @@ namespace SteamKit2
                         continue;
                     }
 
+                    string stringValue = argValue.ToString();
 
-                    apiArgs.Add( argName, argValue.ToString() );
+                    if ( stringValue.Length == 0 )
+                    {
+                        // empty value serves no purpose in URL query string
+                        continue;
+                    }
+
+                    apiArgs.Add( argName, stringValue );
                 }
 
                 Match match = funcNameRegex.Match( binder.Name );


yaakov-h:
Are there any APIs where null is a valid value? Another option would be to emit `value=null` into the query string.

Also, where is the `NullReferenceException`? A cursory glance of the function shows that we would extract a null `value` from the `KeyValuePair`, then pass the `null` into `string.Format`, which should handle it just fine.

JustArchi:
> Are there any APIs where null is a valid value? Another option would be to emit value=null into the query string.

To the best of my knowledge such API endpoint does not exist, but even if it did, as you pointed out yourself, it's null value, not lack of value, which is what `null` means in C#. I wouldn't want for SK2 to convert my nulls into "param=null" in query string.

> Also, where is the NullReferenceException? A cursory glance of the function shows that we would extract a null value from the KeyValuePair, then pass the null into string.Format, which should handle it just fine.

This is because I didn't fix everything yesterday 😀

```
2018-02-03 13:11:25|dotnet-1432|WARN|archi|GetTradeHoldDurationForUser() System.NullReferenceException: Object reference not set to an instance of an object.
   at SteamKit2.WebAPI.AsyncInterface.TryInvokeMember(InvokeMemberBinder binder, Object[] args, Object& result)
   at CallSite.Target(Closure , CallSite , Object , Boolean , UInt64 , String )
   at System.Dynamic.UpdateDelegates.UpdateAndExecute4[T0,T1,T2,T3,TRet](CallSite site, T0 arg0, T1 arg1, T2 arg2, T3 arg3)
   at ArchiSteamFarm.ArchiWebHandler.<GetTradeHoldDurationForUser>d__43.MoveNext()
```

That will be next addition to this PR, then more testing.

JustArchi:
(And of course, the exception happens when we call `ToString()` on null string value)

yaakov-h:
Looking at it again, the null check seems to make sense, but I'm not so sure about eliding anything that `ToString()`s into an empty value.

It would be a lot easier for a consumer to debug if they see `?foo=&bar=1` rather than just `?bar=1`, and some service frameworks or hand-rolled implementations (\*cough\*valve\*cough\*) may treat an empty value differently to a missing key.

JustArchi:
I agree with the fact that some very specific/custom implementation might treat empty value as something else than no value at all.

But I still believe that there is no wanted situation in which user would intentionally want to encode no value at all and I do not believe that there is any existing use case where this actually affects anything. Valve already uses properties that could work like that, for example `active_only` in `GetTradeOffers` and it's a standard (and expected) to call those with a value that explicitly sets the property, such as `1`.

> It would be a lot easier for a consumer to debug if they see ?foo=&bar=1 rather than just ?bar=1

I think quite the opposite, it'd be far more explicit if those would not be rendered at all, as then the consumer can see why SK2 decides to omit his supplied property, only to find out that it's either `null` or empty, both of which are meaningless for the URL query string generator, and I really can't find any valid use case in which the consumer would intentionally want to pollute the URL with them.

Still, it's you deciding here, if despite of my thoughts you'd still want only `null` case handled, let me know and I can update the PR.

Ne3tCode:
Empty value is still valid.
`https://api.steampowered.com/ISteamEconomy/GetAssetClassInfo/v1/?key={your_key_here}&format=json&appid=&class_count=&classid0=`

JustArchi:
I never said that they're invalid, I said that they're meaningless.

yaakov-h:
This is being done differently in https://github.com/SteamRE/SteamKit/pull/668, where `null` values will become empty, and empty values will remain empty. If you want to omit something from the request parameters completely, then don't pass it to SteamKit in the first place.

JustArchi:
@yaakov-h I still find it a bit stupid and I don't understand why we render `null` arguments if empty arguments are exactly what should be used if consumer expects to render a parameter empty (and not `null` which defines **lack of value**).

> If you want to omit something from the request parameters completely, then don't pass it to SteamKit in the first place.

So in order to achieve what I want, I need to get rid of this perfect code:

```c#
await iEconService.GetTradeHoldDurations(
	key: steamApiKey,
	steamid_target: steamID,
	trade_offer_access_token: tradeToken // Or `tradeToken ?? ""` in case somebody would indeed want to render empty argument
);
```

And replace it with this abomination:

```c#
if (!string.IsNullOrEmpty(tradeToken)) {
	await iEconService.GetTradeHoldDurations(
		key: steamApiKey,
		steamid_target: steamID,
		trade_offer_access_token: tradeToken
	);
}

await iEconService.GetTradeHoldDurations(
	key: steamApiKey,
	steamid_target: steamID
);
```

Do you **really** force the consumer to do something like that just because he wants to avoid polluting the URL with **meaningless** arguments?

yaakov-h:
`null` is an abomination that should never have existed.

`null` is a lie that people ignore.

`null` is the computer saying "hey I have that data you want, it's at memory location zero" and all code checking for it saying "yeah, right, I don't believe you, if I go looking at memory location zero I will get my head bitten off."

When it comes to most data structures, `null` is a legitimate nothing. There is nothing there. Not an empty thing, a nothing.

Some data structures have an alternative "empty" state, like Arrays. The difference between "empty" and `null` is the difference between having an empty box, and having no box. Between having a book with no pages, and having no book at all. Between a toilet paper roll where just the cardboard core remains, and having no roll or core at all.

In most cases, this is fine. But when it comes to strings we often compose multiple strings together. Unfortunately, there is no way to compose a `null` value into a larger string. `"a" + Empty + "b" `is clearly `"ab"`, but what is `"a" + null + "b"`? I can't go looking at memory location zero to read data  that should be put there, so instead I have to make my own decision as to what to do.

The general approach taken in .NET is that a null string gets composed as an empty one. You can observe this behaviour with `"a" + null + "b"`, `string.Format("{0}", (string)null)`, or `string.Join("", new[] { "a", null, "b" })`.

In our case though, it's not straight string composition. A consumer such as yourself is feeding SteamKit a set of keys and values, either through `Dictionary<string, string>` (now `Dictionary<string, object>`), or through the `dynamic` object that turns method parameters into said dictionary.

(As an aside, I have never seen this `dynamic` approach before I started working on SteamKit, I have never seen it since, and I kind of hope I never see it again.)

So you're giving us a set of keys and values. Let's focus on just one for a moment - a single key and it's corresponding value.

We have to render this as a string `key=value`.

If you give us a null key, we cannot simply render `=value` as this becomes an invalid query string value or form-url-encoded value. It could even be interpreted as being appended to the previous value in the set. So this has to be an exception of some kind.

On the other hand, if you give us a null key, we could render `key=`, because this is the same that we would do for an empty string, and is understood by many web application frameworks already. Ideally I'd like to make it an exception, but consumers either have to omit both the key and the value when calling SteamKit, or null-coalesce to a value that the server will ignore.

We _could_ omit the entire key-value pair outright, but the consumer has given us a key to send to the server. A null value is no authorization to omit the key too. We should send that key, and it is entirely valid to do so.

Sending `key=` is not only more internally consistent with how .NET operates, but it will also aid consumers in debugging. If their request fails and they see their key-value-pair is not present in the request we send, it's not easy to discover why. On the other hand, if they see `key=`, they will be more inclined to dig through their own code to see what value is actually being passed, and then discover that it is empty or null.

So, to address your points:

> Do you really force the consumer to do something like that just because he wants to avoid polluting the URL with meaningless arguments? Do you have at least one single use case where empty argument makes any difference over lack of it?

If it's really meaningless, then there should be no problem sending `key=`. If there's no difference, then why does it matter?

On the other hand, if it does have a meaning, then you shouldn't be passing a nothing value to SteamKit - pass a proper value, or leave it out.

> And finally do you really have a reason to treat `null` and `""` the same, if they have two distinct use cases?

As above, it's because there is no clear definition for what `null` means as a substring, and we follow the .NET convention here. They don't intrinsically have two distinct use cases, rather you've assigned them different roles based upon your own interpretation, which does not match ours.

> dance around with a tambourine and fight with it to act rationally, at this point it's easier to send the request myself. I can't imagine doing this for 2 or more nullable arguments

This seems fairly melodramatic. You have plenty of simple options. Two of which instantly come to mind:

1. Use the `Dictionary<string, object>` directly, and conditionally add items to the dictionary when you expect that they may be `null`
2. Write your own wrapper over our dynamic object that passes every non-null parameter straight through and just ignores null parameters.

> There is a very good reason why you can't call `Length` on `null` string, but you can on empty one.

Because one is a string, and one is not. However, in this particular scenario, we are choosing to continue to function rather than to throw an exception trying to handle not-a-string.

> Treating both the same in the URL is stupid if you can perfectly map use cases 1:1 to be no argument render, empty argument render and value argument render.

I don't understand why we should _need_ a "no argument renderer". I don't hire a painter to not paint my house. I don't get a mechanic to not service my car. Why should I get an argument render to not render an argument?

JustArchi:
That's a good explanation, thank you.

> Use the Dictionary<string, object> directly, and conditionally add items to the dictionary when you expect that they may be null.

To the best of my knowledge you can't convert this:

```c#
await iEconService.GetTradeHoldDurations(
	key: steamApiKey,
	steamid_target: steamID,
	trade_offer_access_token: tradeToken ?? ""
);
```

To that:

```c#
var args = new Dictionary<string, object> {
	{ "key", steamApiKey },
	{ "steamid_target", steamID },
	{ "trade_offer_access_token", tradeToken ?? "" },
};

await iEconService.GetTradeHoldDurations(args);
```

I'd need to `Call()` stuff myself, and I want to avoid that if I have a convenient interface to use instead, especially considering that I'm using it everywhere else for now.

Would you enhance dynamic execution to allow passing the appropriate dictionary of args while still making use of the named function call? I'm looking at **[this](https://github.com/SteamRE/SteamKit/blob/master/SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs#L363-L395)** code and thinking something along checking `if (argValue is IDictionary<string, object> dictionary) apiArgs = dictionary;`.

yaakov-h:
`if (args.Length == 1 & args[0] is IDictionary....)` could work. I wouldn't want any arg as a dictionary to negate the rest.

yaakov-h:
@VoiDeD thoughts on passing a dictionary of args to the dynamic interface?

VoiDeD:
Sounds fine to me.

PR #527, opened 2018-03-08
fix unreachable tcp endpoints not getting marked

Longi94:
In `TcpConnection` the `CurrentEndPoint` only gets set when the connection is successful. So when the client gets a timeout when opening the connection, `CMClient`'s disconnected handler gets a null endpoint and tries to mark that as a bad server. Simple reconnection algorithms (like sample 5) will end up in an infinite loop, trying to connect to the same endpoint. Additionally the `destination` field is redundant in `TcpConnection` and only used for logging.

This PR fixes this issue.


yaakov-h:
We don't need `CurrentEndPoint = socket.RemoteEndPoint;` around line 190 anymore do we?

PR #530, opened 2018-03-22
Fix NetHook GetMessageList signature/offsets

m4dEngi:
Hope i didn't screw up something here, but looks like those are correct.


yaakov-h:
Looks correct, works fine. Thanks!

PR #539, opened 2018-04-29
SteamID: Add helper methods to convert between Chat IDs and Clan IDs

### LINKED ISSUE #453: Add helper methods to convert chat IDs
See #451.

yaakov-h:
Resolves #453.

The idea here between the two different types of methods:

`ToChatID` should only be called when you know what kind of SteamID you have. It makes absolutely no sense to go calling this on arbitrary Steam IDs.

`TryGetClanID` can be called on any kind of chat ID, but it will only work on persistent group chats. I don't feel it makes any sense to have a`TryXXX` method that handles some failure conditions but not others, so this one fails gracefully both on ad-hoc group chats and on totally-the-wrong-`EAccountType` IDs.

### DIFF
--- a/SteamKit2/SteamKit2/Types/SteamID.cs
+++ b/SteamKit2/SteamKit2/Types/SteamID.cs
@@ -552,6 +552,7 @@ namespace SteamKit2
         /// Converts this clan ID to a chat ID.
         /// </summary>
         /// <returns>The Chat ID for this clan's group chat.</returns>
+        /// <exception cref="InvalidOperationException">This SteamID is not a clan ID.</exception>
         public SteamID ToChatID()
         {
             if ( !IsClanAccount )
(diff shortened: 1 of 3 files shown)


SteamKit2/SteamKit2/Types/SteamID.cs
```diff
@@ -548,6 +548,45 @@ public string Render( bool steam3 = true )
...
 
+        /// <summary>
+        /// Converts this clan ID to a chat ID.
+        /// </summary>
+        /// <returns>The Chat ID for this clan's group chat.</returns>
+        public SteamID ToChatID()
+        {
+            if ( !IsClanAccount )
+            {
+                throw new InvalidOperationException( "Only Clan IDs can be converted to Chat IDs." );
```
VoiDeD: xml doc for this exception.

yaakov-h:
I don't think so... if we do need symmetry, then I'd drop the `ToXXX` methods and have just `TryGetXXX` methods instead.

SteamKit2/SteamKit2/Types/SteamID.cs
yaakov-h: Will do.

Issue #552, opened 2018-05-31
Crypto performance improvements


yaakov-h:
1. `AsnKeyParser` can be made a `ref struct` and can operate on `ReadOnlySpan<byte>`. Needs benchmarking to evaluate performance. This will require taking a dependency on `System.Memory`.

2. `SymmetricEncryptWithHMACIV` doesn't need a separate array to hold random values. We can fill them directly into the end of the `iv` and read them from there.

3. `SymmetricDecrypt` doesn't need to copy the entire ciphertext into a new array. We can use the `MemoryStream` constructor that accepts an array, offset and length.

yaakov-h:
4. `SymmetricEncryptWithIV` unneccesarily allocates a 16 byte array. It is unused.

5. We should be able to optimize the comparison of hmac bytes in `SymmetricDecryptHMACIV` without bringing LINQ into the equation.

6. We might be able to stackalloc the plaintext buffer in `SymmetricDecrypt`. Need to check what the usual size is.

yaakov-h:
By using `ArrayPool` from System.Buffers, without changing the API of `CryptoHelpers`, I can actually cut allocations by nearly 2/3rds.

```

**Comment by yaakov-h** [comment]
Here's another set of test results.

If we preserve objects between calls to `SymmetricEncrypt`/`SymmetricDecrypt` in combination with `ArrayPool`, we can cut memory usage drastically, CPU time not so much.

```
BenchmarkDotNet=v0.10.14, OS=Windows 10.0.17134
Intel Core i5-7600K CPU 3.80GHz (Kaby Lake), 1 CPU, 4 logical and 4 physical cores
.NET Core SDK=2.1.300
  [Host]     : .NET Core 2.1.0 (CoreCLR 4.6.26515.07, CoreFX 4.6.26515.06), 64bit RyuJIT
  DefaultJob : .NET Core 2.1.0 (CoreCLR 4.6.26515.07, CoreFX 4.6.26515.06), 64bit RyuJIT


                                Method |     Mean |     Error |    StdDev |  Gen 0 | Allocated |
-------------------------------------- |---------:|----------:|----------:|-------:|----------:|
 SymmetricEncrypt_WithIndividualArrays | 6.464 us | 0.0398 us | 0.0353 us | 2.0065 |    6320 B |
      SymmetricEncrypt_WithMinorTweaks | 6.342 us | 0.0104 us | 0.0081 us | 1.7471 |    5520 B |
[...]

Issue #547, opened 2018-05-17
DebugLog code coverage is bouncing up and down.


yaakov-h:
Lines 155 and 160-162, which specifically deal with formatting a message.

yaakov-h:
`DebugLog` is enabled by default in `DEBUG` builds.

Our unit tests run in `DEBUG` builds. Since they're not necessarily 100% deterministic, they occasionally bump into error paths which write to DebugLog.

Without going back and fixing all our interesting tests (yet), the simplest fix is to add a unit test that always covers the bouncing part of the coverage analysis.

PR #554, opened 2018-06-03
[WIP] Improve crypto performance

yaakov-h:
### Benchmark:
Benchmark run against a packet about 670 bytes long.

**Current code:**

- Time to encrypt an outgoing packet: 6.464us
- Memory allocations to encrypt an outgoing packet: 6320B
- Time to decrypt an incoming packet: 6.524us
- Memory allocations to decrypt an incoming packet: 5208B

**This branch:**

- Time to encrypt an outgoing packet: 4.054us
- Memory allocations to encrypt an outgoing packet: 848B +size of packet
- Time to decrypt an incoming packet: 3.688us
- Memory allocations to decrypt an incoming packet: 1008B + size of packet

Unfortunately it seems that about 2/3rds of the CPU time is spent computing the HMAC. I don't think we can do anything about that at the moment.

See also: #552

### DIFF
--- a/SteamKit2/SteamKit2/Networking/Steam3/NetFilterEncryptionWithHMAC.cs
+++ b/SteamKit2/SteamKit2/Networking/Steam3/NetFilterEncryptionWithHMAC.cs
@@ -5,30 +5,37 @@
 
 
 using System;
+using System.Buffers;
 using System.IO;
 using System.Security.Cryptography;
+using SteamKit2.Util;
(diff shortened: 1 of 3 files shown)


yaakov-h:
`DebugLog`'s code coverage bounces up and down of it's own volition. #547.

yaakov-h:
Closing in favour of a new draft PR.

PR #562, opened 2018-06-12
Update EPersonaState

JustArchi:
Value of `7` is new invisible mode.


VoiDeD:
Any chance you could also regen the cs files with this change as well?

JustArchi:
That should do, I ran `SteamLanguageParser`, I guess that's how you do it 🙂

VoiDeD:
Thanks!

PR #567, opened 2018-06-14
Bump CurrentProtocol

JustArchi:
I updated those by hand if it matters, latest values from beta SteamClient.

Let me know if there is anything else I should commit together with this PR, this worked for me. It doesn't look like a prerequisite for anything right now (managed to get new callbacks I wanted also on old protocol), but always best to track latest one.

### DIFF
--- a/Resources/SteamLanguage/steammsg.steamd
+++ b/Resources/SteamLanguage/steammsg.steamd
@@ -44,7 +44,7 @@ class MsgClientNewLoginKeyAccepted<EMsg::ClientNewLoginKeyAccepted>
 class MsgClientLogon<EMsg::ClientLogon>
 {
 	const uint ObfuscationMask = 0xBAADF00D;
-	const uint CurrentProtocol = 65579;
+	const uint CurrentProtocol = 65580;
 
 	const uint ProtocolVerMajorMask = 0xFFFF0000;
 	const uint ProtocolVerMinorMask = 0xFFFF;
--- a/SteamKit2/SteamKit2/Base/Generated/SteamLanguageInternal.cs
+++ b/SteamKit2/SteamKit2/Base/Generated/SteamLanguageInternal.cs
@@ -662,7 +662,7 @@ namespace SteamKit2.Internal
 		public EMsg GetEMsg() { return EMsg.ClientLogon; }
 
 		public static readonly uint ObfuscationMask = 0xBAADF00D;
-		public static readonly uint CurrentProtocol = 65579;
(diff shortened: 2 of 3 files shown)


yaakov-h:
Do we know what the side-effects are?

This usually indicates some kind of breaking change.

JustArchi:
I didn't find out any broken functionality yet, but then again I tested only logging in part for now. Would be cool if we had some unit tests for that.

Do you have any idea how we could find out breaking changes? I mean, apart from me putting this into my toys and finding out what broke the hard way 😅. I'll do that the moment I'm done with basic group chat implementation.

yaakov-h:
We can only really rely on it being an observable change.

For example, AFAIK we’ve still never seen a CMsgProtobufWrapped.

JustArchi:
I have a theory that the protocol change might be related to new chat implementation, if the bump happened in the last beta client. I didn't check what value current stable uses.

I base my theory on the fact that when you set current UI mode to chat of `2` (to enable beta chat), a lot of previously-working callbacks no longer arrive, while you're receiving new beta ones. I'm talking mainly about all callbacks related to friend messages. It also looks like Valve will at best deprecate, at worst entirely remove those by making unusable all of old functions such as sending messages, invite to chat and so on, since right now it's running in compat mode, but I don't see how they'd want to keep that obsolete code working when nobody will practically have access to those functions the moment beta update goes stable.

I mean, this is the only thing that I noticed myself. It's entirely possible that the change has nothing to do with it, but just saying what I happened to find.

yaakov-h:
I think the bump happened just prior to the new beta. @xpaw do you recall?

FWIW, Valve keep a fairly long trail of backwards compatibility since IIRC game servers don’t auto-update the Steam components, or at least didn’t used to.

xPaw:
@yaakov-h No idea when it happened, I only saw it in the friends ui.

yaakov-h:
Just checked the binaries, and current stable uses the bumped version too.

It seems that maybe we already do already know what's in the new protocol: https://github.com/JustArchi/SteamKit/blob/75b11fbc6acbaa1b4dc033c01971979a2a17b02d/Resources/SteamLanguage/steammsg.steamd#L73-L74

😕 

If you can re-target this PR against master, and everything works, I'll merge it.

@psychonic do you know what `ClientInstanceIDs` refers to?

VoiDeD:
I'd also like to revert the change to `client_package_version` - rather not
touch that field.

On Sun, Jun 17, 2018 at 5:26 PM, Nicholas Hastings <notifications@github.com

PR #568, opened 2018-06-18
CurrentProtocol bump

JustArchi:
#567 retargetted against master branch with recommended fixes.

@VoiDeD @yaakov-h Any particular reason what issues package version bump can bring and why we should stick to that initial version?


VoiDeD:
I'd be happy to discuss why on IRC if you're interested.

PR #578, opened 2018-07-21
Add ServerRecord.TryCreateSocketServer(string, out).

### LINKED ISSUE #557: No ServerRecord.CreateCreateSocketServer(string address) overload
This is inconsistent with `CreateWebSocketServer` which has a string overload, but no IPEndPoint overload, and vice versa. This makes it harder to consume these APIs.

SK itself has to parse socket servers when using it: https://github.com/SteamRE/SteamKit/blob/5fbf66012ffb8914475037c8859acc465c8b719a/SteamKit2/SteamKit2/Steam/WebAPI/SteamDirectory.cs#L64-L77

yaakov-h:
Closes #557.

New API gets dogfooded.

This is intentionally inconsistent with existing API to preserve the behaviour of not crashing on invalid input. I felt this was a better option than ```catch (ArgumentException)```.

We should probably change existing API to match.


xPaw:
Not going to add `CreateWebSocketServer(IPEndPoint)`?

yaakov-h:
I don't see any need.

Websocket servers all use DNS names - try getting an SSL certificate for an IP address.

xPaw:
Ah yeah. `DnsEndPoint`? ;)

yaakov-h:
Doable, but why?

I'm not sure it's makes sense to support object formats beyond what Valve return through the CM and Web APIs. Otherwise we could just go nuts with a million ways to create one of these.

SteamKit2/SteamKit2/Util/Utils.cs
```diff
@@ -397,6 +397,12 @@ public static bool TryParseIPEndPoint(string stringValue, out IPEndPoint endPoin
                 return false;
             }
 
+            if (port < ushort.MinValue || port > ushort.MaxValue)
```
VoiDeD: Could use unit tests for this new codepath - if we have any at all, cause I have a feeling this util isn't tested.

yaakov-h: The new tests in this PR test the greater than case. I’ll add one for the less than case.

VoiDeD: May as well test the rest too in case there's no coverage here at all.

yaakov-h: It’s indirectly covered. 
https://codecov.io/gh/SteamRE/SteamKit/pull/578/src/SteamKit2/SteamKit2/Util/Utils.cs

PR #579, opened 2018-07-21
[WIP] Change all int wrappers from class to struct

### LINKED ISSUE #559: Why is SteamID a class instead of a struct?
Just curious, is there a design reason behind this? The XML summary also reads `This 64-bit structure is used for identifying various objects on the Steam network.`, but it seems to be declared as a class.

yaakov-h:
Would close #559.

Seems we have a discrepancy:

- `default(JobID)` and all derivates should be `0xFFFFFFFFFFFFFFFF`.
- `default(SteamID)` should be `0x0000000000000000`.

We might need two definitions of `BitVector` with different defaults.


yaakov-h:
Abandoning in favour of a new draft PR

PR #582, opened 2018-07-27
Add sonarqube analysis as part of CI builds

VoiDeD:
Add sonarqube scanning and analysis as part of CI. A peek at what this gives us: https://sonarcloud.io/project/issues?branch=experiment%2Fsonarqube&id=SteamKit&resolved=false

Supposedly this will also decorate PRs, but we'll see - the GitHub integration isn't all there yet.

Phase two will probably be trimming down which files get analyzed, cause we probably don't want autogenerated code running through it. Or maybe we do. We'll see!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,20 +3,33 @@ dist: trusty
 
 language: csharp
 mono: none
-dotnet: 2.0.0
+dotnet: 2.1
+
+git:
+    depth: false
+
+addons:
+    sonarcloud:
+        organization: "steamre"
 
 solution: SteamKit2/SteamKit2.sln
 
 install:
+    - dotnet tool install --global dotnet-sonarscanner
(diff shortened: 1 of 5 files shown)


yaakov-h:
Based on that sample, this seems like a very very heavy version of something that could more simply be a set of Roslyn Analyzers, like [StyleCop](https://github.com/DotNetAnalyzers/StyleCopAnalyzers). 😕

VoiDeD:
I mean, it does technically use analyzers under the hood, but yup - that's a fair assessment. It has a couple other niceties on the dashboard, but that won't be visible until the master branch is analyzed apparently.

yaakov-h:
I'm happy to give it a shot. It seems fairly non-obtrusive, and easy enough to remove.

I just hope it doesn't scare away new developers/contributors on their first PRs.

PR #597, opened 2018-10-08
Fix cm0.steampowered.com port

seishun:
It doesn't listen on 27015.


rossengeorgiev:
```bash
$ dig +short cm0.steampowered.com
162.254.195.44
162.254.193.6
$ nc -zvnw 2 162.254.193.44 27017
nc: connect to 162.254.193.44 port 27017 (tcp) failed: Connection timed out
$ nc -zvnw 2 162.254.193.6 27017
Connection to 162.254.193.6 27017 port [tcp/*] succeeded!

yaakov-h:
It looks like Steam has gone back to having a hardcoded list.

cm0 appears to be listening on 27017, 27018 and 27019 for Valve TCP, and 27020 and 27021 for TLS.

Should we copy Steam's hardcoded list, or branch off into our own thing (like this PR)?

cc/ @azuisleet @VoiDeD 

image

xPaw:
@yaakov-h I see no reason to hardcode the list, just fix the port and be done with it :)

seishun:
What if there are connectivity issues during the first run?

azuisleet:
That wouldn't be much different than not being able to reach Steam.

PR #602, opened 2018-10-25
Improve GameID support for mods and shortcuts

psychonic:
Add support for initializing GameIDs for mods and shortcuts.

### DIFF
--- a/SteamKit2/SteamKit2/Types/GameID.cs
+++ b/SteamKit2/SteamKit2/Types/GameID.cs
@@ -74,7 +74,7 @@ namespace SteamKit2
         {
             AppID = nAppID;
             AppType = GameType.GameMod;
-            ModID = BitConverter.ToUInt32(CryptoHelper.CRCHash(System.Text.Encoding.ASCII.GetBytes(modPath)), 0);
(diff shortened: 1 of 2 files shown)


SteamKit2/SteamKit2/Types/GameID.cs
```diff
@@ -64,6 +64,36 @@ public GameID( Int32 nAppID )
...
+        /// Initializes a new instance of the <see cref="GameID"/> class.
+        /// </summary>
+        /// <param name="nAppID">The base app id of the mod.</param>
+        /// <param name="modPath">The game folder name of the mod.</param>
+        public GameID( UInt32 nAppID, string modPath )
+            : this(0)
+        {
+            AppID = nAppID;
+            AppType = GameType.GameMod;
+            ModID = BitConverter.ToUInt32(CryptoHelper.CRCHash(System.Text.Encoding.ASCII.GetBytes(modPath)), 0);
```
yaakov-h: > `BitConverter.ToUInt32(CryptoHelper.CRCHash`

So we calculate an int, convert it to bytes, and then convert it back to an int?

psychonic:
Thanks. I've switched to Crc32 now. I forgot that was even there.

I also changed to UTF8 encoding.

PR #603, opened 2018-10-25
Throw custom exception in WebAPI when request fails

xPaw:
`HttpRequestException` does not expose response or status code, so consumers are unable to access these.

I suspect other paths still can throw `HttpRequestException` besides `EnsureSuccessStatusCode`?


yaakov-h:
The new exception should subclass the existing exception so that we don’t break compatibility.

We should probably also extract just the status code and maybe headers, unless there’s a need for more, since the response message is disposable. If we throw the response then it limits changes we can make in future.

yaakov-h:
`HttpRequestException` can still be thrown, e.g. if DNS lookup fails or a similar error before we get a valid response. It should remain in the xmldoc.

Also, consider sealing the new exception type.

xPaw:
Done.

Issue #541, opened 2018-05-09
WebAPI should support proxy option


JustArchi:
Currently WebAPI offers no way for accessing `Proxy` and `UseProxy` properties of internal `HttpClient`, making it impossible to supply those. Such option can be beneficial for people behind firewalls (e.g. in China) that can't access several Steam web-services, especially `steamcommunity.com` in a normal way. I'm not sure if API is restricted itself, but it definitely won't hurt to have such possibility.

This issue will be entirely resolved "by the way" while working on https://github.com/SteamRE/SteamKit/issues/450, therefore it might make sense to get down to that one instead and achieve 2 in 1.

yaakov-h:
Yep, #450 should cover this. I’m hoping to start tackling that in the coming weeks.

yaakov-h:
Done in #605.

PR #605, opened 2018-10-28
Make HTTP components configurable

### LINKED ISSUE #450: Making it possible for using custom HttpClient in WebAPI?

yaakov-h:
Users can provide one or two new factory functions to `SteamConfiguration`:

- A `HttpMessageHandler` factory function, which is the inner part of the pipeline that actually executes HTTP requests, and
- A `HttpClient` factory function, which is a nice wrapper class that the rest of the .NET HTTP stack is built around.

Users can use these two injection points to, for example:

- Configure credentials
- Configure proxy settings
- Do something instead of HTTP requests
- Inspect and log outgoing requests
- Inspect and log incoming responses
- Mutate requests and responses (though any errors caused from this are their own fault)
- Use a single underlying HTTP connection pool to increase performance
- Use a single underlying HTTP connection pool to avoid connection exhaustion

Note that this only affects WebAPI, it has no bearing on Websocket CM connections. We could expose `ClientWebSocketOptions` some day, but that's another PR for another issue, if/when it's ever needed.

This resolves #450 and #541.

For now I've simply attached them to `SteamConfiguration`, but as the configuration gets bigger we might want to split it up - for example, these two factory methods could be put on an interface, we provide a default implementation, and `SteamConfiguration` would only have one HTTP-related method/property to get/replace the HTTP configuration object.

### DIFF
--- a/SteamKit2/SteamKit2/Steam/WebAPI/SteamConfigurationWebAPIExtensions.cs
+++ b/SteamKit2/SteamKit2/Steam/WebAPI/SteamConfigurationWebAPIExtensions.cs
@@ -42,23 +42,11 @@ namespace SteamKit2
 
         internal static HttpClient GetHttpClientForWebAPI(this SteamConfiguration config)
         {
-            var client = config.GetHttpClient();
+            var client = config.HttpClientFactory();
 
             client.BaseAddress = config.WebAPIBaseAddress;
             client.Timeout = WebAPI.DefaultTimeout;
 
-            return client;
-        }
-
-        internal static HttpClient GetHttpClient(this SteamConfiguration config)
-        {
-            if (config == null)
-            {
-                throw new ArgumentNullException(nameof(config));
-            }
-
-            var handler = config.HttpMessageHandlerFactory();
-            var client = config.HttpClientFactory(handler);
             return client;
         }
     }
--- a/SteamKit2/SteamKit2/Steam/WebAPI/SteamConfigurationWebAPIExtensions.cs
+++ b/SteamKit2/SteamKit2/Steam/WebAPI/SteamConfigurationWebAPIExtensions.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Net.Http;
 
 namespace SteamKit2
 {
@@ -20,7 +21,7 @@ namespace SteamKit2
                 throw new ArgumentNullException(nameof(config));
             }
 
-            return new WebAPI.Interface(config.WebAPIBaseAddress, iface, config.WebAPIKey);
+            return new WebAPI.Interface(config.GetHttpClientForWebAPI(), iface, config.WebAPIKey);
         }
 
         /// <summary>
@@ -36,7 +37,29 @@ namespace SteamKit2
                 throw new ArgumentNullException(nameof(config));
             }
 
-            return new WebAPI.AsyncInterface(config.WebAPIBaseAddress, iface, config.WebAPIKey);
+            return new WebAPI.AsyncInterface(config.GetHttpClientForWebAPI(), iface, config.WebAPIKey);
+        }
+
+        internal static HttpClient GetHttpClientForWebAPI(this SteamConfiguration config)
+        {
+            var client = config.GetHttpClient();
+
+            client.BaseAddress = config.WebAPIBaseAddress;
+            client.Timeout = WebAPI.DefaultTimeout;
+
+            return client;
+        }
+
+        internal static HttpClient GetHttpClient(this SteamConfiguration config)
+        {
+            if (config == null)
+            {
+                throw new ArgumentNullException(nameof(config));
+            }
+
+            var handler = config.HttpMessageHandlerFactory();
+            var client = config.HttpClientFactory(handler);
+            return client;
         }
     }
 }
--- a/SteamKit2/SteamKit2/Steam/CDNClient.cs
+++ b/SteamKit2/SteamKit2/Steam/CDNClient.cs
@@ -231,7 +231,7 @@ namespace SteamKit2
             }
 
             this.steamClient = steamClient;
-            this.httpClient = steamClient.Configuration.GetHttpClient();
+            this.httpClient = steamClient.Configuration.HttpClientFactory();
 
             this.depotIds = new ConcurrentDictionary<uint, bool>();
             this.appTicket = appTicket;
--- a/SteamKit2/SteamKit2/Steam/CDNClient.cs
+++ b/SteamKit2/SteamKit2/Steam/CDNClient.cs
@@ -231,7 +231,7 @@ namespace SteamKit2
             }
 
             this.steamClient = steamClient;
-            this.httpClient = new HttpClient();
+            this.httpClient = steamClient.Configuration.GetHttpClient();
 
             this.depotIds = new ConcurrentDictionary<uint, bool>();
             this.appTicket = appTicket;
--- a/SteamKit2/SteamKit2/Steam/SteamClient/Configuration/ISteamConfigurationBuilder.cs
+++ b/SteamKit2/SteamKit2/Steam/SteamClient/Configuration/ISteamConfigurationBuilder.cs
@@ -53,15 +53,6 @@ namespace SteamKit2
         ISteamConfigurationBuilder WithHttpClientFactory(HttpClientFactory factoryFunction);
 
         /// <summary>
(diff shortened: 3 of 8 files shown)


xPaw:
Some things that I feel like should be considered, both WebAPI and CDNClient expose timeouts differently, and I think it could be handled with this.

https://github.com/SteamRE/SteamKit/blob/4360d6f1d9137b867b022cc09563c3514985b22b/SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs#L179-L182

https://github.com/SteamRE/SteamKit/blob/aa3e7f1750644286a8f4ee144992f80d157b1f71/SteamKit2/SteamKit2/Steam/CDNClient.cs#L213

And why does CDNClient use cancellation token instead of httpclient's timeout properties?

SteamKit2/SteamKit2/Steam/WebAPI/SteamConfigurationWebAPIExtensions.cs
```diff
@@ -36,7 +37,29 @@ public static WebAPI.AsyncInterface GetAsyncWebAPIInterface(this SteamConfigurat
...
 
-            return new WebAPI.AsyncInterface(config.WebAPIBaseAddress, iface, config.WebAPIKey);
+            return new WebAPI.AsyncInterface(config.GetHttpClientForWebAPI(), iface, config.WebAPIKey);
+        }
+
+        internal static HttpClient GetHttpClientForWebAPI(this SteamConfiguration config)
+        {
+            var client = config.GetHttpClient();
+
+            client.BaseAddress = config.WebAPIBaseAddress;
```
JustArchi: Won't this cause issues if consumer is using HttpClient for his own requests with custom base address (or lack of it)?

xPaw: I would agree here, It should be easy enough to just pass base address into `urlBuilder` directly instead of using `HttpClient.BaseAddress`.

yaakov-h: The xmldoc calls this out by differentiating between “returns a new or existing XXXXX” and “creates a new XXXXX”. 

Sharing an http client outside of our own library would drastically limit how much of its functionality we can actually use, both now and into the future.

xPaw: Well, SK2 itself will share HttpClient in different places (e.g with CDNClient).

yaakov-h: IIRC a single CDNClient has a single HttpClient, so I wouldn't call that sharing.

JustArchi: It depends on general approach and preference, but I wouldn't really call it limitation. I don't see any use case in which you'd be somehow limited from doing a particular action just because you're not expected to edit behaviour of `HttpClient`. I mean, right now SK2 modifies only two of those properties, `BaseAddress` and `Timeout`, out of which one is pretty optional to use and can be achieved with url building logic, while second one is customizable already.

One `HttpClient` could be shared across several different classes and functionalities if wanted, question is if we do want it to happen. On the low level, connection pool is already shared with handler, so I don't have any argument after or against, merely a preference. I don't see a situation in which you'd want to have different `HttpClient` settings for different functions, apart from `BaseAddress` which is the reason why we don't have a single `HttpClient` already.

Maybe it'd make sense to consider this, unless I'm unaware of some specific `HttpClient` setting that would vouch against sharing a single instance. I took a look and there are only headers and buffer size right now.

yaakov-h: Well that's the settings we have to play with, but we could also, perhaps in future, need to wrap the user-supplied HttpMessageHandler with one or more of our own (e.g. a `DelegatingHandler` subclass), and then you _have_ to have a different HttpClient instance to hold that handler.

Also, if we shared a HttpClient, SteamKit requests could never have a longer timeout set than the rest of the user's application. We could only have a shorter one, if we (or the user) so desired.

It just feels like we'd be painting ourselves into a corner for no extra gain.

JustArchi: This is a fair point, if you can see how it'd bite us in the future then I see no need to do any of that. In fact, I'd even vouch for complete removal of custom `HttpClient` and leave handler only, but sadly there are `HttpClient` settings that the consumer might want to supply, mainly default request headers for identification purposes.

Maybe it could be a good idea to allow consumer to supply only its own `HttpClientHandler` and default headers, leaving creating `HttpClient` to SK2. I don't see any other reason for one to force usage of its own `HttpClient`, and because SK2 is expected to modify its properties anyway, it'd avoid a lot of confusion (even if it's stated in the doc).

JustArchi: It feels fine for me to supply message handler, default headers and value specifying that my handler is supposed to be reused in my own routines further, so it shouldn't be disposed.

Considering that there is really not much more to add towards that (I mean, you've defined everything regarding how to create and use `HttpClient` by now), I feel that it's a good tradeoff for having a guarantee that `HttpClient` created by SK2 has expected state and no issues as long as user can guarantee that his message handler is proper.

JustArchi: Still, this is more like a design choice over personal preferences rather some crucial decision, so I leave that up to you, the PR is fine as it is, I just don't see a reason why consumer must create `HttpClient` and ensure it meets all SK2 requirements and can be modified, if SK2 can pretty much do the same with just 1 more argument (headers + disposing vs httpclient alone).

yaakov-h: If the consumer:

- Doesn't supply their own message handler factory
- Supplies a new message handler each time

Then they don't need to create a `HttpClient`. If they want more advanced control, then I'd rather give them more advanced control rather than re-inventthe wheel.

For example, they could use [`HttpClientFactory`](https://github.com/aspnet/HttpClientFactory) with few (if any) modifications.

yaakov-h:
I'm starting to question whether or not we need `Func<HttpMessageHandler>`... could we just have `Func<HttpClient>` as the only overridable function?

yaakov-h:
Hmmm.

I did some more thinking and, in theory, we could possibly maybe some day want to insert our own `DelegatingHandler` into the chain.

However, what we could do then, is:

- Construct our own `HttpClient` with our own `HttpMessageHandler`
- Have our own `HttpMessageHandler` do whatever it needs to do, then delegate the actual sending and recieving of the HTTP message to the consumer-provided `HttpClient`

It's two more object allocations, but if that be comes a problem then we can optimize later down the road some day. No point whatsoever doing that prematurely.

So yeah, it looks like we only need the one of these extensibility points.

I'll refactor.

yaakov-h:
Any last thoughts or comments?

rossengeorgiev:
image

Jillbot has spoken, close dis.

PR #607, opened 2018-11-03
Add Balance64 to WalletInfoCallback

JustArchi:
This can be needed if `int` won't be enough to display the account balance, i.e. when balance reaches more than `21474836.47` in local currency.

Unlikely to happen, but won't hurt to add considering Steam provides this.

### DIFF
--- a/SteamKit2/SteamKit2/Steam/Handlers/SteamUser/Callbacks.cs
+++ b/SteamKit2/SteamKit2/Steam/Handlers/SteamUser/Callbacks.cs
@@ -305,7 +305,7 @@ namespace SteamKit2
             /// <summary>
             /// Gets the balance of the wallet as a 64-bit integer, in cents.
             /// </summary>
-            public long Balance64 { get; private set; }
+            public long LongBalance { get; private set; }
 
 
             internal WalletInfoCallback( CMsgClientWalletInfoUpdate wallet )
@@ -314,7 +314,7 @@ namespace SteamKit2
 
                 Currency = ( ECurrencyCode )wallet.currency;
                 Balance = wallet.balance;
-                Balance64 = wallet.balance64;
+                LongBalance = wallet.balance64;
             }
         }
 
--- a/SteamKit2/SteamKit2/Steam/Handlers/SteamUser/Callbacks.cs
+++ b/SteamKit2/SteamKit2/Steam/Handlers/SteamUser/Callbacks.cs
@@ -296,10 +296,16 @@ namespace SteamKit2
             /// Gets the currency code for this wallet.
             /// </summary>
             public ECurrencyCode Currency { get; private set; }
+
             /// <summary>
-            /// Gets the balance of the wallet, in cents.
+            /// Gets the balance of the wallet as a 32-bit integer, in cents.
             /// </summary>
             public int Balance { get; private set; }
+
+            /// <summary>
+            /// Gets the balance of the wallet as a 64-bit integer, in cents.
+            /// </summary>
+            public long Balance64 { get; private set; }
 
 
             internal WalletInfoCallback( CMsgClientWalletInfoUpdate wallet )
@@ -308,6 +314,7 @@ namespace SteamKit2
 
                 Currency = ( ECurrencyCode )wallet.currency;
                 Balance = wallet.balance;
+                Balance64 = wallet.balance64;
             }
         }
 


VoiDeD:
`LongBalance` maybe? To match convention with things like `Array.LongLength`.

JustArchi:
Sure, why not, if you consider it better.

SteamKit2/SteamKit2/Steam/Handlers/SteamUser/Callbacks.cs
```diff
@@ -296,18 +296,25 @@ public sealed class WalletInfoCallback : CallbackMsg
             /// Gets the currency code for this wallet.
             /// </summary>
             public ECurrencyCode Currency { get; private set; }
+
             /// <summary>
-            /// Gets the balance of the wallet, in cents.
+            /// Gets the balance of the wallet as a 32-bit integer, in cents.
```
JustArchi: I know that stating data type in the doc is a bit excessive and unneeded, but I wanted to point out the difference between those two for people that might not recall at first. I think it's OK in this case where consumer might wonder which one to use.

yaakov-h: Do we know what actually happens to this field when an account has a balance of $42m?

xPaw: Is it in USD or users currency?

Because with vietnamese dong a game can cost 1.300.000₫ which is 130000000 cents.

yaakov-h: Well do we know what happens in that scenario too?

xPaw: No clue, but I pointed it out to ease checking it out if anyone wants to.

yaakov-h: @JustArchi Just saw your comment - yeah, but it's Valve. I'd be quite unsurprised if it's just the lower 32 bits of the actual balance, and wraps around to the negatives.

JustArchi: @yaakov-h you're totally right and it also wouldn't shock me, but I'm not rich enough to check 😀.

And BTW yes, this balance is in local currency (`ECurrencyCode`) which is user's local currency, if supported by Steam. It counts from full cents or other smallest monetary value.

yaakov-h:
I think "in cents" in the xmldoc is slightly misleading, because with some `ECurrencyCode` values there are no cents... but that's how the old xmldoc was, so ¯\_(ツ)_/¯.

xPaw:
> because with some ECurrencyCode values there are no cents

They're all still multiplied by 100, aren't they?

yaakov-h:
I don't think Yen has a hundredth value.

JustArchi:
It depends on the currency. For "traditional" western currencies like USD, CAD, PLN, EUR, the minimum monetary value is 0.01 of that currency and this amount is what Steam uses as integer. In other words, 10 USD will have `ECurrencyCode` of `USD` and `1000` as a balance (and balance64).

Not all currencies follow traditional way though, yen is a good example of that, there is no 0.01th of yen. I'm actually wondering now if Valve decided to go easy route and just record everything as * 100, or there is per-currency-parsing where yen balance should not be divided by 100 to present actual one.

If somebody is interested in checking that out, I'd be interested in knowing that as well, because that routine has to be included in the Steam client itself.

yaakov-h:
> I'm actually wondering now if Valve decided to go easy route and just record everything as * 100, or there is per-currency-parsing where yen balance should not be divided by 100 to present actual one.

From IRC:

```
08:45 <+xPaw>  Netshroud, https://github.com/SteamDatabase/SteamTracking/blob/85cd023702c77d200245f14bbc040e11003bf7f7/Scripts/Community/global.js#L307
08:45 <+xPaw>  they store all currencies as 'cents', even if said currency has no hundreds
08:46 <Netshroud>  so they store a hundredth of a yen? O_o
08:47 <+xPaw>  its multiplied for storage
08:47 <+xPaw>  divided for display
08:47 <Netshroud>  bWholeUnitsOnly
08:48 <Netshroud>  oh come on
08:48 <+xPaw>  yeah thats solely for display
08:48 <Netshroud>  that doesn't affect multiplier, just the trailing ".00"
08:48 <Netshroud>  VALVE!!
```

Issue #16, opened 2013-08-28
Move the protobufs into a separate repo


seishun:
Many projects use Steam .proto files in one way or another. Some maintain local copies of the protobufs, e.g. [SKSteamKit](https://github.com/yaakov-h/SKSteamKit), while others, including [node-dota2](https://github.com/seishun/node-dota2), [go-steam](https://github.com/Philipp15b/go-steam) and [Steam++](https://github.com/seishun/SteamPP), link to this repo. With the former approach, the protos will be constantly out of date, while the latter requires the user to clone the pretty huge SteamKit2 repo.

It would greatly benefit the aforementioned projects if the .proto files were in their own repository. This would ensure a single central source for the protobufs that's easy to use as a dependency.

paralin:
DOTA2 protobufs are now kept up to date in http://github.com/paralin/Dota2 which is now referenced by node-dota2

SinZ163:
Theres also the http://github.com/SteamDatabase/GameTracking repo with maintained protobufs

Netshroud:
Those both have the same drawbacks as cloning SteamKit.

paralin:
That's not exactly true, my repo is tiny in comparison

xPaw:
So I want to make a separate repository entirely just for protobufs which will be synced by @SteamTracker automatically.

Currently it will track steamclient, tf2, csgo, dota2. Are these enough to be in a single repository?

psychonic:
Protobufs are now used from the SteamDB repo mentioned above. I'm not sure if there's still any consideration being done to move anything else to separate repos.

PR #610, opened 2018-11-25
Use protobufs from submodule

psychonic:
As discussed from time to time, use SteamDB's Protobufs submodule instead of continuing to maintain our own. This makes keeping up to date simpler.

Using
https://github.com/SteamDatabase/Protobufs

Since we can't have individual generate scripts for each app in the folder with their proto files, instead switched to single script to regenerate all, running from the Protogen folder. If that's not desired, we can still do individual bats, just ran from elsewhere. This just removed much of the boilerplate needed.

Also, with this change, the classes for the new Steam chat are using the protos from Steam Client itself, rather than from the friends js.


xPaw:
I suggest removing `gc.proto` and using the proto from `steam/steammessages_base.proto` instead.

That has all the new fields, except for 200/201 which I suspect are only sent between GC servers anyway.

yaakov-h:
Note to self: close #16 on merge.

Resources/Protogen/generate-all.ps1
```diff
@@ -0,0 +1,59 @@
...
+param([string[]]$ProtoDir)
+
+$ProtoGenDir = $PSScriptRoot
+$ProtoGenExe = Join-Path $PSScriptRoot 'protogen.exe'
+$ProtoBase = Join-Path $PSScriptRoot '..\ProtoBufs'
+$SK2Base = Join-Path $PSScriptRoot '..\..\SteamKit2\SteamKit2\Base\Generated'
+
+Push-Location
+
+$protos = Import-Csv -LiteralPath (Join-Path $PSScriptRoot 'protos.csv') |
```
yaakov-h: Neat!

yaakov-h:
Probably something for a later stage, but I'd love to have this automatically submit a PR when the protobufs repo gets updated.

xPaw:
@yaakov-h That proto only exists in CS:GO now: https://github.com/SteamDatabase/Protobufs/search?q=CSOPartyInvite&unscoped_q=CSOPartyInvite

It was generated from tf2 beta before.

PR #612, opened 2018-11-26
Remove protobuf package reference from samples

xPaw:
All samples reference SteamKit project, and only these 2 samples listed protobuf directly.


yaakov-h:
Seems like this was an artefact of the old csproj/sdk system. The new one seems to handle transient dependencies of project dependencies just fine.

PR #616, opened 2018-12-17
WIP: Initial implementation of content delta patches

xPaw:
Opening a PR for tracking and discussion.

All the code is in steamclient, reference functions:
- ContentDeltaChunks proto
- CDepotReconstructCtx::BInitializeDeltaChunks
- CContentDeltaChunks::BDeserializeProtobuf

I haven't looked into what `patch_method` is yet.

Example: `DownloadPatchAsync( 441, 1210411661116720898, 6498161485441060059 );` (with depot key, and cdn auth set)

### DIFF
--- a/SteamKit2/SteamKit2/Steam/CDNClient.cs
+++ b/SteamKit2/SteamKit2/Steam/CDNClient.cs
@@ -20,7 +20,7 @@ namespace SteamKit2
     /// <summary>
     /// The CDNClient class is used for downloading game content from the Steam servers.
     /// </summary>
-    public sealed class CDNClient : IDisposable
+    public sealed partial class CDNClient : IDisposable
     {
         /// <summary>
         /// Represents a single Steam3 'Steampipe' content server.
--- a/SteamKit2/SteamKit2/Steam/CDNClient_Patch.cs
+++ b/SteamKit2/SteamKit2/Steam/CDNClient_Patch.cs
@@ -1,0 +1,101 @@
+﻿/*
+ * This file is subject to the terms and conditions defined in
+ * file 'license.txt', which is part of this source code package.
+ */
+
+using System;
+using System.IO;
+using System.Net.Http;
+using System.Threading.Tasks;
+
+namespace SteamKit2
+{
+    public sealed partial class CDNClient : IDisposable
+    {
+        const uint MAGIC_PATCH_HEADER = 0x502F15E5;
+        const uint MAGIC_PATCH_FOOTER = 0xF3DEA982;
+
+        /// <summary>
+        /// Downloads the depot delta patch specified by the given manifest IDs.
+        /// </summary>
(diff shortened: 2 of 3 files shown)


SteamKit2/SteamKit2/Steam/CDNClient.cs
```diff
@@ -20,7 +20,7 @@ namespace SteamKit2
     /// <summary>
     /// The CDNClient class is used for downloading game content from the Steam servers.
     /// </summary>
-    public sealed class CDNClient : IDisposable
+    public sealed partial class CDNClient : IDisposable
```
xPaw: CDNClient is pretty big already, is it okay to use partials?

SteamKit2/SteamKit2/Steam/CDNClient_Patch.cs
```diff
@@ -0,0 +1,101 @@
...
+            {
+                throw new ArgumentNullException( nameof( depotKey ) );
+            }
+
+            if ( sourceManifestId == targetManifestId )
+            {
+                throw new ArgumentException( "Source manifest cannot equal target manifest" );
+            }
+
+            // TODO: This request can return 404 with html content, need to account for that
```
xPaw: I haven't checked if the status code is actually 404, but the content is a `<html> [...] Not Found` error so the decryption fails on that currently.

xPaw:
@azuisleet Have you looked into applying delta patches yet?

rossengeorgiev:
Looking at `CWorkItemDepotDecrypt::ThreadProcess`, the valid values for `patch_method` are `1` and `2`.
- `1` decompress using `VZa` or `zip` (just like regular chunks)
- `2` decompress using `VZd`. See `VZDecompressToBufferWithPresetDict` (yet to figure this out)

Followed by `CDepotReconstruct::OnChunkReadUnpacked` magic

xPaw:
Delta updates are not dead, and they added `EContentDeltaChunkDataLocation`
https://github.com/SteamDatabase/Protobufs/commit/9e5182872555f968bf9ff08b752ecde9e5d2fb88

fireundubh:
Why was this closed? @xPaw

xPaw:
I have no interest in reverse engineering the actual delta patching algorhitms. Feel free to pick it up.

fireundubh:
> I have no interest in reverse engineering the actual delta patching algorhitms. Feel free to pick it up.

Why not diff the target newer manifest against each intermediate manifest, if any, and finally against the delta older manifest? This is super straightforward since each manifest contains hashes. From a DepotDownloader perspective, the main trick is you need an ordered index of manifests.

SteamDB seems to have that info, but DepotDownloader doesn't appear to offer a way to get it. If you could though, you wouldn't even need to implement delta support - because that could be accomplished via the CLI via Python, etc.

PR #624, opened 2019-01-11
Use TLS 1.2 in Nethook's SetupDependencies.ps1

Longi94:
Powershell uses TLS 1.0 which is not supported by github.com anymore.

### DIFF
--- a/Resources/NetHook2/SetupDependencies.ps1
+++ b/Resources/NetHook2/SetupDependencies.ps1
@@ -6,7 +6,9 @@ $ProtobufSourceZipUrl = "https://github.com/google/protobuf/releases/download/v2
 $ProtobufSourceFile = [System.IO.Path]::Combine($NetHook2DependenciesTemporaryDirectory, "protobuf.zip")
 $ProtobufSourceInnerFolderName = "protobuf-2.5.0"
 
-[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
+if ([Net.ServicePointManager]::SecurityProtocol -notcontains 'Tls12') {
+    [Net.ServicePointManager]::SecurityProtocol += [Net.SecurityProtocolType]::Tls12
+}
 
 Set-Location $PSScriptRoot
 
--- a/Resources/NetHook2/SetupDependencies.ps1
+++ b/Resources/NetHook2/SetupDependencies.ps1
@@ -6,6 +6,8 @@ $ProtobufSourceZipUrl = "https://github.com/google/protobuf/releases/download/v2
 $ProtobufSourceFile = [System.IO.Path]::Combine($NetHook2DependenciesTemporaryDirectory, "protobuf.zip")
 $ProtobufSourceInnerFolderName = "protobuf-2.5.0"
 
+[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
+
 Set-Location $PSScriptRoot
 
 if (-Not (Test-Path $NetHook2DependenciesTemporaryDirectory))


JustArchi:
That's one of the dumbest arguments I've ever heard for not including 2 lines of code more in order to be safe against whatever PS devs decide to stick with in the future, especially considering that there is not a single reason to **not** do it the correct way to begin with.

But whatever, I'm not the one deciding here.

yaakov-h:
I agree with Archi. We should add to what is supported, not replace it exclusively.

Longi94:
I've misunderstood Archi's suggestion, sorry about that. I've made the changes.

yaakov-h:
If I can address the base issue here:

The problem seems to be that the script does not work because of TLS issues?

I'm not seeing evidence of this problem at the moment. This script is used in our CI, which is [still working](https://ci.appveyor.com/project/SteamRE/steamkit/builds/21449907/job/55f3rxj990xqqa71#L114), and actually [throws an error](https://ci.appveyor.com/project/SteamRE/steamkit/builds/21571772/job/150dht3vmbdv9lyd#L118) with this PR's changes.

What's the actual scenario here that you're trying to improve?

yaakov-h:
Considering that SteamKit targets .NET Standard 2.0, which requires .NET 4.6.1 (with SDK hacks), or 4.7.2 (without them), I don't think .NET 4.5 is something we should be concerned with.

JustArchi:
@yaakov-h that error is thrown when the enum flags already include the flag in `+=`, since CI includes Tls12 already, this is why it started to throw now. I wonder why my `if` didn't work properly, I could swear I used something similar a while back.

In any case, if you want to fix this PR then only `if` should be fixed to not execute when `Tls12` already exists. However, I also find it quite useless to fix considering SK2 doesn't support this setup.

Longi94:
I've tried to run the script locally and the https request throws an error and adding the line fixes it, locally at least. .NET 4.7.2 is installed on my computer.

```PowerShell
PS C:\Users\lngtr> $PSVersionTable.PSVersion

Major  Minor  Build  Revision
-----  -----  -----  --------
5      1      17134  407


PS C:\Users\lngtr> Invoke-WebRequest "https://github.com/google/protobuf/releases/download/v2.5.0/protobuf-2.5.0.zip" -O
utFile "test.zip"
Invoke-WebRequest : The request was aborted: Could not create SSL/TLS secure channel.
At line:1 char:1
+ Invoke-WebRequest "https://github.com/google/protobuf/releases/downlo ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (System.Net.HttpWebRequest:HttpWebRequest) [Invoke-WebRequest], WebExc
   eption
    + FullyQualifiedErrorId : WebCmdletWebResponseException,Microsoft.PowerShell.Commands.InvokeWebRequestCommand

PS C:\Users\lngtr>
```

psychonic:
@Longi94 what version of Windows are you on? It's possible that PowerShell 5.1 for older than Windows 10 still targets .NET Framework 4.5.x, leading to that issue.

Longi94:
Windows 10 build 17134. Since these are flags, a bitwise OR could do the trick too.
`[Net.ServicePointManager]::SecurityProtocol |= [Net.SecurityProtocolType]::Tls12`

JustArchi:
Yeah but I'm not a big fan of not making it obvious enough that we need it only when we don't have it already, and that we're not for example enabling it unconditionally.

yaakov-h:
I'd be fine with `|=`, but I'm still confused as to why it's a problem in the first place, and I'd like to understand it before working around it.

17134 is v1803, which is the second-most-recent release.

Are you using x64 Powershell, or x86?

Longi94:
x64. But I've just tried with x86 and it throws the same error.

Longi94:
With a quick google search it seems like it's a continuing issue with .NET 4.5.0 and above and the only way to make tls 1.2 default is editing a registry, which the CI might already have.

JustArchi:
I didn't edit any registry and I do not suffer from your issue.

Longi94:
I've updated Windows to build 17134 (v1809 I believe) and the issue is gone. If a fix like this is unnecessary, feel free to close this PR.

yaakov-h:
Closing as this was a runtime issue, and will be counter-productive when TLS 1.3 becomes commonplace.

PR #628, opened 2019-01-18
Update to latest protogen

xPaw:
I patched protogen with the following:

- Skip generating `Extensions`
- Always generate `List<>` instead of arrays types to match current behaviour
- Force proto2 if syntax is not specified

The patch file is attached in the PR.

Things that need to be done:
- [x] Fix `DefaultValue` - https://github.com/mgravell/protobuf-net/issues/495 https://github.com/mgravell/protobuf-net/pull/516
- [x] Generate interfaces for unified services like `IPlayer` - https://github.com/mgravell/protobuf-net/issues/487 https://github.com/mgravell/protobuf-net/pull/492


yaakov-h:
I really don't like patching the binaries, and I don't think it will be maintainable going forwards.

Can we follow the original suggestion from the other issue and have our own generator frontend that's more tailored for our use?

xPaw:
The patch is a proof of concept, as it currently compiles steamkit but throws during serialisation.

mgravell:
Oof, yeah, that's just a huge number overflow. I'm *guessing* that adding
UL as a suffix to the number will fix it. Any chance you could test that
hypothesis? I can update protogen if it is that.

xPaw:
Adding `UL` does not fix it, and I just checked, it passes the value into `DefaultValue(float)` overload, as there doesn't appear to be any unsigned overloads: https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.defaultvalueattribute.-ctor?view=netframework-4.7.2

mgravell:
K, I'll need to take a look at options, then. I guess at a push the
type+string overload might work. Sheesh, all more fun.

xPaw:
Turns out protogen already had the underlying support for writing services, the c# generator simply didn't implement them. So I implemented the interface generation, it skips any options though, but it should be good enough for SteamKit.

Technically speaking this is now fully compatible with SteamKit (I'm not sure if skipping DefaultValue will have any issues or not).

mgravell:
@xPaw  I'd be happy to look at a PR, if the service code would be more generally useful

xPaw:
Yeah I made a PR in the protobuf repo.

mgravell:
ah, k; I'll check that later; btw - if you don't want to keep having to upload binaries, all of the codegen code is available as a package ref to [protobuf-net.Reflection](https://www.nuget.org/packages/protobuf-net.Reflection/) - `protogen` is a trivial wrapper that just nudges the lib

xPaw:
Well the Reflection binary is where all the patches are.

yaakov-h:
Fixing `[DefaultValue]` seems like it should be a trivial PR to protobuf-net.

We can't easily implement service generation ourselves (by subclassing `ProtoBuf.Reflection.CSharpCodeGenerator`) because we need access to some private and internal members such as `IType`, `NullIfInherit` and `FindNameFromCommonAncestor`.

We can do it, but with reflection hackery and duplicated code copied from from `protobuf-net.Reflection`.

@mgravell Can we make those types/members part of public API, or do we have to upstream this, keeping in mind that it may conflict with / need to be switchable with gRPC as you mentioned in https://github.com/mgravell/protobuf-net/pull/492?

In the meantime I'm tempted to proceed with reflection...

xPaw:
Closing in favor of #686

PR #637, opened 2019-02-08
Add .editorconfig file

### LINKED ISSUE #382: Add .editorconfig
See https://github.com/dotnet/corefx/blob/master/.editorconfig

yaakov-h:
Lots of whitespace changes, I hope nobody minds.

Resolves #382.

I've also silenced (`:silent`) a bunch of the simplification rules with the intent that we can un-silence and fix them separately. In the meantime it was just hundreds of diagnostics as noise.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,60 @@
+root = true
+
+[*]
+indent_style = space
+insert_final_newline = true
+indent_size = 4
+
+[*.csproj]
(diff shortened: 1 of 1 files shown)


yaakov-h:
As discussed in IRC, scope changed to only add editorconfig. Code will be reformatted over time as it is touched.

PR #652, opened 2019-03-27
Throw exception in CDNClient if http request is not successful

### LINKED ISSUE #650: CDNClient no longer throws exceptions
Beyond HttpRequestException for network exceptions, WebException isn't being thrown. Instead, the status code is available in HttpResponseMessage. The status code is currently being lost.

xPaw:
Fixes #650


yaakov-h:
Does `CDNClient` need a unit test? 😃

azuisleet:
This will be a breaking change.

yaakov-h:
Yes.

It should be a less-breaking? change if we have `WebAPIRequestException -> SteamKitWebRequestException -> HttpRequestException`, but if/when CoreCLR gets it's act together I'd like to drop both custom exception types and just use `HttpRequestException`...

yaakov-h:
@xPaw what do you think of making this non-breaking?

xPaw:
Fixed by #672

PR #653, opened 2019-03-28
Update and regenerate protobufs

xPaw:
`steammessages_base` uses `oneof` which protogen is unable to generate, so I manually edited it out into normal optional fields.


yaakov-h:
Where are we up to with newer protogen / proper `oneof` support?

yaakov-h:
```
DotaClient.cs(206,25): error CS0246: The type or namespace name 'EGCSystemMsg' could not be found
```

ruh-roh

xPaw:
> Where are we up to with newer protogen / proper oneof support?

I believe my PR is ready to go if you're willing to accept it with patches (for now). There hasn't been any movement in the protobuf-net repo it self.

xPaw:
As for the `EGCSystemMsg`, looking at the protos repo it was actually removed from Dota.

Issue #641, opened 2019-02-11
WebAPI does not properly encode special characters


JustArchi:
While digging in my own toys, I've found out that current WebAPI code does not properly encode special characters. This is **[well documented](https://github.com/SteamRE/SteamKit/blob/e2d6f88b2d2840cabb92bc5fe3d72b40432caa41/SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs#L271-L277)** in SK2 code, except it's not universally correct approach. It's totally valid to use characters that can be problematic for the URL, namely `&` as an example. If I use a key-value pair such as `name -> bob&` then raw data should be encoded as `name=bob%26`, while right now it's encoded as `name=bob&`. This isn't just incorrect as it is, it can create issues with other arguments.

Moreover, the rabbit hole is much deeper since changing it to proper encoding would actually **regress** in valid code. I was testing `ISteamUserAuth/AuthenticateUser` API endpoint today and wondering why `FormUrlEncodedContent()` doesn't work nicely with my args, turns out it's because Steam actually expects binary there, and you can't encode those bytes for special characters, as the data will be no longer correct (and this is why my `FormUrlEncodedContent()` didn't work).

Considering the facts above this can be a problematic issue to fix, since you can't know in advance whether something should be encoded or not. The easiest solution that would not regress and fix the underlying issue would be a check for type of each WebAPI argument, and apply new encoding logic for strings, while keeping current no-encoding logic for `byte[]`. The only problem with this is that it's so abstract to do things this way that it might be better to rewrite entire http content into something like `multipart/form-data` and handle strings and binary in different way - I just didn't test how nicely Steam servers play with it, and it'd still require extra logic like the one above what exactly type the user is supplying, so `byte[]` becomes allowed and encoded as binary.

Right now current behaviour of SK2 breaks **[RFC 1866](https://tools.ietf.org/html/rfc1866#section-8.2.1)**:

```
        1. The form field names and values are escaped: space
        characters are replaced by `+', and then reserved characters
        are escaped as per [URL]; that is, non-alphanumeric
        characters are replaced by `%HH', a percent sign and two
        hexadecimal digits representing the ASCII code of the
        character. Line breaks, as in multi-line text field values,
        are represented as CR LF pairs, i.e. `%0D%0A'.
```

And the most hilarious thing is how making it **right** would regress in my own working code, and make it impossible to use for binary data.

Let me know what you think, I'm super confused myself.

yaakov-h:
[...]

PR #668, opened 2019-04-27
Pass WebAPI args as generic objects. Encode with correct HttpUtility.UrlEncode override

azuisleet:
Breaking change. Rather than forcing everything into a string up front, keep them as generic objects. See #641 

These comments need clarification:

```
// TODO: the WebAPI is a special snowflake that needs to appropriately handle url encoding
// this is in contrast to the steam3 content server APIs which use an entirely different scheme of encoding```

### DIFF
--- a/Samples/6.WebAPI/Program.cs
+++ b/Samples/6.WebAPI/Program.cs
@@ -92,7 +92,7 @@ namespace Sample6_WebAPI
             // you can call interface functions through a Call method
             using ( WebAPI.Interface steamNews = WebAPI.GetInterface( "ISteamNews" ) )
             {
-                Dictionary<string, object> newsArgs = new Dictionary<string,object>();
+                Dictionary<string, object> newsArgs = new Dictionary<string, object>();
                 newsArgs[ "appid" ] = "440";
 
                 KeyValue results = steamNews.Call( "GetNewsForApp", /* version */ 1, newsArgs );
--- a/Samples/6.WebAPI/Program.cs
+++ b/Samples/6.WebAPI/Program.cs
@@ -92,10 +92,15 @@ namespace Sample6_WebAPI
             // you can call interface functions through a Call method
             using ( WebAPI.Interface steamNews = WebAPI.GetInterface( "ISteamNews" ) )
             {
-                Dictionary<string, string> newsArgs = new Dictionary<string,string>();
+                Dictionary<string, object> newsArgs = new Dictionary<string,object>();
                 newsArgs[ "appid" ] = "440";
 
                 KeyValue results = steamNews.Call( "GetNewsForApp", /* version */ 1, newsArgs );
+
+                foreach ( KeyValue news in results[ "newsitems" ][ "newsitem" ].Children )
+                {
+                    Console.WriteLine( "News: {0}", news[ "title" ].AsString() );
+                }
             }
         }
     }
--- a/SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs
+++ b/SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs
@@ -272,9 +272,13 @@ namespace SteamKit2
                     string key = HttpUtility.UrlEncode( kvp.Key );
                     string value;
 
-                    if ( kvp.Value is byte[] )
+                    if ( kvp.Value == null )
                     {
-                        value = HttpUtility.UrlEncode( kvp.Value as byte[] );
+                        value = string.Empty;
+                    }
+                    else if ( kvp.Value is byte[] buffer )
+                    {
+                        value = HttpUtility.UrlEncode( buffer );
                     }
                     else
                     {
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by JustArchi** [comment]
I'm just not sure if this change won't break stuff in regards to server-based interpretation.

Previously, I had this code:

```
await iSteamUserAuth.AuthenticateUser(
	encrypted_loginkey: Encoding.ASCII.GetString(WebUtility.UrlEncodeToBytes(encryptedLoginKey, 0, encryptedLoginKey.Length)),
	method: WebRequestMethods.Http.Post,
	sessionkey: Encoding.ASCII.GetString(WebUtility.UrlEncodeToBytes(encryptedSessionKey, 0, encryptedSessionKey.Length)),
	steamid: steamID
)
```

After this change, I expect to use this:

```
await iSteamUserAuth.AuthenticateUser(
	encrypted_loginkey: encryptedLoginKey,
	method: WebRequestMethods.Http.Post,
	sessionkey: encryptedSessionKey,
	steamid: steamID
)
```

I guess I'll just test this change once it's ready, feel free to ping me if you'd like me to.

**Comment by azuisleet** [comment]
Yes, the intention is to pull the encoding logic out of user code and into the library. You should be using the second example. I have not done extensive testing yet.

File: Samples/6.WebAPI/Program.cs [hunk 1]
```diff
@@ -92,10 +92,15 @@ static void Main( string[] args )
             // you can call interface functions through a Call method
             using ( WebAPI.Interface steamNews = WebAPI.GetInterface( "ISteamNews" ) )
             {
-                Dictionary<string, string> newsArgs = new Dictionary<string,string>();
+                Dictionary<string, object> newsArgs = new Dictionary<string,object>();
```
VoiDeD: Nit: whitespace

SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs
```diff
@@ -268,11 +269,17 @@ async Task<KeyValue> CallAsyncCore( HttpMethod method, string func, int version
...
                 paramBuilder.Append( string.Join( "&", args.Select( kvp =>
                 {
-                    // TODO: the WebAPI is a special snowflake that needs to appropriately handle url encoding
-                    // this is in contrast to the steam3 content server APIs which use an entirely different scheme of encoding
+                    string key = HttpUtility.UrlEncode( kvp.Key );
+                    string value;
 
-                    string key = WebHelpers.UrlEncode( kvp.Key );
-                    string value = kvp.Value; // WebHelpers.UrlEncode( kvp.Value );
+                    if ( kvp.Value is byte[] )
```
VoiDeD:
```cs
if ( kvp.Value is byte[] buffer )
{
    value = HttpUtility.UrlEncode( buffer );
}
```

Saves an extra conversion.

SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs
```diff
@@ -268,11 +269,17 @@ async Task<KeyValue> CallAsyncCore( HttpMethod method, string func, int version
...
+                    {
+                        value = HttpUtility.UrlEncode( kvp.Value as byte[] );
+                    }
+                    else
+                    {
+                        value = HttpUtility.UrlEncode( kvp.Value.ToString() );
```
yaakov-h: Will this explode on null?

JustArchi:
> I have not done extensive testing yet.

Something is wrong since it's no longer possible to make ISteamUserAuth/AuthenticateUser work in any form. Steam responds with 400 bad request when supplied with `byte[]` arguments as in my example above. Could you take a look? I couldn't make it work through any means. I have login code in ASF if you'd like full login procedure.

JustArchi:
In any case, just ensure the above is solved before triggering any stable release, since right now it's not just a regression but regression with no means of correction (@yaakov-h).

xPaw:
For what it's worth, the non dynamic method works for me:

https://github.com/SteamDatabase/SteamDatabaseBackend/blob/1a26defd9750006f4dac553a7bec6deef7de160d/Steam/WebAuth.cs#L70-L77

JustArchi:
You gave me good enough reason to ditch dynamic calls and just use non-dynamic ones instead, dynamic ones feel far more problematic than what it's worth 🤔. I'll see if it fixes my issue (and if yes, we'll know where the issue is).

JustArchi:
I can confirm what @xPaw said above (thank you!), the issue doesn't exist when using `CallAsync()` (or `Call()` for synchronous), only when using dynamic interface.

azuisleet:
Check fa128f1f219d70ba6097444b5069efa8e59fdf73

The setup of the dictionary should be the same now. It was enumerating byte[] arrays.

JustArchi:
Thanks for the quick fix! ❤️

yaakov-h:
@JustArchi A new build is rolling out now, should be live on NuGet by the time you read this.

PR #669, opened 2019-04-27
add LICENSE file

rossengeorgiev:
Text taken from https://www.gnu.org/licenses/old-licenses/lgpl-2.1.txt

This will let github automatically detect the license and display it prominently on the repo's summary tab bar.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,458 @@
+                  GNU LESSER GENERAL PUBLIC LICENSE
+                       Version 2.1, February 1999
+
+ Copyright (C) 1991, 1999 Free Software Foundation, Inc.
+ 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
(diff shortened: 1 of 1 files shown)


yaakov-h:
This looks like a duplicate of #515 without addressing concerns raised there.

asherkin:
> This looks like a duplicate of #515 without addressing concerns raised there.

The protobufs are no longer in-repo and protogen appears to be binaries only (and appears to be MIT-licensed, so fine to incorporate with our modifications under LGPL terms, but should certainly have it's copyright declaration added to the folder as it is unclear - upstream too.)

xPaw:
License files in `SteamKit2/SteamKit2` probably need to be cleaned up.

yaakov-h:
Revisiting this. Correct me if I'm wrong or missing anything, but we would need to:

- [ ] Merge this PR to add LICENSE to the root
- [ ] Add readme or LICENCE to protogen folder
- [ ] Remove SteamKit2/SteamKit2/gpl.txt
- [ ] Remove SteamKit2/SteamKit2/lgpl.txt (use LICENSE instead)
- [ ] Keep SteamKit2/SteamKit2/license.txt?

PR #681, opened 2019-05-20
CI: Fix Travis build

yaakov-h:
- Updated from Ubuntu 14.04 LTS to ~~18.04~~ 16.04 LTS.
- Updated .NET Core SDK from 2.1 LTS to 2.2 (while I'm in here anyway). This does not change our target frameworks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-dist: bionic
+dist: xenial
 
 language: csharp
 mono: none
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,6 @@
-dist: trusty
-
 language: csharp
 mono: none
-dotnet: 2.1
+dotnet: 2.2
 
 git:
     depth: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+dist: bionic
+
 language: csharp
 mono: none
 dotnet: 2.2


xPaw:
FYI Travis builds work fine for me without specifying the distro: https://github.com/SteamDatabase/SteamDatabaseBackend/blob/master/.travis.yml

yaakov-h:
And yet it [didn't work for me](https://travis-ci.org/SteamRE/SteamKit/builds/532708475#L1) and I have no idea what's different.

JustArchi:
@yaakov-h during upload/update process of .NET Core packages, it's possible for a short while that there will be no package available to download, and since Travis always installs it from scratch (as opposed to image being bundled with it), if you catch CI during that period then apt will report that there is no package found.

I had that a few times with ASF, re-running the travis build always fixes that for me. I don't see a reason otherwise, since Ubuntu 14.04 (default for travis) is supported by .NET Core, so either travis guys screwed something up, or it should work if you re-try.

JustArchi:
Regardless, I believe it's a good idea to specify the distro version, especially since Travis likes to default to outdated ones, those that are very likely to be out of .NET Core support sooner than Travis bumps the new default to another one, not to mention that this bump can cause other regressions, while referencing distro yourself allows you to appropriately bump it and test it before it's happening without your knowledge.

yaakov-h:
@JustArchi Apt was failing for several days straight, and 14.04 hit end of life three weeks ago.

(I assumed the two events were related, apparently they were not.)

PR #683, opened 2019-05-23
Improve crypto performance

yaakov-h:
### Benchmark:
Benchmark run against a packet about 670 bytes long.

**Current code:**

- Time to encrypt an outgoing packet: 6.464us
- Memory allocations to encrypt an outgoing packet: 6320B
- Time to decrypt an incoming packet: 6.524us
- Memory allocations to decrypt an incoming packet: 5208B

**This branch:**

- Time to encrypt an outgoing packet: 4.054us
- Memory allocations to encrypt an outgoing packet: 848B +size of packet
- Time to decrypt an incoming packet: 3.688us
- Memory allocations to decrypt an incoming packet: 1008B + size of packet

Unfortunately it seems that about 2/3rds of the CPU time is spent computing the HMAC. I don't think we can do anything about that at the moment.

See also: #552 

Some discussion on previous PR at #554.

### DIFF
--- a/SteamKit2/SteamKit2/Networking/Steam3/NetFilterEncryptionWithHMAC.cs
+++ b/SteamKit2/SteamKit2/Networking/Steam3/NetFilterEncryptionWithHMAC.cs
@@ -5,30 +5,37 @@
 
 
 using System;
+using System.Buffers;
 using System.IO;
 using System.Security.Cryptography;
+using SteamKit2.Util;
 
 namespace SteamKit2
 {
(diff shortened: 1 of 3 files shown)


xPaw:
What is this blocked on?

yaakov-h:
Merge conflicts, tests, and ideally I'd rather use `Span<>` than `ArraySegment<>` but that means either taking a package dependency or dropping <= netstandard2.1 I think, and I'm not sure which way we want to go on that overall.

xPaw:
Superceded by #1381

PR #684, opened 2019-05-23
KeyValue: Handle (ignore) conditionals

yaakov-h:
As pointed out by @xPaw, we crash when reading `tf_english.txt`.

I narrowed this down to our handling of conditionals - it's completely broken. We count the conditional as a key, then read the next key as a value, etc. until our keys and values are swapped.

Therefore, an empty **value** after a conditional will trigger the exception that fails a parse on an empty **key**.

I don't think this is necessarily the best way to handle them, but at least it doesn't crash :smile: 

Do not merge (yet), I'm only posting this for discussion. How do we want to handle conditionals, and is this the best way?

I think I can see some implementation flaws in this code (i.e. the green part of the diff) but it's getting late and I need to go to bed so I'm stopping for tonight. I know it's not perfect.

Previous PR and discussion at #232.

### DIFF
--- a/SteamKit2/SteamKit2/Types/KeyValue.cs
+++ b/SteamKit2/SteamKit2/Types/KeyValue.cs
@@ -25,7 +25,7 @@ namespace SteamKit2
             : base( input )
         {
             bool wasQuoted;
-            bool wasConditional;
+            string condition;
 
             KeyValue currentKey = kv;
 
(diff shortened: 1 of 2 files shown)


xPaw:
Worth pointing out that @yaakov-h wrote a different keyvalue parser and it has good test coverage: https://github.com/SteamDatabase/ValveKeyValue

yaakov-h:
I only did that after stumbling across this and needing a better solution. 😄

xPaw:
I've copied tests from this PR into master: 0c6b64388b2599a64178065a7ea84923bbfe9c87

At this point I'm thinking we should just intentionally not support parsing conditionals and tell people to use the VKV lib if they need to support it.

yaakov-h:
Is there anything where SteamKit2 parses the KV and makes decisions from that, or where we parse it without letting the consumer access the raw data to parse it themselves?

xPaw:
We only parse appinfo which has no conditionals

PR #686, opened 2019-05-23
Generate protobufs ourselves using protobuf-net.Reflection

yaakov-h:
Another approach to #628.

This assumes/ignores the `[DefaultValue]` problem. I think if we can get that upstreamed, this will be fine.

Note that:

- I'm just outright ignoring the "hey you didn't specify `proto2` or `proto3`" warning.
- This removes XmlSerializer attributes
- This removes `[Serializable]`
- This removes `EnumPassthru`, though I don't know if that is still needed or not.


yaakov-h:
The DefaultValue bug has been fixed (https://github.com/mgravell/protobuf-net/pull/516) but I have no idea if/when that code will be released.

I guess compiling protobuf-net ourselves (without any custom patches) wouldn't be so bad...

yaakov-h:
With these change, `_` characaters are being dropped in enum names, e.g. `k_Foo` becomes just `kFoo`.

Also, it seems to crash when logging in to Steam. 😅

yaakov-h:
`_` characters are no longer being dropped after updating to latest (alpha) `protobuf-net.Reflection`.

The flip side is that enum values have exactly the same name as in the protobuf, so many are now `SCREAMING_SNAKE_CASE` as per the original protos.

This PR now relies on https://github.com/mgravell/protobuf-net/pull/518.

Resources/ProtobufGen/gc.proto
```diff
@@ -1,21 +1,21 @@
-enum GCProtoBufMsgSrc {
```
xPaw: This file changed for no reason (line endings I assume?)

xPaw:
This looks like it achieves the same result as my patch approach, except by extending protobuf.net, so that's good.

The only missing change would be forced default to proto2 version, instead of leaving it null, I am unsure if this actually affects any of the generated output.

Resources/ProtobufGen/gc.proto
yaakov-h: I only moved the files... possibly something in my git install tweaked the line endings?

Resources/ProtobufGen/ProtobufGen/Program.cs
```diff
@@ -0,0 +1,108 @@
...
+            set.Process();
+
+            var errors = set.GetErrors();
+            if ( errors.Length > 0 )
+            {
+                var errorsCount = 0;
+
+                foreach ( var error in errors )
+                {
+                    Console.Out.WriteLine( $"{error.File} ({error.LineNumber}, {error.ColumnNumber}): {error.Message}" );
```
xPaw:
```suggestion
                    Console.Error.WriteLine( $"{error.File} ({error.LineNumber}, {error.ColumnNumber}): {error.Message}" );
```

yaakov-h:
Tested and verified that `EnumPassthru` is no longer required.

Resources/ProtobufGen/generate-all.ps1
```diff
@@ -1,59 +1,48 @@
...
+.PARAMETER ProtoDir
+    Protobuf folders to process. (Default: all)
+.EXAMPLE
+    PS C:\> .\generate-all.ps1 -ProtoDir steam,tf2
+#>
+param([string[]]$ProtoDir)
+
+$ProtoGenDir = $PSScriptRoot
+$ProtoGenDll = Join-Path $PSScriptRoot 'ProtobufGen\bin\Debug\netcoreapp2.1\ProtobufGen.dll'
+$ProtoBase = Join-Path $PSScriptRoot '..\ProtoBufs'
```
psychonic: Are we still planning on shipping the precompiled binary as before? Or should a step be added to build this first?

Resources/ProtobufGen/generate-all.ps1
```diff
@@ -1,59 +1,48 @@
...
+$SK2Base = Join-Path $PSScriptRoot '..\..\SteamKit2\SteamKit2\Base\Generated'
+
+Push-Location
+
+$protos = Import-Csv -LiteralPath (Join-Path $PSScriptRoot 'protos.csv') # |
```
psychonic: This looks like maybe a temporary change for testing that as accidentally left in.

Resources/ProtobufGen/generate-all.ps1
yaakov-h: No point shipping it when we have the code, but I also don't want to rebuild the generator every time.

We could leave it up to the user... or I'll see how smart the incremental build chain is.

yaakov-h: Incremental build system seems good enough.

Issue #656, opened 2019-03-30
Lobby filters


Benjamin-Dobell:
`CMsgClientMMSGetLobbyList.Filter` takes a `filter_type` integer for which there is (to my knowledge) no corresponding enum or documentation.

Based on the Steamworks docs there ought to be a way to define:

* string filter
* numerical filter
* near value filter
* filter slots available
* distance filter
* result count filter

I suppose it's possible there's a unique `filter_type` for each of the above, but even if there was it's still not quite clear if/when `key` ought to be specified; in addition to `value`, which is presumably always included.

It'd be great to have a sample, assuming someone has the necessary knowledge.

Issue #487, opened 2017-11-10
Get remote IP for current connection


leonard-thieu:
I'd like to report the remote IP for telemetry purposes. I'm connecting through TCP. Is it possible to get the remote IP for the current connection? From looking through the code, it doesn't appear to be publicly exposed.

yaakov-h:
Not currently.

In theory we could expose this for TCP and UDP connections, but for WebSocket we actually don't know what IP we're connected to - name resolution is handled internally through WinHTTP on Windows, and is similarly done through the CoreFX implementations on .NET Core Linux/macOS.

leonard-thieu:
Okay, so if I wanted to expose this in my own fork, I just need to expose [`CurrentEndPoint`](https://github.com/SteamRE/SteamKit/blob/9677726a829df25b4ebb9febb8e8048fc5cec8dc/SteamKit2/SteamKit2/Networking/Steam3/IConnection.cs#L25) on https://github.com/SteamRE/SteamKit/blob/9677726a829df25b4ebb9febb8e8048fc5cec8dc/SteamKit2/SteamKit2/Steam/CMClient.cs#L99
and expose https://github.com/SteamRE/SteamKit/blob/9677726a829df25b4ebb9febb8e8048fc5cec8dc/SteamKit2/SteamKit2/Networking/Steam3/TcpConnection.cs#L19
through [`CurrentEndPoint`](https://github.com/SteamRE/SteamKit/blob/9677726a829df25b4ebb9febb8e8048fc5cec8dc/SteamKit2/SteamKit2/Networking/Steam3/TcpConnection.cs#L43).

Does that sound about right?

yaakov-h:
Sounds about right. 

The second part *should* have already been done here...

xPaw:
This issue was closed incorrectly, #704 added `PublicIP` which is not the same thing as the remote ip of the server you're connected to.

Issue #131, opened 2015-06-29
MatchMaking interface?


Jvs34:
Hi, I was wondering if there's any way to implement the ISteamMatchMaking interface within steamkit, I've looked at the code and there's no protobufs of it at all.

Is it something overlooked or does it simply not use protobufs ( or am I blind)?

Otherwise I guess I would have to do it myself with nethook2.

Netshroud:
Have a look at the protobufs starting with `CMsgClientMMS`.

Jvs34:
Alright I was just blind then, thanks.

Jvs34:
Finally decided to give this a try, although for now I'm just working on the lobbies side of things, then I'll head onto the server list stuff and just work down other stuff.

Question: should I keep the API kind of similar to what steam exposes with steamworks?
Things like ISteamMatchmaking::GetLobbyByIndex are kind of lame to do in c# since the proto itself returns the lobbies straight up, and it'd require pretty much caching all the lobby data received from the messages just like SteamFriends does.

Considering the caching too, the lobbies would probably be readonly for the user, and then updated when steam sends the lobby data updates.

Also, should I provide async methods with the AsyncJob way or stick to callback based approach? What's the consensus on that?

One last thing, should I offer overloads that let you pass the Lobby class itself instead of just the steamid?

ISteamMatchmaking
- [X] CreateLobby( appid, privacy , maxmembers )
- [ ] JoinLobby( appid , lobbysteamid )
- [X] RequestLobbyList( appid , filterclass ) (you can call chaining methods like so ) 
```csharp
            steamMatchMaking.RequestLobbyList( 480 , new SteamMatchMaking.LobbyFilter()
                .AddDistanceFilter( ELobbyDistanceFilter.Worldwide )
                .AddNearValueFilter( "LEVEL" , 5 )
                .AddStringFilter( "MAP" , "whatever" , ELobbyComparison.Equal )
            );       
```
- [ ] InviteUserToLobby( lobbysteamid , usersteamid ) 
- [ ] GetNumLobbyMembers
- [ ] Etc etc

Nothing done on these atm

- ISteamMatchmakingServers
- ISteamGameSearch
- ISteamParties

Jvs34:
The original scope was just matchmaking, will open a new issue for the other interfaces in the future.

PR #704, opened 2019-07-15
SteamMatchmaking - creating, retrieving, listing, joining and leaving lobbies.

Benjamin-Dobell:
An implementation of SteamMatchmaking.

### Currently supported

- [x] `CreateLobby` (Sending of `CMsgClientMMSCreateLobby`)
- [x] `CreateLobbyCallback` (Receiving of `CMsgClientMMSCreateLobbyResponse`)
- [x] `SetLobbyData` (Sending of `CMsgClientMMSSetLobbyData`)
- [x] `SetLobbyDataCallback` (Receiving of `CMsgClientMMSSetLobbyDataResponse`)
- [x] `SetLobbyOwner` (Sending of `CMsgClientMMSSetLobbyOwner`)
- [x] `SetLobbyOwnerCallback` (Receiving of `CMsgClientMMSSetLobbyOwnerResponse`)
- [x] `GetLobbyList` (Sending of `CMsgClientMMSGetLobbyList`) i.e. Searching for lobbies.
- [x] `LobbyListCallback` (Receiving of `CMsgClientMMSGetLobbyListResponse`)
   - [x] `DistanceFilter`
   - [x] `NearValueFilter`
   - [x] `NumericalFilter`
   - [x] `SlotsAvailableFilter`
   - [x] `StringFilter`
- [x] `GetLobby` (Equivalent to functionality offered by https://partner.steamgames.com/doc/api/ISteamMatchmaking i.e. non-blocking synchronous access to lobbies we already "know about" .
- [x] `LobbyDataCallback` (Receiving of `CMsgClientMMSLobbyData`) i.e. Lobby change tracking.
- [x] `UserJoinedLobbyCallback` (Receiving of `CMsgClientMMSUserJoinedLobby`)
- [x] `UserLeftLobbyCallback` (Receiving of `CMsgClientMMSUserLeftLobby`).

Since originally opening the PR, the following has been added:

- [x] Joining lobbies.
- [x] Leaving lobbies.
- [x] Inviting another user to a lobby.
- [x] Getting data for a specific lobby rather than from a lobby list (probably only useful if you were invited to a lobby).

### What's not implemented in this PR

- [ ] Assigning a game server to a lobby.
- [ ] Lobby chat.
- [ ] Steam's "frenemy" stuff, which is officially deprecated and according to ISteamNetworking documentation doesn't actually do anything 🤷‍♂

### Related Issues

Closes https://github.com/SteamRE/SteamKit/issues/656

Closes https://github.com/SteamRE/SteamKit/issues/487 ... as the requested functionality is necessary to implement Matchmaking. Specifically, when we create a lobby we need to submit our public IP, so we require persistent access to it. Same is true of `IPCountryCode`.

Ping https://github.com/SteamRE/SteamKit/issues/131 (only partially implemented, so not closing).

### DIFF
--- a/Resources/SteamLanguage/enums.steamd
+++ b/Resources/SteamLanguage/enums.steamd
@@ -1473,3 +1473,39 @@ enum ETradeOfferConfirmationMethod
 	Email = 1;
 	MobileApp = 2;
 };
+
+enum ELobbyType
+{
+    Private = 0;
+    FriendsOnly = 1;
+    Public = 2;
+    Invisible = 3;
+    PrivateUnique = 4;
+};
+
+enum ELobbyFilterType
+{
+    String = 0;
+    Numerical = 1;
+    SlotsAvailable = 2;
+    NearValue = 3;
+    Distance = 4;
+};
+
+enum ELobbyComparison
+{
+    EqualToOrLessThan = -2;
+    LessThan = -1;
+    Equal = 0;
+    GreaterThan = 1;
+    EqualToOrGreaterThan = 2;
+    NotEqual = 3;
+};
+
+public enum ELobbyDistanceFilter
+{
+    Close = 0;
+    Default = 1;
+    Far = 2;
+    Worldwide = 3;
+};
--- a/SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/Callbacks.cs
+++ b/SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/Callbacks.cs
@@ -1,0 +1,260 @@
+using System.Collections.Generic;
+
+namespace SteamKit2
+{
+    public partial class SteamMatchmaking
+    {
+        /// <summary>
+        /// This callback is fired in response to <see cref="GetLobbyList"/>.
+        /// </summary>
+        public sealed class GetLobbyListCallback : CallbackMsg
+        {
+            /// <summary>
+            /// ID of the app the lobbies belongs to.
+            /// </summary>
+            public uint AppID { get; }
+
+            /// <summary>
+            /// The result of the request.
+            /// </summary>
+            public EResult Result { get; }
+
+            /// <summary>
+            /// The list of lobbies matching the criteria specified with <see cref="GetLobbyList"/>.
+            /// </summary>
+            public List<Lobby> Lobbies { get; }
+
+            internal GetLobbyListCallback( uint appId, EResult res, List<Lobby> lobbies )
+            {
+                AppID = appId;
+                Result = res;
+                Lobbies = lobbies;
+            }
+        }
+
+        /// <summary>
+        /// This callback is fired in response to <see cref="CreateLobby"/>.
+        /// </summary>
+        public sealed class CreateLobbyCallback : CallbackMsg
+        {
+            /// <summary>
+            /// ID of the app the created lobby belongs to.
+            /// </summary>
+            public uint AppID { get; }
+
+            /// <summary>
+            /// The result of the request.
+            /// </summary>
+            public EResult Result { get; }
+
+            /// <summary>
+            /// The SteamID of the created lobby.
+            /// </summary>
+            public SteamID LobbySteamID { get; }
+
+            internal CreateLobbyCallback( JobID jobId, uint appId, EResult res, SteamID lobbySteamId )
+            {
+                JobID = jobId;
+                AppID = appId;
+                Result = res;
+                LobbySteamID = lobbySteamId;
+            }
+        }
+
+        /// <summary>
+        /// This callback is fired in response to <see cref="SetLobbyData"/>.
+        /// </summary>
+        public sealed class SetLobbyDataCallback : CallbackMsg
+        {
+            /// <summary>
+            /// ID of the app the targeted lobby belongs to.
+            /// </summary>
+            public uint AppID { get; }
+
+            /// <summary>
+            /// The result of the request.
+            /// </summary>
+            public EResult Result { get; }
+
+            /// <summary>
+            /// The SteamID of the targeted Lobby.
+            /// </summary>
+            public SteamID LobbySteamID { get; }
+
+            internal SetLobbyDataCallback( JobID jobId, uint appId, EResult res, SteamID lobbySteamId )
+            {
+                JobID = jobId;
+                AppID = appId;
+                Result = res;
+                LobbySteamID = lobbySteamId;
+            }
+        }
+
+        /// <summary>
+        /// This callback is fired in response to <see cref="SetLobbyOwner"/>.
+        /// </summary>
+        public sealed class SetLobbyOwnerCallback : CallbackMsg
+        {
+            /// <summary>
+            /// ID of the app the targeted lobby belongs to.
+            /// </summary>
+            public uint AppID { get; }
+
+            /// <summary>
+            /// The result of the request.
+            /// </summary>
+            public EResult Result { get; }
+
+            /// <summary>
+            /// The SteamID of the targeted Lobby.
+            /// </summary>
+            public SteamID LobbySteamID { get; }
+
+            internal SetLobbyOwnerCallback( JobID jobId, uint appId, EResult res, SteamID lobbySteamId )
+            {
+                JobID = jobId;
+                AppID = appId;
+                Result = res;
+                LobbySteamID = lobbySteamId;
+            }
+        }
+
+        /// <summary>
+        /// This callback is fired in response to <see cref="JoinLobby"/>.
+        /// </summary>
+        public sealed class JoinLobbyCallback : CallbackMsg
+        {
+            /// <summary>
+            /// ID of the app the targeted lobby belongs to.
+            /// </summary>
+            public uint AppID { get; }
+
+            /// <summary>
+            /// The result of the request.
+            /// </summary>
+            public EChatRoomEnterResponse ChatRoomEnterResponse { get; }
+
+            /// <summary>
+            /// The joined <see cref="Lobby"/>, when <see cref="ChatRoomEnterResponse"/> equals
+            /// <see cref="EChatRoomEnterResponse.Success"/>, otherwise <c>null</c>
+            /// </summary>
+            public Lobby Lobby { get; }
+
+            internal JoinLobbyCallback( JobID jobId, uint appId, EChatRoomEnterResponse res, Lobby lobby )
+            {
+                JobID = jobId;
+                AppID = appId;
+                ChatRoomEnterResponse = res;
+                Lobby = lobby;
+            }
+        }
+
+        /// <summary>
+        /// This callback is fired in response to <see cref="LeaveLobby"/>.
+        /// </summary>
+        public sealed class LeaveLobbyCallback : CallbackMsg
+        {
+            /// <summary>
+            /// ID of the app the targeted lobby belongs to.
+            /// </summary>
+            public uint AppID { get; }
+
+            /// <summary>
+            /// The result of the request.
+            /// </summary>
+            public EResult Result { get; }
+
+            /// <summary>
+            /// The SteamID of the targeted Lobby.
+            /// </summary>
+            public SteamID LobbySteamID { get; }
+
+            internal LeaveLobbyCallback( JobID jobId, uint appId, EResult res, SteamID lobbySteamId )
+            {
+                JobID = jobId;
+                AppID = appId;
+                Result = res;
+                LobbySteamID = lobbySteamId;
+            }
+        }
+
+        /// <summary>
+        /// This callback is fired in response to <see cref="GetLobbyData"/>, as well as whenever Steam sends us updated lobby data.
+        /// </summary>
+        public sealed class LobbyDataCallback : CallbackMsg
+        {
+            /// <summary>
+            /// ID of the app the updated lobby belongs to.
+            /// </summary>
+            public uint AppID { get; }
+
+            /// <summary>
+            /// The lobby that was updated.
+            /// </summary>
+            public Lobby Lobby { get; }
+
+            internal LobbyDataCallback( JobID jobId, uint appId, Lobby lobby )
+            {
+                JobID = jobId;
+                AppID = appId;
+                Lobby = lobby;
+            }
+        }
+
+        /// <summary>
+        /// This callback is fired whenever Steam informs us a user has joined a lobby.
+        /// </summary>
+        public sealed class UserJoinedLobbyCallback : CallbackMsg
+        {
+            /// <summary>
+            /// ID of the app the lobby belongs to.
+            /// </summary>
+            public uint AppID { get; }
+
+            /// <summary>
+            /// The SteamID of the lobby that a member joined.
+            /// </summary>
+            public SteamID LobbySteamID { get; }
+
+            /// <summary>
+            /// The lobby member that joined.
+            /// </summary>
+            public Lobby.Member User { get; }
+
+            internal UserJoinedLobbyCallback( uint appId, SteamID lobbySteamId, Lobby.Member user )
+            {
+                AppID = appId;
+                LobbySteamID = lobbySteamId;
+                User = user;
+            }
+        }
+
+        /// <summary>
+        /// This callback is fired whenever Steam informs us a user has left a lobby.
+        /// </summary>
+        public sealed class UserLeftLobbyCallback : CallbackMsg
+        {
+            /// <summary>
+            /// ID of the app the lobby belongs to.
+            /// </summary>
+            public uint AppID { get; }
+
+            /// <summary>
+            /// The SteamID of the lobby that a member left.
+            /// </summary>
+            public SteamID LobbySteamID { get; }
+
+            /// <summary>
+            /// The lobby member that left.
+            /// </summary>
+            public Lobby.Member User { get; }
+
+            internal UserLeftLobbyCallback( uint appId, SteamID lobbySteamId, Lobby.Member user )
+            {
+                AppID = appId;
+                LobbySteamID = lobbySteamId;
+                User = user;
+            }
+        }
+    }
+}
--- a/SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/Lobby.cs
+++ b/SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/Lobby.cs
@@ -1,0 +1,399 @@
+using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.IO;
+using System.Linq;
+using SteamKit2.Internal;
+
+namespace SteamKit2
+{
+    public partial class SteamMatchmaking
+    {
+        /// <summary>
+        /// Represents a Steam lobby.
+        /// </summary>
+        public sealed class Lobby
+        {
+            /// <summary>
+            /// The lobby filter base class.
+            /// </summary>
+            public abstract class Filter
+            {
+                /// <summary>
+                /// The type of filter.
+                /// </summary>
+                public ELobbyFilterType FilterType { get; }
+
+                /// <summary>
+                /// The metadata key this filter pertains to. Under certain circumstances e.g. a distance
+                /// filter, this will be an empty string.
+                /// </summary>
+                public string Key { get; }
+
+                /// <summary>
+                /// The comparison method used by this filter.
+                /// </summary>
+                public ELobbyComparison Comparison { get; }
+
+                /// <summary>
+                /// Base constructor for all filter sub-classes.
+                /// </summary>
+                /// <param name="filterType">The type of filter.</param>
+                /// <param name="key">The metadata key this filter pertains to.</param>
+                /// <param name="comparison">The comparison method used by this filter.</param>
+                protected Filter( ELobbyFilterType filterType, string key, ELobbyComparison comparison )
+                {
+                    FilterType = filterType;
+                    Key = key;
+                    Comparison = comparison;
+                }
+
+                /// <summary>
+                /// Serializes the filter into a representation used internally by SteamMatchmaking.
+                /// </summary>
+                /// <returns>A protobuf serializable representation of this filter.</returns>
+                public virtual CMsgClientMMSGetLobbyList.Filter Serialize()
+                {
+                    var filter = new CMsgClientMMSGetLobbyList.Filter();
+                    filter.filter_type = ( int )FilterType;
+                    filter.key = Key;
+                    filter.comparision = ( int )Comparison;
+                    return filter;
+                }
+            }
+
+            /// <summary>
+            /// Can be used to filter lobbies geographically (based on IP according to Steam's IP database).
+            /// </summary>
+            public sealed class DistanceFilter : Filter
+            {
+                /// <summary>
+                /// Steam distance filter value.
+                /// </summary>
+                public ELobbyDistanceFilter Value { get; }
+
+                /// <summary>
+                /// Initializes a new instance of the <see cref="DistanceFilter"/> class.
+                /// </summary>
+                /// <param name="value">Steam distance filter value.</param>
+                public DistanceFilter( ELobbyDistanceFilter value ) : base( ELobbyFilterType.Distance, "", ELobbyComparison.Equal )
+                {
+                    Value = value;
+                }
+
+                /// <summary>
+                /// Serializes the distance filter into a representation used internally by SteamMatchmaking.
+                /// </summary>
+                /// <returns>A protobuf serializable representation of this filter.</returns>
+                public override CMsgClientMMSGetLobbyList.Filter Serialize()
+                {
+                    var filter = base.Serialize();
+                    filter.value = ( ( int )Value ).ToString();
+                    return filter;
+                }
+            }
+
+            /// <summary>
+            /// Can be used to filter lobbies with a metadata value closest to the specified value. Multiple
+            /// near filters can be specified, with former filters taking precedence over latter filters.
+            /// </summary>
+            public sealed class NearValueFilter : Filter
+            {
+                /// <summary>
+                /// Integer value that lobbies' metadata value should be close to.
+                /// </summary>
+                public int Value { get; }
+
+                /// <summary>
+                /// Initializes a new instance of the <see cref="NearValueFilter"/> class.
+                /// </summary>
+                /// <param name="key">The metadata key this filter pertains to.</param>
+                /// <param name="value">Integer value to compare against.</param>
+                public NearValueFilter( string key, int value ) : base( ELobbyFilterType.NearValue, key, ELobbyComparison.Equal )
+                {
+                    Value = value;
+                }
+
+                /// <summary>
+                /// Serializes the slots available filter into a representation used internally by SteamMatchmaking.
+                /// </summary>
+                /// <returns>A protobuf serializable representation of this filter.</returns>
+                public override CMsgClientMMSGetLobbyList.Filter Serialize()
+                {
+                    var filter = base.Serialize();
+                    filter.value = Value.ToString();
+                    return filter;
+                }
+            }
+
+            /// <summary>
+            /// Can be used to filter lobbies by comparing an integer against a value in each lobby's metadata.
+            /// </summary>
+            public sealed class NumericalFilter : Filter
+            {
+                /// <summary>
+                /// Integer value to compare against.
+                /// </summary>
+                public int Value { get; }
+
+                /// <summary>
+                /// Initializes a new instance of the <see cref="NumericalFilter"/> class.
+                /// </summary>
+                /// <param name="key">The metadata key this filter pertains to.</param>
+                /// <param name="comparison">The comparison method used by this filter.</param>
+                /// <param name="value">Integer value to compare against.</param>
+                public NumericalFilter( string key, ELobbyComparison comparison, int value ) : base( ELobbyFilterType.Numerical, key, comparison )
+                {
+                    Value = value;
+                }
+
+                /// <summary>
+                /// Serializes the numerical filter into a representation used internally by SteamMatchmaking.
+                /// </summary>
+                /// <returns>A protobuf serializable representation of this filter.</returns>
+                public override CMsgClientMMSGetLobbyList.Filter Serialize()
+                {
+                    var filter = base.Serialize();
+                    filter.value = Value.ToString();
+                    return filter;
+                }
+            }
+
+            /// <summary>
+            /// Can be used to filter lobbies by minimum number of slots available.
+            /// </summary>
+            public sealed class SlotsAvailableFilter : Filter
+            {
+                /// <summary>
+                /// Minumum number of slots available in the lobby.
+                /// </summary>
+                public int SlotsAvailable { get; }
+
+                /// <summary>
+                /// Initializes a new instance of the <see cref="SlotsAvailableFilter"/> class.
+                /// </summary>
+                /// <param name="key">The metadata key this filter pertains to.</param>
+                /// <param name="slotsAvailable">Integer value to compare against.</param>
+                public SlotsAvailableFilter( int slotsAvailable ) : base( ELobbyFilterType.SlotsAvailable, "", ELobbyComparison.Equal )
+                {
+                    SlotsAvailable = slotsAvailable;
+                }
+
+                /// <summary>
+                /// Serializes the slots available filter into a representation used internally by SteamMatchmaking.
+                /// </summary>
+                /// <returns>A protobuf serializable representation of this filter.</returns>
+                public override CMsgClientMMSGetLobbyList.Filter Serialize()
+                {
+                    var filter = base.Serialize();
+                    filter.value = SlotsAvailable.ToString();
+                    return filter;
+                }
+            }
+
+            /// <summary>
+            /// Can be used to filter lobbies by comparing a string against a value in each lobby's metadata.
+            /// </summary>
+            public sealed class StringFilter : Filter
+            {
+                /// <summary>
+                /// String value to compare against.
+                /// </summary>
+                public string Value { get; }
+
+                /// <summary>
+                /// Initializes a new instance of the <see cref="StringFilter"/> class.
+                /// </summary>
+                /// <param name="key">The metadata key this filter pertains to.</param>
+                /// <param name="comparison">The comparison method used by this filter.</param>
+                /// <param name="value">String value to compare against.</param>
+                public StringFilter( string key, ELobbyComparison comparison, string value ) : base( ELobbyFilterType.String, key, comparison )
+                {
+                    Value = value;
+                }
+
+                /// <summary>
+                /// Serializes the string filter into a representation used internally by SteamMatchmaking.
+                /// </summary>
+                /// <returns>A protobuf serializable representation of this filter.</returns>
+                public override CMsgClientMMSGetLobbyList.Filter Serialize()
+                {
+                    var filter = base.Serialize();
+                    filter.value = Value;
+                    return filter;
+                }
+            }
+
+            /// <summary>
+            /// Represents a Steam user within a lobby.
+            /// </summary>
+            public sealed class Member
+            {
+                /// <summary>
+                /// SteamID of the lobby member.
+                /// </summary>
+                public SteamID SteamID { get; }
+
+                /// <summary>
+                /// Steam persona of the lobby member.
+                /// </summary>
+                public string PersonaName { get; }
+
+                /// <summary>
+                /// Metadata attached to the lobby member.
+                /// </summary>
+                public IReadOnlyDictionary<string, string> Metadata { get; }
+
+                internal Member( SteamID steamId, string personaName, IReadOnlyDictionary<string, string> metadata = null )
+                {
+                    SteamID = steamId;
+                    PersonaName = personaName;
+                    Metadata = metadata ?? EmptyMetadata;
+                }
+
+                /// <summary>
+                /// Checks to see if this lobby member is equal to another. Only the SteamID of the lobby member is taken into account.
+                /// </summary>
+                /// <param name="obj"></param>
+                /// <returns>true, if obj is <see cref="Member"/> with a matching SteamID. Otherwise, false.</returns>
+                public override bool Equals( object obj )
+                {
+                    if ( obj is Member member )
+                    {
+                        return SteamID.Equals( member.SteamID );
+                    }
+
+                    return false;
+                }
+
+                /// <summary>
+                /// Hash code of the lobby member. Only the SteamID of the lobby member is taken into account.
+                /// </summary>
+                /// <returns>The hash code of this lobby member.</returns>
+                public override int GetHashCode()
+                {
+                    return SteamID.GetHashCode();
+                }
+            }
+
+            /// <summary>
+            /// SteamID of the lobby.
+            /// </summary>
+            public SteamID SteamID { get; }
+
+            /// <summary>
+            /// The type of the lobby.
+            /// </summary>
+            public ELobbyType LobbyType { get; }
+
+            /// <summary>
+            /// The lobby's flags.
+            /// </summary>
+            public int LobbyFlags { get; }
+
+            /// <summary>
+            /// The SteamID of the lobby's owner. Please keep in mind that Steam does not provide lobby
+            /// owner details for lobbies returned in a lobby list. As such, lobbies that have been
+            /// obtained/updated as a result of calling <see cref="SteamMatchmaking.GetLobbyList"/>
+            /// may have a null (or non-null but state) owner.
+            /// </summary>
+            public SteamID OwnerSteamID { get; }
+
+            /// <summary>
+            /// The metadata of the lobby; string key-value pairs.
+            /// </summary>
+            public IReadOnlyDictionary<string, string> Metadata { get; }
+
+            /// <summary>
+            /// The maximum number of members that can occupy the lobby.
+            /// </summary>
+            public int MaxMembers { get; }
+
+            /// <summary>
+            /// The number of members that are currently occupying the lobby.
+            /// </summary>
+            public int NumMembers { get; }
+
+            /// <summary>
+            /// A list of lobby members. This will only be populated for the user's current lobby.
+            /// </summary>
+            public IReadOnlyList<Member> Members { get; }
+
+            /// <summary>
+            /// The distance of the lobby.
+            /// </summary>
+            public float? Distance { get; }
+
+            /// <summary>
+            /// The weight of the lobby.
+            /// </summary>
+            public long? Weight { get; }
+
+            static readonly ReadOnlyDictionary<string, string> EmptyMetadata =
+                new ReadOnlyDictionary<string, string>( new Dictionary<string, string>() );
+
+            static readonly IReadOnlyList<Member> EmptyMembers = Array.AsReadOnly(Array.Empty<Member>());
+
+            internal Lobby( SteamID steamId, ELobbyType lobbyType, int lobbyFlags, SteamID ownerSteamId, IReadOnlyDictionary<string, string> metadata,
+                int maxMembers, int numMembers, IReadOnlyList<Member> members, float? distance, long? weight )
+            {
+                SteamID = steamId;
+                LobbyType = lobbyType;
+                LobbyFlags = lobbyFlags;
+                OwnerSteamID = ownerSteamId;
+                Metadata = metadata ?? EmptyMetadata;
+                MaxMembers = maxMembers;
+                NumMembers = numMembers;
+                Members = members ?? EmptyMembers;
+                Distance = distance;
+                Weight = weight;
+            }
+
+            internal static byte[] EncodeMetadata( IReadOnlyDictionary<string, string> metadata )
+            {
+                var keyValue = new KeyValue( "" );
+
+                if ( metadata != null )
+                {
+                    foreach ( var entry in metadata )
+                    {
+                        keyValue[ entry.Key ] = new KeyValue( null, entry.Value );
+                    }
+                }
+
+                using ( var ms = new MemoryStream() )
+                {
+                    keyValue.SaveToStream( ms, true );
+                    return ms.ToArray();
+                }
+            }
+
+            internal static ReadOnlyDictionary<string, string> DecodeMetadata( byte[] buffer )
+            {
+                if ( buffer.Length == 0 )
+                {
+                    return EmptyMetadata;
+                }
+
+                var keyValue = new KeyValue();
+
+                using ( var ms = new MemoryStream( buffer ) )
+                {
+                    if ( !keyValue.TryReadAsBinary( ms ) )
+                    {
+                        throw new FormatException( "Lobby metadata is of an unexpected format" );
+                    }
+                }
+
+                var metadata = new Dictionary<string, string>();
+
+                foreach ( var value in keyValue.Children )
+                {
+                    metadata[ value.Name ] = value.Value;
+                }
+
+                return new ReadOnlyDictionary<string, string>( metadata );
+            }
+        }
+    }
+}
--- a/SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/SteamMatchmaking.cs
+++ b/SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/SteamMatchmaking.cs
@@ -1,0 +1,667 @@
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using SteamKit2.Internal;
+
+namespace SteamKit2
+{
+    /// <summary>
+    /// This handler is used for creating, joining and obtaining lobby information.
+    /// </summary>
+    public partial class SteamMatchmaking : ClientMsgHandler
+    {
+        readonly Dictionary<EMsg, Action<IPacketMsg>> dispatchMap;
+
+        readonly ConcurrentDictionary<JobID, ProtoBuf.IExtensible> lobbyManipulationRequests = new ConcurrentDictionary<JobID, ProtoBuf.IExtensible>();
+
+        readonly LobbyCache lobbyCache = new LobbyCache();
+
+        internal SteamMatchmaking()
+        {
+            dispatchMap = new Dictionary<EMsg, Action<IPacketMsg>>
+            {
+                { EMsg.ClientMMSCreateLobbyResponse, HandleCreateLobbyResponse },
+                { EMsg.ClientMMSSetLobbyDataResponse, HandleSetLobbyDataResponse },
+                { EMsg.ClientMMSSetLobbyOwnerResponse, HandleSetLobbyOwnerResponse },
+                { EMsg.ClientMMSLobbyData, HandleLobbyData },
+                { EMsg.ClientMMSGetLobbyListResponse, HandleGetLobbyListResponse },
+                { EMsg.ClientMMSJoinLobbyResponse, HandleJoinLobbyResponse },
+                { EMsg.ClientMMSLeaveLobbyResponse, HandleLeaveLobbyResponse },
+                { EMsg.ClientMMSUserJoinedLobby, HandleUserJoinedLobby },
+                { EMsg.ClientMMSUserLeftLobby, HandleUserLeftLobby },
+            };
+        }
+
+        /// <summary>
+        /// Sends a request to create a new lobby.
+        /// </summary>
+        /// <param name="appId">ID of the app the lobby will belong to.</param>
+        /// <param name="lobbyType">The new lobby type.</param>
+        /// <param name="maxMembers">The new maximum number of members that may occupy the lobby.</param>
+        /// <param name="lobbyFlags">The new lobby flags. Defaults to 0.</param>
+        /// <param name="metadata">The new metadata for the lobby. Defaults to <c>null</c> (treated as an empty dictionary).</param>
+        /// <returns><c>null</c>, if the request could not be submitted i.e. not yet logged in. Otherwise, an <see cref="AsyncJob{CreateLobbyCallback}"/>.</returns>
+        public AsyncJob<CreateLobbyCallback> CreateLobby( uint appId, ELobbyType lobbyType, int maxMembers, int lobbyFlags = 0,
+            IReadOnlyDictionary<string, string> metadata = null )
+        {
+            if ( Client.CellID == null )
+            {
+                return null;
+            }
+
+            var personaName = Client.GetHandler<SteamFriends>().GetPersonaName();
+
+            var createLobby = new ClientMsgProtobuf<CMsgClientMMSCreateLobby>( EMsg.ClientMMSCreateLobby )
+            {
+                Body =
+                {
+                    app_id = appId,
+                    lobby_type = ( int )lobbyType,
+                    max_members = maxMembers,
+                    lobby_flags = lobbyFlags,
+                    metadata = Lobby.EncodeMetadata( metadata ),
+                    cell_id = Client.CellID.Value,
+                    public_ip = NetHelpers.GetIPAddress( Client.PublicIP ),
+                    persona_name_owner = personaName
+                },
+                SourceJobID = Client.GetNextJobID()
+            };
+
+            Send( createLobby, appId );
+
+            lobbyManipulationRequests[ createLobby.SourceJobID ] = createLobby.Body;
+            return AttachIncompleteManipulationHandler( new AsyncJob<CreateLobbyCallback>( Client, createLobby.SourceJobID ) );
+        }
+
+        /// <summary>
+        /// Sends a request to update a lobby.
+        /// </summary>
+        /// <param name="appId">ID of app the lobby belongs to.</param>
+        /// <param name="lobbySteamId">The SteamID of the lobby that should be updated.</param>
+        /// <param name="lobbyType">The new lobby type.</param>
+        /// <param name="maxMembers">The new maximum number of members that may occupy the lobby.</param>
+        /// <param name="lobbyFlags">The new lobby flags. Defaults to 0.</param>
+        /// <param name="metadata">The new metadata for the lobby. Defaults to <c>null</c> (treated as an empty dictionary).</param>
+        /// <returns>An <see cref="AsyncJob{SetLobbyDataCallback}"/>.</returns>
+        public AsyncJob<SetLobbyDataCallback> SetLobbyData( uint appId, SteamID lobbySteamId, ELobbyType lobbyType, int maxMembers, int lobbyFlags = 0,
+            IReadOnlyDictionary<string, string> metadata = null )
+        {
+            var setLobbyData = new ClientMsgProtobuf<CMsgClientMMSSetLobbyData>( EMsg.ClientMMSSetLobbyData )
+            {
+                Body =
+                {
+                    app_id = appId,
+                    steam_id_lobby = lobbySteamId,
+                    steam_id_member = 0,
+                    lobby_type = ( int )lobbyType,
+                    max_members = maxMembers,
+                    lobby_flags = lobbyFlags,
+                    metadata = Lobby.EncodeMetadata( metadata ),
+                },
+                SourceJobID = Client.GetNextJobID()
+            };
+
+            Send( setLobbyData, appId );
+
+            lobbyManipulationRequests[ setLobbyData.SourceJobID ] = setLobbyData.Body;
+            return AttachIncompleteManipulationHandler( new AsyncJob<SetLobbyDataCallback>( Client, setLobbyData.SourceJobID ) );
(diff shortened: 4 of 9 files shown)


Benjamin-Dobell:
I'm actually using this in conjunction with a _almost_ complete implementation of Steam's P2P (UDP) protocol that I've develoepd. Combining this PR, with the rest of SteamKit and my Steam P2P implementation is enough to get dedicated servers working for games that use Steam P2P exclusively for networking. Which is precisely why I implemented this and is confirmed as working for Tabletop Simulator.

I will of course open source the P2P stuff at some point, however my gut feel is that it may need to be a stand-alone project that depends on SteamKit2, rather than part of the core. The reason for this is that unlike the rest of SteamKit2, it's specifically designed to be integrated into existing games' run-loops, which means there's a strong focus on performance and avoiding garbage collection. Its API is a bit hairy as a result.

azuisleet:
The scope of this seems OK, as additional lobby functionality can be further built on top.

I don't see anything unusual after a quick review, but I do want to double check the usage and rationale of lobby.Clone() in those callback handlers.

azuisleet:
Would it make sense to have Members be an immutable list of some sort? It shouldn't be necessary to clone Lobby, as reference assignment should be atomic. It would also prevent accidental mutation of the Members list by any other code.

Benjamin-Dobell:
> I don't see anything unusual after a quick review, but I do want to double check the usage and rationale of lobby.Clone() in those callback handlers.
Excellent catch on such a quick review!

I actually _just_ changed it to be "thread-safe by default". In the past the `Clone()` method existed, but wasn't used anywhere. Instead there was documentation stating that `Lobby` (specifically `Lobby.members`) can be mutated out from under you, potentially leading to a crash if you were to simultaneously access it whilst it was being mutated - so you could call `Clone()` in your handler (running on the same thread as `SteamMatchmaking`) to avoid that if you wanted a copy of the data.

That approach seemed a little to error prone, so the objects returned in handlers are not immutable and _won't_ ever be mutated. Instead of mutating lobbies we first `Clone()`, mutate, then update our cached lobby with a new one.

Granted, given `Clone()` is only used internally it may make sense to either:
1. Make it internal.
2. Replace it with a private method on `SteamMatchmaking`.

Benjamin-Dobell:
> Would it make sense to have Members be an immutable list of some sort.

It is actually essentially immutable as currently implemented. `IReadOnlyList` in the public API, it's just internally we clone, then mutate. However, that's realistically a left-over of my previous rationale that I backtracked on.

I think I should just delete the Clone method and explicitly create a new `Lobby` each time a member is added or removed. It's done in two places, so a convenience method would be in order still, but I can make that a private member of `SteamMatchmaking` itself.

Would that be desirable?

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/Lobby.cs
```diff
@@ -0,0 +1,383 @@
...
+                        {
+                            if ( pair.Value == null ) continue;
+                            writer.Write( ( byte )1 );
+                            writer.Write( UTF8.GetBytes( pair.Key ) );
+                            writer.Write( ( byte )0 );
+                            writer.Write( UTF8.GetBytes( pair.Value ) );
+                            writer.Write( ( byte )0 );
+                        }
+
+                        ;
```
Benjamin-Dobell: Ugh, I'll fix this when I address any other feedback.

azuisleet:
Ah, I see what you mean about the interface being IReadOnlyList, but the implementation itself isn't an immutable collection.

I'm suspicious about the clone - mutate pattern. I don't think it's necessary or desirable to clone or create a new Lobby every time. 

I think the implementation should simply create a truly immutable Members list, and assign a new immutable list to Members every time. That way a reference to a Lobby will always be correct, and iterators over Members are never invalidated (because they would always be a reference to the old or new immutable collection)

azuisleet:
Although in this case, lobbies are never mutated in place, so perhaps it may be an unnecessary optimization.

Benjamin-Dobell:
> I think the implementation should simply create a truly immutable Members list, and assign a new immutable list to Members every time. That way a reference to a Lobby will always be correct, and iterators over Members are never invalidated (because they would always be a reference to the old or new immutable collection)

_Ideally_ I'd rather not mutate any objects returned to users. Is there a precedence for this?

The issue would be with users somewhat naively writing:

```C#
// ..lobby is some Lobby retrieved a while back and now being used on another thread.

int count = lobby.Members.Count;
for ( int i = 0; i < count; i++ )
{
    var member = lobby.Members[ i ]; // BOOM!
    // We replaced `Members` with a smaller list mid-iteration because a user left the lobby.
}
```

I think returning purely immutable objects is safest. However, I definitely think making `Clone()` public is a mistake - an oversight from my previous line of thinking.

I could instead have something like this in `SteamMatchmaking`:

```C#
private void UpdateLobbyMembers(uint appId, Lobby lobby, IReadOnlyList<Lobby.Member> members)
{
    var updatedLobby = new Lobby(
        lobby.SteamID,
        lobby.lobbyType,
        // ..etc
        members,
        lobby.distance,
        /// ...etc
    );

    appLobbies = lobbies.GetOrAdd( appId, k => new ConcurrentDictionary<SteamID, Lobby>() );
    appLobbies[ lobby.SteamID ] = updatedLobby;
}
```

Then call that from `HandleUserJoinedLobby` and `HandleUserLeftLobby` with a newly constructed `List`.

Then I can kill this entire concept of `Clone()` and it's all just logic internal to `SteamMatchmaking`.

P.S. Off to bed now, will pick this up tomorrow 😄 Thanks for your time thus far!

azuisleet:
I see the case in that example where they didn't hold an iterator. I think we agree that the Lobby should be wholly immutable, so it's just a matter of details at this point.

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/Callbacks.cs
```diff
@@ -0,0 +1,198 @@
...
+                AppID = appId;
+                Result = res;
+                LobbySteamID = lobbySteamId;
+            }
+        }
+
+        /// <summary>
+        /// This callback is fired in response to <see cref="SetLobbyData"/>.
+        /// </summary>
+        public class SetLobbyDataCallback : CallbackMsg
```
Benjamin-Dobell: Note to self: This and all other callbacks that follow should be `sealed`.

azuisleet:
To clarify one last point, I think this issue also exists in the implementation as-is. Wouldn't an existing reference to Members have this same issue of mutation by the handler, because I don't think the IReadOnlyList interface provides any guarantees, as the underlying implementation is a mutable List?

Ne3tCode:
I don't know C# very well, but it does not seem to work as intended :thinking: 
```csharp
internal List<Member> members;
internal Dictionary<string, string> metadata;

internal Lobby( ... Dictionary<string, string> metadata, List<Member> members, ... )
{
    ...
    metadata = metadata ?? new Dictionary<string, string>();
    members = members ?? new List<Member>();
    ...
}
```

Benjamin-Dobell:
Haha, yeah no knowledge of C# required to notice that's wrong! That's what I get for last second refactoring. Thanks, will fix it shortly.

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/Lobby.cs
```diff
@@ -0,0 +1,383 @@
...
+                }
+
+                /// <summary>
+                /// Checks to see if this lobby member is equal to another. Only the SteamID of the lobby member is taken into account.
+                /// </summary>
+                /// <param name="obj"></param>
+                /// <returns>true, if obj is <see cref="Member"/> with a matching SteamID. Otherwise, false.</returns>
+                public override bool Equals( object obj )
+                {
+                    if ( obj is Member )
```
yaakov-h: Use `obj is Member member` to avoid a second type-check below (when casting).

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/Lobby.cs
```diff
@@ -0,0 +1,383 @@
...
+                /// <summary>
+                /// Metadata attached to the lobby member.
+                /// </summary>
+                public IReadOnlyDictionary<string, string> Metadata { get; }
+
+                internal Member( SteamID steamId, string personaName, IReadOnlyDictionary<string, string> metadata = null )
+                {
+                    SteamID = steamId;
+                    PersonaName = personaName;
+                    Metadata = metadata ?? new Dictionary<string, string>();
```
yaakov-h: Can we have a single empty `Dictionary` instead of one per `Member`?

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/Lobby.cs
```diff
@@ -0,0 +1,383 @@
...
+            }
+
+            /// <summary>
+            /// Return a new <see cref="Lobby"/> instance with equivalent data as this lobby, but is safe to
+            /// access on another thread.
+            /// </summary>
+            /// <returns>A new <see cref="Lobby"/> instance that is equivalent to this one.</returns>
+            public Lobby Clone()
+            {
+                return new Lobby(
```
yaakov-h: How about `object.MemberwiseClone()`?

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/Lobby.cs
```diff
@@ -0,0 +1,383 @@
...
+                    SteamID,
+                    LobbyType,
+                    LobbyFlags,
+                    OwnerSteamID,
+                    new Dictionary<string, string>( metadata ),
+                    MaxMembers,
+                    NumMembers,
+                    new List<Member>( members ), // We're not doing a deep copy as Member is never mutated.
```
yaakov-h: I don't see where we mutate either the dictionary or the list, so why create new ones?

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/Lobby.cs
```diff
@@ -0,0 +1,383 @@
...
+                );
+            }
+
+            internal static byte[] EncodeMetadata( Dictionary<string, string> metadata )
+            {
+                using ( var ms = new MemoryStream() )
+                {
+                    using ( var writer = new BinaryWriter( ms ) )
+                    {
+                        byte[] header = { 0, 0 };
```
yaakov-h: Consider making both of these static+readonly

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/Lobby.cs
```diff
@@ -0,0 +1,383 @@
...
+                        writer.Write( footer );
+                    }
+
+                    return ms.ToArray();
+                }
+            }
+
+            internal static Dictionary<string, string> DecodeMetadata( byte[] data )
+            {
+                Dictionary<string, string> metadata = new Dictionary<string, string>();
```
yaakov-h: Is there a reason you're avoiding `var`?

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/Lobby.cs
```diff
@@ -0,0 +1,383 @@
...
+
+                bool parsingKey = true;
+                string key = null;
+
+                int dataEnd = data.Length - 3;
+
+                if ( dataEnd < 0 || data[ 0 ] != 0 || data[ 1 ] != 0 || data[ 2 ] != 1 || data[ dataEnd + 1 ] != 8 || data[ dataEnd + 1 ] != 8 )
```
yaakov-h: These encoder/decoder functions could do with some unit tests.

yaakov-h: Looking at this binary structure... is this just a `KeyValues` blob? We already have code to deal with that.

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/SteamMatchmaking.cs
```diff
@@ -0,0 +1,392 @@
...
+        /// </summary>
+        /// <param name="appId">The ID of app which we're attempting to obtain a lobby for.</param>
+        /// <param name="lobbyId">The SteamID of the lobby that we wish to obtain.</param>
+        /// <returns>The <see cref="Lobby"/> corresponding with the specified app and lobby ID, if we have its data locally. Otherwise, null.</returns>
+        public Lobby GetLobby( uint appId, SteamID lobbyId )
+        {
+            Lobby lobby;
+            var appLobbies = lobbies.GetOrAdd( appId, k => new ConcurrentDictionary<SteamID, Lobby>() );
+            appLobbies.TryGetValue( lobbyId, out lobby );
+            return lobby;
```
yaakov-h: I don't like ignoring a `TryGet` method and using the `out` param regardless.

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/SteamMatchmaking.cs
```diff
@@ -0,0 +1,392 @@
...
+        /// This method does not send a network request.
+        /// </summary>
+        /// <param name="appId">The ID of app which we're attempting to obtain a lobby for.</param>
+        /// <param name="lobbyId">The SteamID of the lobby that we wish to obtain.</param>
+        /// <returns>The <see cref="Lobby"/> corresponding with the specified app and lobby ID, if we have its data locally. Otherwise, null.</returns>
+        public Lobby GetLobby( uint appId, SteamID lobbyId )
+        {
+            Lobby lobby;
+            var appLobbies = lobbies.GetOrAdd( appId, k => new ConcurrentDictionary<SteamID, Lobby>() );
+            appLobbies.TryGetValue( lobbyId, out lobby );
```
yaakov-h: `out var lobby`

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/SteamMatchmaking.cs
```diff
@@ -0,0 +1,392 @@
...
+        public override void HandleMsg( IPacketMsg packetMsg )
+        {
+            if ( packetMsg == null )
+            {
+                throw new ArgumentNullException( nameof(packetMsg) );
+            }
+
+            Action<IPacketMsg> handler;
+
+            if ( dispatchMap.TryGetValue( packetMsg.MsgType, out handler ) )
```
yaakov-h: `out var handler`

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/SteamMatchmaking.cs
```diff
@@ -0,0 +1,392 @@
...
+                handler( packetMsg );
+            }
+        }
+
+        #region ClientMsg Handlers
+
+        void HandleCreateLobbyResponse( IPacketMsg packetMsg )
+        {
+            var lobbyListResponse = new ClientMsgProtobuf<CMsgClientMMSCreateLobbyResponse>( packetMsg );
+            CMsgClientMMSCreateLobbyResponse body = lobbyListResponse.Body;
```
yaakov-h: inconsistent use of `var` in these functions.

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/SteamMatchmaking.cs
```diff
@@ -0,0 +1,392 @@
...
+            {
+                // Already in lobby
+                return;
+            }
+
+            var member = new Lobby.Member( new SteamID( body.steam_id_user ), body.persona_name );
+
+            lobby = lobby.Clone(); // Cloning to avoid mutating the original Lobby, as it was previously returned in a callback (not thread-safe).
+            lobby.members.Add( member );
+            appLobbies[ lobbyId ] = lobby;
```
yaakov-h: Potential race condition here between lines 323 and now.

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/SteamMatchmaking.cs
```diff
@@ -0,0 +1,392 @@
...
+        void HandleUserJoinedLobby( IPacketMsg packetMsg )
+        {
+            var userJoinedLobby = new ClientMsgProtobuf<CMsgClientMMSUserJoinedLobby>( packetMsg );
+            CMsgClientMMSUserJoinedLobby body = userJoinedLobby.Body;
+
+            var lobbyId = new SteamID( body.steam_id_lobby );
+            var userId = new SteamID( body.steam_id_user );
+
+            var appLobbies = lobbies.GetOrAdd( body.app_id, k => new ConcurrentDictionary<SteamID, Lobby>() );
+            var lobby = appLobbies.ContainsKey( lobbyId ) ? appLobbies[ lobbyId ] : null;
```
yaakov-h: This does two dictionary lookups. Better to use `TryGetValue` for only one lookup.

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/SteamMatchmaking.cs
```diff
@@ -0,0 +1,392 @@
...
+        void HandleUserLeftLobby( IPacketMsg packetMsg )
+        {
+            var userLeftLobby = new ClientMsgProtobuf<CMsgClientMMSUserLeftLobby>( packetMsg );
+            CMsgClientMMSUserLeftLobby body = userLeftLobby.Body;
+
+            var lobbyId = new SteamID( body.steam_id_lobby );
+            var userId = new SteamID( body.steam_id_user );
+
+            var appLobbies = lobbies.GetOrAdd( body.app_id, k => new ConcurrentDictionary<SteamID, Lobby>() );
+            var lobby = appLobbies.ContainsKey( lobbyId ) ? appLobbies[ lobbyId ] : null;
```
yaakov-h: `TryGetValue` as above

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/SteamMatchmaking.cs
```diff
@@ -0,0 +1,392 @@
...
+            {
+                // Not in lobby
+                return;
+            }
+
+            var member = lobby.members[ memberIndex ];
+
+            lobby = lobby.Clone(); // Cloning to avoid mutating the original Lobby, as it was previously returned in a callback (not thread-safe).
+            lobby.members.RemoveAt( memberIndex );
+            appLobbies[ lobbyId ] = lobby;
```
yaakov-h: Should we maybe return clones in the callback rather than having this racy clone-and-replace dance internally every time?

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/Lobby.cs
yaakov-h: We probably should use `AsReadOnly`, otherwise consumers can cast the object reference back to it's mutable implementation.

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/SteamMatchmaking.cs
Benjamin-Dobell: Hmm, aren't the handlers always called on the one thread (Client thread)?

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/SteamMatchmaking.cs
Benjamin-Dobell: I really dislike that syntax, a surprising amount actually. I typically couldn't care less about syntax (brace position, tabs/spaces, whitespace, semicolons, trailing commas, whatever).

However, I can't stand that syntax. It's not a matter of aesthetics either, it just fundamentally violates scope semantics; a dodgy language hack for not having advanced (multi-line) static type inference.

I'd be fine with having the var declared in the correct scope and it inferring the type when used, but it's not supported.

Granted, it's not my decision to make. I'll make the change as requested. However, that's my reasoning.

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/SteamMatchmaking.cs
Benjamin-Dobell: Thanks, I've actually already fixed this locally. Actually, I've fundamentally refactored based on some earlier reviews. Will push shortly.

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/SteamMatchmaking.cs
Benjamin-Dobell: I've refactored this based on earlier feedback. There is no longer a concept of clone. This logic has moved into a `LobbyCache` in a vaguely similar fashion to the `FriendCache`. Will push soon.

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/Lobby.cs
Benjamin-Dobell: Possible, I pulled this code out of another code-base that was quite hacky, and this is my first time using SteamKit, so I probably missed it. I'll take a look, thanks for the hint.

Benjamin-Dobell: As per below, I may be able to piggyback off existing functionality, so unit tests may not be need.

However, the lack of tests in general was concerning me. So I'm actually putting together integration tests at the moment. Granted they can't easily be run by a CI, because you need a Steam account. So I won't push them, but will link out to it in the PR. The idea is you drop in some Steam credentials and it'll run some tests.

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/Lobby.cs
Benjamin-Dobell: Just a 3am oversight when refactoring this code out of my own private codebase back into SteamKit. Already fixed locally.

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/Lobby.cs
Benjamin-Dobell: 👍 Good idea. Will do.

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/Lobby.cs
Benjamin-Dobell: We mutate the list on line 342, granted I've refactored locally. It's purely immutable now.

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/Lobby.cs
Benjamin-Dobell: Ah yeah, sorry, this was running with my .NET 3.5 fork of SteamKit (https://github.com/Benjamin-Dobell/SteamKit/tree/net35) with an old Mono compiler not supporting the syntax. Will make the change 👍

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/Lobby.cs
Benjamin-Dobell: > Can we have a single empty Dictionary instead of one per `Member`?

Can do. Just wasn't sure what the precedent for that was withing this code-base. Very happy to make the change as I do similar in my own code-bases.

> We probably should use `AsReadOnly`, otherwise consumers can cast the object reference back to it's mutable implementation.

I'm not typically a fan of doing of stuff like this. If the user is determined to violate our public contract they can always use reflection etc. It's a bit of fruitless battle.

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/Lobby.cs
yaakov-h: In theory you should be able to use this syntax and still target .NET 3.5... in practice setting up the SDKs like that can be a pain.

Benjamin-Dobell: Yeah, you're spot on.

I have managed to bump the language level a couple of times with my tool-chain. I can probably pull this off now and drastically clean up my (very hacky) backport.

My original issue was that I needed MDBs (Mono debug symbols), which requires using `xbuild` instead of `msbuild`. However, `xbuild` can't go above Language Level 6(?).

I ended up getting fed-up with the archaic tooling, swapped to `msbuild` and threw together a trivial Cecil-based CLI tool to convert the symbols from PDB to MDB (https://github.com/Benjamin-Dobell/pdbtomdb) - `pdb2mdb` doesn't handle the latest "portable" PDB file format. Ah, the joys of the ever changing .NET tooling ecosystem.

Benjamin-Dobell:
Heads up to anyone following this, given the effort I'm going to refactor and test I figured I may as well just add the remaining requests/callbacks. As such, I'll soon be pushing a complete implementation with the exception of:

* Additional filter types; simply because I haven't captured any data for them.
* Steam's "frenemy" stuff, which is deprecated and apparently useless anyway.

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/Lobby.cs
yaakov-h: `<DebugType>full</DebugType>` should get you older PDBs if you need.

Benjamin-Dobell: Ah, sorry, forgot to mention, I'm (generally) on macOS. Older PDBs are Windows only.

Jvs34:
Oh I didn't realize you were also working on this, good job so far.
As far as code goes, I can only give you a tip on the lobby filters, I've looked at the proto packets sent by steamworks and it seems like everything is casted as a string and the filter types that I've seen are implemented in the code snippet below.

https://github.com/Jvs34/SteamKit/blob/c2379a80ef10d36c084e56dd5dd2bc1abefb30a3/SteamKit2/SteamKit2/Steam/Handlers/SteamMatchMaking/SteamMatchMaking.cs#L123-L186

Not sure if I'm actually helping here or not, just wanted to pitch in seems you've been making more progress that I did.

Benjamin-Dobell:
@Jvs34 Thanks, those filter snippets are indeed very helpful. I only had captures of the 3 filters I've implemented. Will happily add those ones in.

At the moment I'm just trying to sort out chat, which has proved painful to capture because it has a bazillion weird variants https://partner.steamgames.com/doc/api/steam_api#EChatEntryType). Tempted to forego it for now. Will see.

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/SteamMatchmaking.cs
azuisleet: I'm not totally certain there's a potential race here. I think most HandleMsg based implementations expect to run in a single thread.

Benjamin-Dobell:
**Not ready to merge**

Still a WIP, mostly regarding chat messages, which I'm admittedly a bit confused about at present.

However, still a couple of house keeping things to do as well e.g. `KeyValue` instead of custom metadata encoding/decoding.

I've force pushed here so I can get some feedback on the direction. Admittedly, that's going to be a pain to diff, so here's the actual change log:

https://github.com/Benjamin-Dobell/SteamKit/commits/matchmaking2

Benjamin-Dobell:
P.S. This is *totally* untested at present. I'm still working on the aforementioned integration tests. I've pushed purely so I can confirm I'm at least taking this in right direction design-wise. Granted, the changes are technically less efficient (more `ConcurrentDictionary` look-ups), but it's a lot more concise and easier to reason about, so I think it's a worthwhile trade-off.

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/SteamMatchmaking.cs
yaakov-h: They are, but I'd need to look closer at what the thread safety concerns _are_ here and how they're being addressed, since we have comments like `// ... (not thread-safe)`.

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/Lobby.cs
```diff
@@ -0,0 +1,360 @@
...
+                public override CMsgClientMMSGetLobbyList.Filter Serialize()
+                {
+                    CMsgClientMMSGetLobbyList.Filter filter = base.Serialize();
+                    filter.value = Value;
+                    return filter;
+                }
+            }
+
+            /// <summary>
+            /// Can be used to filter lobbies by comparing a string against a value in each lobby's metadata.
```
Benjamin-Dobell: Note to self: Incorrect docs.

Benjamin-Dobell:
This is now ready for review and subsequent merging.

## Integration tests

Integration test are available at:

https://github.com/Benjamin-Dobell/SteamKitIntegrationTests

README has instructions. They can be run easily as a Docker container or built and run in an IDE.

However, to specifically run the tests against this PR you'd do:

```
docker build -f dockerfiles/Dockerfile -t steamkit2_tests .
docker run -e PRIMARY_STEAM_USER=account1 -e PRIMARY_STEAM_KEY=key1 -e SECONDARY_STEAM_USER=account2 -e SECONDARY_STEAM_KEY=key2 -e "STEAMKIT_REPO=https://github.com/Benjamin-Dobell/SteamKit.git" -e STEAMKIT_REF=matchmaking steamkit2_tests
```

Note that you must provide credentials for two Steam accounts. These can be free (limited) accounts (lobbies are created against DOTA2 which works because it is also free).

## What's still not included in this PR

- [ ] Additional filters.
- [ ] Assigning a game server to a lobby.
- [ ] Lobby chat.
- [ ] Steam's deprecated "frenemy" functionality.

I basically just don't have a decent way to capture additional filters or chat (nor am I particularly interested in the functionality), however they should be reasonably easy to add on.

## Merging/Review

Changes have been squashed to facilitate merging and a clean history. However, change-sets since previous reviews are available at:

https://github.com/Benjamin-Dobell/SteamKit/commits/matchmaking2

Benjamin-Dobell:
Sorry, just force pushed to eliminate an unwanted change to the `Resources/Protobufs` submodule version.

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/Lobby.cs
Benjamin-Dobell: Just realised I'd missed these two pieces of feedback. Addressing both points now.

Benjamin-Dobell:
## Changes

TLDR; Caching has improved significantly and we now return `AsyncJob` where it makes sense.

In addition to addressing all outstanding feedback.

I've made some significant improvements to caching behavior. This was in particular motivated by the need to simplify the semantics of `Lobby`.`Members`.

1. When a lobby was freshly created, we _didn't_ add it to our cache, because Steam only return confirmation, and not Lobby data in that callback. This is not what a user would expect, and has been addressed with:

    https://github.com/SteamRE/SteamKit/blob/c606289db67be049064b77b938f10339fcb397fd/SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/SteamMatchmaking.cs#L358-L381

2. In addressing the above I realized I'd need a unique ID for each request, so that upon completion I could identify _which_ lobby had been created. I'd seen the `JobID` stuff, but being new to SteamKit I hadn't really paid it much attention. Well, that was silly of me!
    
    After looking into how some other APIs worked I realized I should be using `AsyncJob` for calls that have a direct corresponding callback. So I've gone ahead and done so.

3. Originally I was naively under the impression that the Steam owner wasn't considered a "member" of a lobby. It struck me as odd, however whenever I created a lobby, Steam was returning data stating I was the owner, but the member list returned was blank.
     
     However, whilst writing a couple of additional integration tests, I realised that Steam _does_ include the owner in the member list for other users. So I did some digging and the `ISteamMatchmaking` APIs do indeed consider the lobby owner as a member. Basically, Steam just sends you an empty member list (which is impossible for a lobby) to signify that the lobby's members "haven't changed" i.e. Immediately after creating a lobby, there's indeed one member, you, and you know who you are, so Steam expects that you reconcile this information client-side and doesn't bother sending this down to you.

    Additionally, Steam doesn't provide member or lobby owner details for lobbies in the lobby list. As a result, I had a weird documentation for `Lobby`.`Members` stating that what is represented my be stale and even come out of sync with `NumMembers`; if for example you were in a lobby, and left it, you wouldn't receive updates anymore but the members and owner would still have information in the cache.

    That's confusing, so we now follow `ISteamMatchmaking`'s lead and only make Lobby member (and owner) information available when you're _in the lobby_. We wipe the stale data when the user leaves the lobby, so as to not be confusing. So if you're _not_ in a lobby, `Members` will now consistently be empty _unless_ you're in the lobby, in which case `Member.Count` will always correspond with `NumMembers`... nothing confusing!

4. As with points 1 and 3, there are other ways to modify a lobby (set owner, lobby metadata or member metadata) that worked, but when the callback was being called with `EResult.OK`, if the user was to immediately look in the cache, the data would be state. That's no longer the case, as with point 1, we track all lobby mutation requests such that when a callback is called, if a user calls `Matchmaking.GetLobby(appId, lobbyId)` the lobby returned will be in sync with the callback that's being executed; rather than waiting for Steam to notify us that the lobby data has changed - which happens later.

## Integration Tests

The integration tests have been updated to cover all the changes and aggressively test that data in our cache is as expected, and in line with data returned from Steam itself:

https://github.com/Benjamin-Dobell/SteamKitIntegrationTests/

Output:

https://gist.github.com/Benjamin-Dobell/68b3353a2469b646854a814aaa6f0e54#file-gistfile1-txt-L31-L34

By the way, I'd be quite happy to contribute them back to this repository if anyone thinks it makes sense. Especially since this sort of stuff doesn't really unit test in any sort of meaningful way. However, the integration tests were indeed _very_ helpful in testing the implementation.

azuisleet:
This is looking pretty good. I think my last question is around the lifetime of lobbyManipulationRequests. If the connection is closed, we're probably unlikely to ever see a response.

Resources/SteamLanguage/enums.steamd
```diff
@@ -1205,3 +1205,36 @@ enum EUCMFilePrivacyState flags
...
 	All = Public | FriendsOnly | Private; // 14
 };
+
+enum ELobbyType
+{
+    Private = 0;
+    FriendsOnly = 1;
+    Public = 2;
+    Invisible = 3;
+};
```
xPaw: Add `PrivateUnique = 4` (from v1.45 sdk)

Resources/SteamLanguage/enums.steamd
```diff
@@ -1205,3 +1205,36 @@ enum EUCMFilePrivacyState flags
...
+
+enum ELobbyFilterType
```
xPaw: This is not a Steam enum, is it?

Resources/SteamLanguage/enums.steamd
Benjamin-Dobell: Will do.

Resources/SteamLanguage/enums.steamd
Benjamin-Dobell: @xPaw, it's part of the Steam API, yes. It's just not publicly documented by Valve because it's only used internally.

xPaw: Where did you find the name though?

Benjamin-Dobell: Well, it's reverse engineered, so _I've_ provided a semantic name to what would otherwise be a magic constant.

Resources/SteamLanguage/enums.steamd
```diff
@@ -1205,3 +1205,36 @@ enum EUCMFilePrivacyState flags
...
+{
+    String = 0;
+    Numerical = 1;
+    Distance = 4;
+};
+
+enum ELobbyFilterComparison
```
xPaw: SDK name is `ELobbyComparison`

Resources/SteamLanguage/enums.steamd
```diff
@@ -1205,3 +1205,36 @@ enum EUCMFilePrivacyState flags
...
+{
+    EqualToOrLessThan = -2;
+    LessThan = -1;
+    Equal = 0;
+    GreaterThan = 1;
+    EqualToOrGreaterThan = 2;
+    NotEqual = 3;
+};
+
+public enum ELobbyFilterDistance
```
xPaw: SDK name is `ELobbyDistanceFilter`

Resources/SteamLanguage/enums.steamd
Benjamin-Dobell: I'm _am_ open to changing this. However, what's the precedent here? SteamKit is not Steamworks.

Personally, I think `ELobbyComparison` is a pretty poorly "scoped" name; however, honestly I'm not that attached to it.

If the precedent is to copy Steamworks, then I'll make the change.

xPaw: SteamKit tries to follow Valve's names (SDK or RE). Makes it easier to cross-reference and update.

Benjamin-Dobell: Alright, I'll make the change.

However, is that policy documented somewhere? If not, would be good to jot it down so future contributors know.

Benjamin-Dobell:
> I think my last question is around the lifetime of lobbyManipulationRequests. If the connection is closed, we're probably unlikely to ever see a response.

Does the job manager not provide callback guarantees (a "promise")? If not, wouldn't this be an issue for callbacks users make as well?

If there's no guarantee, then I'd be inclined to think that ought to be fixed up in another PR. Unless I've fundamentally misunderstood something... possible 😉

Resources/SteamLanguage/enums.steamd
yaakov-h: There's not much that's documented full stop

azuisleet:
The job manager does provide promises, which are all cancelled in OnClientDisconnected, as it's unlikely they will ever get resolved at that point. `lobbyManipulationRequests` should be cleared that point for sure, as we know those requests are abandoned.

The last piece I'm worried about is the ever-growing lobby cache. Is it sufficient to clear the `lobbies` cache on disconnect too?

Benjamin-Dobell:
> `lobbyManipulationRequests` should be cleared that point for sure, as we know those requests are abandoned.

Makes sense, will do.

> The last piece I'm worried about is the ever-growing lobby cache. Is it sufficient to clear the lobbies cache on disconnect too?

I'll clear it on disconnect also, however if there's further suggestions surrounding this, I'll happily take those on board.

I'm pretty sleep deprived at the moment, so may have just missed it, however I had a quick glance at the friend cache and couldn't find clean-up code. Is it being cleaned up on disconnect? Should it be?

azuisleet:
The friend cache I believe suffers from this issue, but of course to fix it someone will have to bring Friends back up to date.

Benjamin-Dobell:
I've made the requested changes and also added the two outstanding filter types (`NearValueFilter` and `SlotsAvailableFilter`).

As usual, changes are squashed but individual commits are available at https://github.com/Benjamin-Dobell/SteamKit/commits/matchmaking2. Although, it's worth noting that even `matchmaking2` has been force pushed since previous reviews as I've rebased this PR on master (in order to resolve merge conflicts re: steam language/enums).

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/Lobby.cs
```diff
@@ -0,0 +1,402 @@
...
+                }
+
+                using ( var ms = new MemoryStream() )
+                {
+                    keyValue.SaveToStream( ms, true );
+
+                    var buffer = new byte[ ms.Length ];
+                    ms.Seek( 0, SeekOrigin.Begin );
+                    ms.Read( buffer, 0, buffer.Length );
+                    return buffer;
```
yaakov-h: This seems like a reimplementation of `MemoryStream.ToArray()`?

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/Lobby.cs
```diff
@@ -0,0 +1,402 @@
...
+
+            /// <summary>
+            /// The weight of the lobby.
+            /// </summary>
+            public long? Weight { get; }
+
+            static readonly IReadOnlyDictionary<string, string> EmptyMetadata =
+                new ReadOnlyDictionary<string, string>( new Dictionary<string, string>() );
+
+            static readonly IReadOnlyList<Member> EmptyMembers = new List<Member>().AsReadOnly();
```
yaakov-h: Nit: I'd prefer using `Array.Empty<Member>()` instead of a new list, but the difference is negligible.

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/Lobby.cs
```diff
@@ -0,0 +1,402 @@
...
+                Members = members ?? EmptyMembers;
+                Distance = distance;
+                Weight = weight;
+            }
+
+            internal static byte[] EncodeMetadata( IReadOnlyDictionary<string, string> metadata )
+            {
+                if ( metadata == null )
+                {
+                    metadata = EmptyMetadata;
```
yaakov-h: Why do we assign EmptyMetadata here and then iterate over it? If we're doing a null check anyway, then we can just do `if (metadata != null) { foreach (var entry in metadata) { ... } }`.

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/Lobby.cs
```diff
@@ -0,0 +1,402 @@
...
+                    return buffer;
+                }
+            }
+
+            internal static ReadOnlyDictionary<string, string> DecodeMetadata( byte[] buffer )
+            {
+                var metadata = new Dictionary<string, string>();
+
+                if ( buffer.Length > 0 )
+                {
```
yaakov-h: Can we move the allocation of the dictionary here, and re-use `EmptyMetadata` if the length is zero?

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/SteamMatchmaking.cs
```diff
@@ -0,0 +1,664 @@
...
+            lobbyCache.Clear();
+        }
+
+        AsyncJob<T> MonitorJob<T>( AsyncJob<T> job )
+            where T : CallbackMsg
+        {
+            job.ToTask().ContinueWith( task =>
+            {
+                lobbyManipulationRequests.TryRemove( job.JobID, out _ );
+            }, TaskContinuationOptions.NotOnRanToCompletion );
```
yaakov-h: It took me a while to figure out what this is doing and why we're using this particular `TaskContinuationOptions`. Can you leave a comment in here explaining it?

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/Lobby.cs
Benjamin-Dobell: Yep, good catch.

Other stuff I've been working on recently uses this sort of pattern to avoid memory allocations by using an existing buffer. However, that makes zero sense here because I've just created a brand new buffer on line 360.

I'll replace it with `buffer.ToArray()`.

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/Lobby.cs
Benjamin-Dobell: It'd have to be:

```C#
Array.AsReadOnly(Array.Empty<Member>());
```

However, yeah, I might as well change it seems as it's conceivable that user code may also use `Array.Empty<Member>()` and we'd get the benefit of just the one copy.

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/Lobby.cs
Benjamin-Dobell: I don't think I thought about it _that_ hard when I wrote it. However, I'd say my thought process was something along the lines of "this is essentially a default parameter so we'll handle it up front and minimise cyclomatic" i.e. it's optimised for comprehension rather than runtime efficiency.

However, honestly, I really don't have a strong opinion about it.

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/Lobby.cs
Benjamin-Dobell: Sure... and after my previous comment; here I am writing code with worse cyclomatic complexity :wink:

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/Lobby.cs
yaakov-h: Also, allocating an empty list gets you both a List<> object and an empty array underneath (though it's its own version instead of Array.Empty's). Using the empty array directly means one less object - the List.

SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/SteamMatchmaking.cs
Benjamin-Dobell: If the code's not easy to comprehend it's probably just subpar code. So I'll refactor a smidgen and see if that hits the spot. If that still doesn't do it, let me know and then as a last resort I'll throw in a comment as well.

Benjamin-Dobell: It's probably not the _best_ named method ever, however do you think what I've got now is sufficient?

https://github.com/Benjamin-Dobell/SteamKit/commit/ae8487d47462e1cbffe6ca4d655f59446a439c18

yaakov-h: I don't think so.

The bit I had trouble with was "okay we'll remove it on faulted or cancelled, but why not on completed?"

Renaming the method to indicate that we handle incomplete outcomes doesn't explain why we don't do it for completed ones.

Benjamin-Dobell: That's fair. Comment added.

Benjamin-Dobell:
@yaakov-h Yeah, I felt a bit weird about namespaces/scoping as well e.g. using partial classes for non-generated code seemed a bit unnecessary. I was _trying_ to follow established patterns, but I however this is certainly one of the more involved handlers. So I think I certainly stretched the current code layout pattern to its limits.

I'd have absolutely zero issue with someone coming along and reorganising.

yaakov-h:
Thanks a bunch!

If you still want to implement the remaining few items mentioned in the OP, please open a new PR.

If you don't, that's perfectly fine as well. 😄

Tyaap:
Hello,

I am working on a project that requires checking if a lobby with a given Steam ID exists, and if so retrieve its data. I believe this requires the RequestLobbyData() function and LobbyDataUpdate_t callback to check for success. The LobbyDataUpdate_t callback would also useful for checking if GetLobbyData() fails, without waiting for a timeout. Is it possible for these to be implemented?

Benjamin-Dobell:
@Tyaap Everything you need is already implemented. It's a simple, fairly self explanatory API, as such please refer to the changes in this PR to find the appropriate methods and callbacks.

Tyaap:
Hello Benjamin,

I am aware that lobbies may be checked for existence by calling GetLobbyList and searching the returned list. However there is a problem when the lobby is full: Steam never returns these lobbies. I have tried adding a filter to request lobbies with >= 0 slots but this does not help. So it seems I can't check if a full lobby exists using this approach. Also I can't use GetLobbyData since this only works for lobbies already retrieved via GetLobbyList. The approach mentioned previously should work, since the Steamworks documentation says that RequestLobbyData does not require a prior call to GetLobbyList. However this function does not appear to be present in SteamKit. Is there something I'm still not aware of that can resolve this problem?

Benjamin-Dobell:
Steamworks documentation applies to Steamworks, not SteamKit. SteamKit is a C# library with C# semantics. What you want to do is already possible, as previously suggested, please refer to the changes made in this PR.

Tyaap:
I have read the code in SteamMatchmaking, which seems to suggest using GetLobbyData. However I have tried using the implementation of GetLobbyData in SteamKit (without a ptior call of GetLobbyList) and the client is immediately disconnected from Steam.

Benjamin-Dobell:
@Tyaap How did you retrieve the steam ID for this lobby? Steam will only let you retrieve data for lobbies you should know about. That means those publicly available in the lobby list, or a friend's lobby. Steam's backend won't allow you request the data for an arbitrary lobby.

By the way, there is no such API as "RequestLobbyData". Steamworks's `RequestLobbyData()` function would just be calling the same API we're calling behind the scenes in our `GetLobbyData()`, which is called `ClientMMSGetLobbyData`.

Steamworks `RequestLobbyData()` will just be wrapping that call with a local check to see if it's a lobby you're allowed to request. If not, I imagine it prevents the request being sent in the first place, rather than disconnecting you.

Now, technically, even if your friend is in a lobby, you can only know this information if you've requested their `PersonaState`. Steam's back-end is _very_ good at keeping track of what information you _should_ know. So if you haven't requested said friend's persona state, then a disconnection is probably in order.

There _is_ actually some related functionality that is missing from SteamKit. Specifically, [PersonaStateCallback](https://github.com/SteamRE/SteamKit/blob/d7b13702319b1dcd0997772551de6028030c83e3/SteamKit2/SteamKit2/Steam/Handlers/SteamFriends/Callbacks.cs#L70) does not expose the lobby ID. If you'd like to implement that, it'd be two line (+ 4 lines of documentation) change.

Still though, this is only ever going to work for lobbies that your friends are in; or users which Steam's backend has for some reason allowed you to request the persona state of.

Tyaap:
Thanks for the information. I am retrieving the lobby list on a regular basis using GetLobbyList, and storing the details. The goal is to keep track of all active lobbies. Sometimes on retrieving a new list a lobby will disappear. It will either disappear because: 
1. It has been deleted.
2. The lobby is full.
I wish to check which case it is, and retrieve the lobby data in case 2.

For reference, this is the relevant code I'm working on:
https://github.com/Tyaap/SuperLobbyBot/blob/3583d8242ab8a4bff7694895be0fb191cf6e9f05/Steam.cs#L388

Tyaap:
So after reviewing your message, it appears futile to try work around the "disappearance" of full lobbies from the list retrieved from GetLobbyList. The Steamwork's documentation states that hiding full lobbies is the "default" behaviour, but gives no indication that this default behaviour can be changed. Are you aware of any part of your Matchmaking implementation which may allow me to retrieve information about a full lobby?

Tyaap:
I have tried this, but no luck: https://github.com/Tyaap/SuperLobbyBot/blob/d207fb2f852b95c40e822bb5d1da10db4a8f9199/Steam.cs#L171

Tyaap:
Update: I have tried using the native Steamworks API (via Steamworks.NET) and have successfully kept track of full  lobbies using the RequestLobbyData() function. However the behaviour is somewhat unexpected: the requested lobby data is not retrieved until I call RetrieveLobbyList(). This may be caused by Steamworks.NET but I can't tell. I have created a proof of concept project: https://github.com/Tyaap/LobbyTrackingPOC/blob/master/Program.cs
I wish to be able to do the same as this with SteamKit, but it is not clear how this can be done.

Benjamin-Dobell:
@Tyaap Just to clarify, you've confirmed the data returned is up-to-date i.e. it's not just serving you cached data (before the lobby was full)?

Benjamin-Dobell:
@Tyaap Assuming you're not receiving stale data, then what I suspect is happening is that Steam is just sending the client a regular lobby update for any lobbies that are no longer in the latest lobby list. Your call to `RequestLobbyData()` will be unrelated.

Have you subscribed for regular lobby data updates (`LobbyDataCallback`)?

Tyaap:
I can confirm it is up to date, because the player count is reported as e.g. 4/4. It must have been less than this perviously, otherwise it wouldn't have shown up in a prior retrieval of the lobby list. This data is not retrieved unless I call RequestLobbyData().

In SteamKit, I have tried subscribing to LobbyDataCallback. This is only getting triggered if I run GetLobbyData() on a lobby that was retrieved from GetLobbyList(). The test I have done is: run GetLobbyList() and periodically execute GetLobbyData() on each of the returned lobbies. Eventually this resulted in a Steam disconnection, which I believe is because the lobby disappears. I am unable to confirm this though, or determine whether its due to a lobby that truly no longer exists or one that is just full. I've made a test project for that and will upload it shortly.

Tyaap:
https://github.com/Tyaap/GetLobbyDataTest/blob/master/Program.cs

Benjamin-Dobell:
I was just thinking that the idea of Steam disconnecting you simply for requesting data for a lobby you'd previously been told about (but is now full) leading to a disconnection seemed a bit dubious. Because that'd be susceptible to all sorts of race conditions and would like break standard clients...

This does not appear to be the case.


I don't know why Steam is disconnecting you, however here's a very simple test that demonstrates you can query data for full lobbies you've previously seen in a lobby list, but are _not_ being returned in the latest lobby list.

https://github.com/Benjamin-Dobell/SteamKitIntegrationTests/blob/full-lobby-data/SteamKitIntegrationTests/Program.cs#L38-L98

To run the tests, checkout https://github.com/Benjamin-Dobell/SteamKitIntegrationTests/tree/full-lobby-data (Note: `full-lobby-data` branch). You can either follow the README to run it in a Docker container, or create a `.env` file in the project (`./SteamKitIntegrationTests`) directory with the contents:

```
PRIMARY_STEAM_USER=your_first_steam_account
PRIMARY_STEAM_PASSWORD=some_secret_password1
SECONDARY_STEAM_USER=your_second_steam_account
SECONDARY_STEAM_PASSWORD=some_secret_password2
```
_or `_KEY` instead of `_PASSWORD` if you prefer._

Then run either run/debug the singular test from your IDE of choice, or from command line with:

```
dotnet restore
dotnet build /p:Configuration=Debug /p:EnableSourceLink=false
dotnet test --no-build
```

Tyaap:
Thank you very much for the test code. :) Unfortunately I can't test it (without further work) as I'm getting AccountLogonDenied due to my Steam Guard. I've had a look and so far haven't found anything within the code snippet that is majorly different to what I have been trying to do, which is very odd. I will have to investigate some more.

Tyaap:
I suspect the disconnects are caused by attempting to get data for a lobby that truly no longer exists. If this is the case, then there is still an issue of being able to differentiate between these two cases in order to avoid being disconnected.

Benjamin-Dobell:
@Tyaap You're being disconnected because you're requesting a _deleted_ lobby, not a full one. Just verified this happens.

However, Steam _does_ send you updates automatically (without you explictly requesting) for lobbies that you've previously received data for but have since been deleted.

Now, that still seems a bit off to me, because there's potential for a race condition in that you could go to request the data for a lobby, but it's _just_ deleted.

I'm thinking the only way the real Steam client could get around this is that Steam's backend will allow you to make this request, _until_ your client has _confirmed_ receiving the packet that indicates the lobby has been deleted. Then when processing its queue Steam would see a queued outbound message pertains to a deleted resource (lobby), delete the message, and return the data from the lobby cache.

That's actually insanely convoluted, so I'd love to hear I'm wrong and there's a simpler way.

If Steam are doing the above, then we have an huge problem on our hands, as we send and receive logic operate in different threads. Everything is event driven, we really can't process everything serially like that without _enormous_ changes to the architecture of SteamKit 😕

I'd suggest you play around and ensure you're seeing what I am, if you are, you're going to need to open an issue... but ah, if it really is the above, I wouldn't expect it to be fixed anytime soon.

Tyaap:
Okay I am getting the same results as you for the case of a full lobby and the case of a deleted lobby. :(
This is certainly a major over-complication on Steam's part. It seems almost absurd to implement all that when a success/deleted/denied status code on GetLobbyData would have been sufficient to indicate active/deleted lobbies without revealing info about lobbies not previously received.

Tyaap:
I will open an issue about this. For now I have thought of a workaround although its not pullet proof. I will just assume that any lobbies that disappear with less than e.g. MaxMembers/2 was a deleted lobby, and a lobby with more than that is a full lobby. Hopefully the likelihood of these assumptions being incorrect is negligible.

Tyaap:
Do you have code that I can refer to in the new issue, which can quickly reproduce it?

Benjamin-Dobell:
Sure:

https://github.com/Benjamin-Dobell/SteamKitIntegrationTests/commit/bd8fd31a15723c3ae87d4256b2c3618122f9df41

The failing code is commented out, the code that's there is actually guaranteed to work in this circumstance.

Although far from ideal, it's probably worth keeping in mind that the likelihood requesting lobby data just as a lobby is deleted and _before_ `LobbyDataCallback` is fairly small - well, depending on how many lobbies you're monitoring.

Rather than doing the weirdness you suggested above with guessing about lobby states, it'd probably be easier just to throw your Bot up behind foreman and restart it if it crashes. Could split the Steam logic out into a separate process/container, so if the Steam querying component goes down, your Bot doesn't go down itself.

Well, unless you're worried about being banned by Valve. I imagine making bad requests does raise some eyebrows - even if they are infrequent.

Tyaap:
Okay thanks for the help. I have just read through your code. I was not aware that after retrieving a lobby list we get LobbyDataCallbacks from Steam to update the lobby data, including one that changes the steam ID of the lobby owner to 0 to indicate the destruction of the lobby. I will have to change the issue description again...

Benjamin-Dobell:
Unfortunately, it seems that it's not simply retrieving the lobby list that sets you up for lobby updates. If it was, you'd actually have a 100% crash-free solution on offer, with the ever so slight possibility of mislabelling a deleted lobby as full for a few milliseconds i.e. Never call `GetLobbyData()`.

_However_, it would seem you only receive deletion updates if you've called `GetLobbyData()` on the lobby at least once before - and of course, as we've discussed, we can't guarantee it's safe to call even once.

expected-ingot:
> I'm actually using this in conjunction with a _almost_ complete implementation of Steam's P2P (UDP) protocol that I've developed. Combining this PR, with the rest of SteamKit and my Steam P2P implementation is enough to get dedicated servers working for games that use Steam P2P exclusively for networking. Which is precisely why I implemented this and is confirmed as working for Tabletop Simulator.
> 
> I will of course open source the P2P stuff at some point, however my gut feel is that it may need to be a stand-alone project that depends on SteamKit2, rather than part of the core. The reason for this is that unlike the rest of SteamKit2, it's specifically designed to be integrated into existing games' run-loops, which means there's a strong focus on performance and avoiding garbage collection. Its API is a bit hairy as a result.
> 
> **EDIT:** Oh, and related to above, in case anyone is interested. I've also back-ported SteamKit2 to .NET 3.5 (https://github.com/Benjamin-Dobell/SteamKit/tree/net35). It's a few months behind upstream master, but surprisingly usable.
@Benjamin-Dobell 

I'm very sorry for bringing up such an old issue, however, I have a question: How did you manage to get SteamKit working with Tabletop Simulator?

I want to port a game to another platform (which doesn't have a Steamworks SDK), and I'm interested in using SteamKit in order to implement Steam functions, such as Steam lobbies. However, as far as I know, there aren't any projects that use SteamKit in order to make a DLL that acts similarly to Steamworks.

Benjamin-Dobell:
@expected-ingot I didn't make a drop-in DLL replacement for the full Steamworks SDK

Instead I replaced another DLL (an entrypoint to my own code) and then used .NET runtime patching of relevant Tabletop Simulator specific APIs. I patched those APIs (networking etc.) to call through to both SteamKit and my own Steam P2P implementation. I used these lobby APIs to register my custom dedicated server so regular TTS clients could find and connect to my server.

Actually, I implemented SteamKit and the Steam P2P in a separate process (basically a proxy) and the TTS patches communicated with it over a local socket. The whole solution was very specific to TTS, not a general purpose solution for replacing Steam.

PR #707, opened 2019-07-22
Add ParentalSettings to LoggedOnCallback

JustArchi:
I've verified that this works properly, and it's a big help for those apps that want/need to interpret those settings, for example in order to determine if `ParentalSettings.is_enabled`.

I could expose `resp.parental_settings` alone and expect from the consumer to do the deserialization himself, but I believe that we can aid him a little since there is nothing else he should be doing with that data anyway. Let me know if there is anything you'd change in this implementation.

Thank you in advance for considering this PR.

### DIFF
--- a/SteamKit2/SteamKit2/Steam/Handlers/SteamUser/Callbacks.cs
+++ b/SteamKit2/SteamKit2/Steam/Handlers/SteamUser/Callbacks.cs
@@ -158,7 +158,7 @@ namespace SteamKit2
                 {
                     using ( var ms = new MemoryStream( resp.parental_settings ) )
                     {
-                        ParentalSettings = Serializer.Deserialize<ParentalSettings>( ms );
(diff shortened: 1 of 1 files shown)


SteamKit2/SteamKit2/Steam/Handlers/SteamUser/Callbacks.cs
```diff
@@ -147,6 +153,14 @@ internal LoggedOnCallback( CMsgClientLogonResponse resp )
 
                 this.NumLoginFailuresToMigrate = resp.count_loginfailures_to_migrate;
                 this.NumDisconnectsToMigrate = resp.count_disconnects_to_migrate;
+
+                if ( resp.parental_settings != null )
```
JustArchi: To the best of my knowledge this should never be `null` but at worst null data in a deserialized protobuf body, but better safe than sorry.

PR #708, opened 2019-07-28
Improve SK2 versioning for CI and non-CI builds

JustArchi:
Developers building/publishing in non-CI environments shouldn't be expected to correct the versioning, at least in regards to the static part of SK2 versioning, such as "2.2.0". This is required for assembly compatibility, developer targetting 2.2.0+, self-compiling SK2 before this PR produced `2.0.0.0` assembly, which is incorrect (we've built `master` which is 2.2.0 as of today). Assembly patching done in AppVeyor CI is OK for automating publishing procedure, and the build number can be very helpful for verification, but self-compiled builds lacking that info should be compatible with static SK2 versioning.

I've also killed duplicate definition in `appveyor.yml` to hopefully make up for the extra hassle caused 😅.

Thanks for considering.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,0 +1,30 @@
+<Project>
+    <PropertyGroup>
+        <SteamKitRootDir>$(MSBuildThisFileDirectory)</SteamKitRootDir>
+    </PropertyGroup>
+
+    <PropertyGroup Condition=" '$(APPVEYOR_BUILD_VERSION)' == ''">
+        <AppVeyorConfigPath>$(SteamKitRootDir)appveyor.yml</AppVeyorConfigPath>
+        <AppVeyorConfigText>$([System.IO.File]::ReadAllText($(AppVeyorConfigPath)))</AppVeyorConfigText>
+        <AppVeyorVersionExpression>CoreVersion: (\d+\.\d+\.\d+)</AppVeyorVersionExpression>
+        <SteamKitCoreVersion>$([System.Text.RegularExpressions.Regex]::Match($(AppVeyorConfigText), $(AppVeyorVersionExpression)).Groups[1].Value)</SteamKitCoreVersion>
+        <SteamKitVersion>$(SteamKitCoreVersion).0</SteamKitVersion>
+        <AssemblyVersion>$(SteamKitVersion)</AssemblyVersion>
+        <FileVersion>$(SteamKitVersion)</FileVersion>
+        <InformationalVersion>$(SteamKitVersion) - Development</InformationalVersion>
+    </PropertyGroup>
+    <PropertyGroup Condition=" '$(APPVEYOR_BUILD_VERSION)' != ''">
+        <AssemblyVersion>$(APPVEYOR_BUILD_VERSION)</AssemblyVersion>
+        <FileVersion>$(APPVEYOR_BUILD_VERSION)</FileVersion>
+        <SteamKitBranch Condition=" '$(APPVEYOR_PULL_REQUEST_HEAD_REPO_BRANCH)' != '' ">$(APPVEYOR_PULL_REQUEST_HEAD_REPO_BRANCH)</SteamKitBranch>
+        <SteamKitBranch Condition=" '$(SteamKitBranch)' == '' ">$(APPVEYOR_REPO_BRANCH)</SteamKitBranch>
+        <InformationalVersion>$(APPVEYOR_BUILD_VERSION) - CI (AppVeyor, branch: $(SteamKitBranch))</InformationalVersion>
+    </PropertyGroup>
+    <PropertyGroup>
+        <GenerateAssemblyTitleAttribute>false</GenerateAssemblyTitleAttribute>
+        <GenerateAssemblyDescriptionAttribute>false</GenerateAssemblyDescriptionAttribute>
+        <GenerateAssemblyConfigurationAttribute>false</GenerateAssemblyConfigurationAttribute>
+        <GenerateAssemblyCompanyAttribute>false</GenerateAssemblyCompanyAttribute>
+        <GenerateAssemblyProductAttribute>false</GenerateAssemblyProductAttribute>
+    </PropertyGroup>
+</Project>
--- a/SteamKit2/SteamKit2/Properties/AssemblyInfo.cs
+++ b/SteamKit2/SteamKit2/Properties/AssemblyInfo.cs
@@ -29,11 +29,6 @@ using System.Runtime.InteropServices;
 // The following GUID is for the ID of the typelib if this project is exposed to COM
 [assembly: Guid( "cd42d0bc-72e4-451e-bcd0-5d09c4bca2a9" )]
 
-// These are automatically modified by AppVeyor for CI builds and automated deployment, but you should still try to keep them up-to-date for non-CI builds.
-[assembly: AssemblyVersion( "2.2.0" )]
-[assembly: AssemblyFileVersion( "2.2.0.0" )]
-[assembly: AssemblyInformationalVersion( "2.2.0.0 - Development" )]
-
 #if DEBUG
 [assembly: InternalsVisibleTo( "Tests" )]
 #endif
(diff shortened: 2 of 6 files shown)


yaakov-h:
The appveyor changes are welcome, but I'm not sold on the assembly attributes.

Local builds should be easily distinguishable from official releases. I know Roslyn used to use, or maybe still does, `42.42.42.42` for example.

JustArchi:
Except this brings problem on dependencies. Example: I have my ASF project where I declare SK2 in version 2.2.3, and people base their own plugins on my ASF and SK2 in version 2.2.0. Then I temporarily need to include my own SK2 version, which is let's say 2.2.4 or 2.3.0. I build it and include, except assembly is still `2.0.0` now. My project is no longer able to load that plugin, because plugins are targetting `2.2.3`, and I have `2.0.0`, while in reality it's `2.2.4` or `2.3.0`.

To answer your future question, yes, just changing the version fixes **everything** for me. This is what I did.

> Local builds should be easily distinguishable from official releases

They already are. Last version number will always be `0` for the local builds, and I don't need/want to change that. This can be used for official releases as AppVeyor always appends build number there. I want first three numbers to match local builds, last one can differ without issues.

Let me know what you think.

yaakov-h:
On Framework you should be able to solve your use case with assembly binding redirects.

On Core, the version number isn't used by the assembly loader in the same manner, so it should just work AFAIK.

In any event, we don't (usually) promise binary compatibility between versions, so this sounds like a disaster waiting to happen regardless.

JustArchi:
You don't need to promise anything, it's simply aid for the developer so he doesn't need to set his version manually according to the current SK2 build.

Look, if AppVeyor builds from commit `abcde1` and it results in version `2.2.1.123`, then the developer building from the same commit `abcde1` should **also** get `2.2.1.0` at the minimum, those two builds are equal, there is no reason in the world why second one should have `2.0.0.0` just because it wasn't built by CI.

Likewise, if the developer adds his own changes and compiles, the burden on binary compatibility is on him, exactly the same as now. You don't want to tell me that his build is binary-compatible with `2.0.0.0` build, right? Then why not claim `2.2.1` right away, if we can and it's definitely closer to the real version than `2.0.0`?

I just don't understand the point behind this, you need to set some version anyway, why not claim the version that is the closest to the real one, while **still** leaving the room for distinction between self-builds and CI? That `2.2.1` announced as `2.0.0` is not binary-compatible with it either.

yaakov-h:
> if AppVeyor builds from commit `abcde1` and it results in version `2.2.1.123`, then the developer building from the same commit `abcde1` should also get `2.2.1.0`

_Maybe_, and even then, only if their repo has no changes. A dirty repository will yield a different build. And if the repository was perfectly clean, then why not use the CI build?

Typically a developer building locally doesn't need to set a version number, period.

Loading pre-alpha builds into a version-sensitive environment isn't a supported use-case... at least, it's not something we've thought of, and not something that I think we want to support. And if it's something that on occasion we do need to do, then we can modify the versions manually.

(Though by now we can probably migrate that into MSBuild properties anyway)

JustArchi:
> Maybe, and even then, only if their repo has no changes. A dirty repository will yield a different build.

Except you do not control it. Developer can set whatever version he wants for whatever source he wants, including copying the exact version from your CI and applying it to his dirty build with troublesome changes, and you still have no influence over that. I could even write a script that reads your `appveyor.yml` and writes appropriate lines before each build, exactly the same what CI does as post-build step. Problem is that this doesn't benefit anybody, people with ill intentions will do ill things anyway, and you're only causing maintenance burden on those that would like from their self-builds to automatically inherit the closest SK2 version that they've compiled against.

Moreover, I find your entire reasoning feeling like made-up arguments just to not merge this PR and justify rejecting it, and I don't understand why. You're talking about pre-alpha builds into version-sensitive environment and how it isn't a supported use-case, but every single CI build, including out-of-tagged-release debug builds carry appropriate version set. You said yourself that the developer can set the versions manually as well, and the existing version isn't binary-compatible with `2.0.0` release either. I don't understand your point entirely. In my ASF I did a very simple trick, I'm appending `-custom` to the program identifier if the program was built from the source and not my CI. Is developer will ill intentions able to cover that up and confuse me when asking for support, help or anything else? Of course he is able to, he'd be able to do that regardless how hard I'd make it for him, except since he inherits my versioning, I know range of commits he could've based against, and I can help him without even looking at his source modifications, because either something applies to my code in that particular version, or something does not. I prefer to know that the source was modified, and at what point of time, rather than developer coming to me with faked version lying that he's using official release just to get help regarding broken functionality that doesn't consider my code.

I still don't see any real reason for you to stick with static `2.0.0` as opposed to using correct first three versioning numbers. If anything, it only encourages people with forks to outright lie in eventual future issues, covering up their modifications and wasting your time on answering matters that do not apply to your project, instead of having clear sign of forked version, checking against the same source and either reproducing the problem or not.

But of course the decision is up to you and the rest of the development team, you can just close this PR if you're still not convinced and you prefer to cause burden to both parties for some unknown to me reasons, just think twice if what you're doing makes any sense, as it'd be better for everybody if people could base on the existing logic and benefit both you and them because of it, instead of doing secret versioning shenanigans just because you force them to.

xPaw:
Here's a simple issue: your PR makes it so you need to change version in two files instead of one, making releases through github's web interface is harder because of that.

If you can make appveyor take version from AssemblyInfo.cs (or append build id to it) then it could be less of a problem.

JustArchi:
And this is a valid concern I've mentioned myself:

> I've also killed duplicate definition in appveyor.yml to hopefully make up for the extra hassle caused.

I had no good idea how to do it in a good way, feel free to improve on this idea if you do.

yaakov-h:
This is way off track from a simple

> but I'm not sold on the assembly attributes

and frankly, you haven't convinced me.

I don't see a supported use-case where that part of the PR assists the project's goals, and it introduces a maintenance burden as @xPaw mentioned, which most likely nobody is going to pay attention to.

Instead of being "stuck" at 2.0.0.0, we'll be stuck at 2.2.0.0...

> you're only causing maintenance burden on those that would like from their self-builds to automatically inherit the closest SK2 version that they've compiled against

Like I said, there's no compelling use-case that makes this an aim of SteamKit. This isn't something we've tried to do, and have no reason to start doing so.

> If anything, it only encourages people with forks to outright lie

This sounds like a made-up argument too. We have no high-profile forks, and as far as I know, nobody has formalized a fork that is actually shipping code. If someone decides to create a full-blown fork, I'd expect they'd manage their own CI/CD etc., but however they manage it is not our concern as the upstream project.

> I still don't see any real reason for you to stick with static `2.0.0` as opposed to using correct first three versioning numbers

Inertia and maintenance cost?

---

I see basically three parts to this PR:

1. Bump copyright year
2. Bump assembly attributes
3. Remove version duplicate in CI/CD script.

(1) and (3) are welcome and I would gladly merge a PR that does only these.

(2) doesn't seem to achieve anything within the project scope.

The downsides are that it adds an additional task to the release process which either needs to be documented and followed, or will just be neglected and we land up back where we are now, until somebody comes along and adds a PR to bump it to `2.3.0.0` or whatever.

The reasons to have it don't seem to outweigh the downsides above. It doesn't drive the project forwards in a meaningful way, and the version number doesn't really _mean_ anything except on .NET Framework, which can easily be overcome with a small snippet of XML.

---

> If you can make appveyor take version from AssemblyInfo.cs (or append build id to it) then it could be less of a problem.

This would be going backwards, since the new Core SDK / project format emits assembly attributes at build time from MSBuild properties...

yaakov-h:
I have an idea, we might be able to get MSBuild to calculate it's version number from appveyor.yml...

JustArchi:
Wouldn't the opposite way be better? You could put versioning in msbuild properties and make AppVeyor read it and change only the revision number.

yaakov-h:
I'd rather define it in appveyor so that it gets inherited from `git tag` events etc.

yaakov-h:
Double-PR at https://github.com/JustArchi/SteamKit/pull/1 requesting amendments to this PR.

JustArchi:
You can just commit directly if you want.

image

But of course I'll happily marge, thanks!

JustArchi:
I like a lot what @yaakov-h has made, I didn't even know you could do such stuff within msbuild process. This is pretty much what I wanted to achieve initially but had no clue how.

Thanks a lot!

yaakov-h:
(I don't like stomping on PRs without permission. 😄)

Issue #587, opened 2018-08-13
Proxy in steam client


MakcStudio:
Can I use a proxy to connect to steam client? How?

mSteam.client.Connect();

MakcStudio:
I need to boost about 1000 accounts from one PC at the same time

yaakov-h:
This sounds incredibly dodgy...

We have #541 tracking the ability to specify a proxy server for WebAPI requests, but nothing for the CM connection itself. It’s unlikely that we would do that, without a compelling use-case.

MakcStudio:
will I be able to do it myself in a dirtier way? I just need to specify a proxy server before authorizing each account.

MakcStudio:
The problem occurs when you authorize more than 200 accounts in a row, so I need a proxy connection. If you can help to solve this problem, I will be very grateful

azuisleet:
Proxy support is probably possible for SOCKS and HTTP proxies depending on the protocol. The implementation would require modifying the TCP/UDP connections and the WebSocket connection. The easiest implementation is likely the WebSocket with a standard HTTP proxy, since it is supported through the library itself.

I don't believe proxy support is a high priority at the moment since there isn't a need to defeat IP checks or rate limits for standard use cases.

MakcStudio:
And I think that the problem is priority. For now steam is very active began to ban accounts for everything

MakcStudio:
If it is not difficult for you, could you tell me in which files I can see the Declaration of these variables? I need work authorization and clien.send();

PR #717, opened 2019-09-09
Proxy support for WebSocket connections

Vital7:
Adds proxy support for WebSocket connections

### DIFF
--- a/SteamKit2/SteamKit2/Networking/Steam3/WebSocketConnection.cs
+++ b/SteamKit2/SteamKit2/Networking/Steam3/WebSocketConnection.cs
@@ -18,9 +18,11 @@ namespace SteamKit2
 
         public ProtocolTypes ProtocolTypes => ProtocolTypes.WebSocket;
 
-        public void Connect(EndPoint endPoint, int timeout = 5000)
+        public void Connect(EndPoint endPoint, int timeout = 5000) => Connect(endPoint, null, timeout);
+
+        public void Connect(EndPoint endPoint, IWebProxy proxy, int timeout = 5000)
         {
-            var newContext = new WebSocketContext(this, endPoint);
+            var newContext = new WebSocketContext(this, endPoint, proxy);
             var oldContext = Interlocked.Exchange(ref currentContext, newContext);
             if (oldContext != null)
             {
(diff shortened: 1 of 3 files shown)


JustArchi:
You most likely want to also make it possible to supply `IWebProxy` to `SteamConfiguration` so this can be used without manual connections.

yaakov-h:
https://github.com/SteamRE/SteamKit/issues/587#issuecomment-412677406

JustArchi:
Enabling access to "our" Steam for users in China would be one compelling use case for this.

yaakov-h:
If Steam global gets blocked by China, an itty bitty SOCKS proxy is probably not going to hold up against the GFW. You would need a lower-level VPN.

Regardless, any solution which works for Steam today or in future would also work for a SteamKit2-based codebase, unless Steam itself adds proxy support.

MakcStudio:
I do not really understand, you did not add a ready-made solution for proxy support to the master branch? If I want to use this solution, do I need to compile my own version of the library?

Vital7:
@MakcStudio Yes, this PR wasn't accepted, so you have to compile this branch manually.

fjch1997:
> If Steam global gets blocked by China, an itty bitty SOCKS proxy is probably not going to hold up against the GFW. You would need a lower-level VPN.
> 
> Regardless, any solution which works for Steam today or in future would also work for a SteamKit2-based codebase, unless Steam itself adds proxy support.

Many people in China are using Shadowsocks which encrypts SOCKS5 traffic between the client and server. Also, WebSocket proxy goes throughh HTTPS and would be encrypted as well. The GFW has no way of seeing that it is trying to access Steam

This is useful and quite a simple update. I am not sure why it can't be merged.

fazelukario:
@yaakov-h As @fjch1997 mentioned, it makes sense to add a proxy, and I think that to ignore such a possibility (given that this feature is optional) in a library that is designed for software solutions such as bots, etc. at least wrong, and at most egoistically and stupidly.

yaakov-h:
We're not ignoring it, we're selective about adding features that can be easily misused for harmful and abusive purposes. If you have a compelling, legitimate, and practical use-case, then I'm all ears. If something is theoretically useful only in future, but practically abusable now, then I'll wait for that future.

PR #718, opened 2019-09-09
Support for providing custom machine ID on login

Vital7:
Adds a feature that allows user to provide own machine ID

### DIFF
--- a/SteamKit2/SteamKit2/Steam/Handlers/SteamUser/SteamUser.cs
+++ b/SteamKit2/SteamKit2/Steam/Handlers/SteamUser/SteamUser.cs
@@ -92,6 +92,14 @@ namespace SteamKit2
             public uint AccountID { get; set; }
 
             /// <summary>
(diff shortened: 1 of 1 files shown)


VoiDeD:
Not entirely sure that this is something we want to support. Outside of the potential for abuse this enables - there's a structure to the machine id that'd be lost if we allow anybody to supply arbitrary data.

JustArchi:
There is also an advantage of consistency in case one would like from his bot to always announce with fixed machineID instead of whatever machine it's running on. I'd definitely be interested in that for my private (non-public) projects that have a single instance.

VoiDeD:
It's fairly obvious that it wouldn't be difficult to edit a few lines of code. However, our intent isn't to enable, condone, or support this.

yaakov-h:
Whilst we can't really publish any code without making it _possible_ to screw with Valve's security features, we definitely don't intend to make it _easy_.

Issue #719, opened 2019-09-12
Make MachineInfoProvider class public


ezhevita:
Making MachineInfoProvider public will allow users to implement own info providers, which may be the better case for some systems.

PR #720, opened 2019-09-23
Add ability to provide custom MachineInfoProvider

Vital7:
Allows user to provide own MachineInfoProvider in order to provide correct machine ID in case if SteamKit default ones aren't useful in their environment.
Implements #719


yaakov-h:
This is still dodgy at hell, but it does have some legitimate uses (e.g. Xamarin).

A few notes:

- For a consumer implementation, it should be an interface implementation, not an abstract class. We should have an `IMachineInfoProvider` that consumers implement.
- If it's going to be part of SteamConfiguration, it has to stay local to that configuration or any SteamClients instantiated from it. Using statics with things like `HardwareUtils.Init` breaks that isolation and different values could end up cross-polluting other instances.
- Since it's supposed to be **hardware**, I'm tempted to not make it part of SteamConfiguration and just make it a static singleton that can be user-replaced, though it would have to be set before any other SK2 operation is triggered and not allowed to be re-set. Otherwise we could land up with race conditions.

SteamKit2/SteamKit2/Util/HardwareUtils.cs
```diff
@@ -281,12 +272,63 @@ public override byte[] GetDiskId()
...
+    public static class HardwareUtils
     {
+        private static IMachineInfoProvider userMachineInfoProvider;
+        private static bool IsMachineInfoProviderInitialized;
+
+        /// <summary>
+        /// User-provided MachineInfoProvider.
+        /// </summary>
+        /// <exception cref="InvalidOperationException">Occurs when a user tries to (re-)set this property after its first usage.</exception>
+        public static IMachineInfoProvider UserMachineInfoProvider
```
yaakov-h: This is static but not thread-safe.

SteamKit2/SteamKit2/Util/HardwareUtils.cs
```diff
@@ -281,12 +272,63 @@ public override byte[] GetDiskId()
...
+        {
+            get => userMachineInfoProvider;
```
yaakov-h: Why do we need a property for a "user-provided machine info provider", rather than just having a property for a "machine info provider"? It seems odd that a user can't just get the provider themselves.

SteamKit2/SteamKit2/Util/HardwareUtils.cs
```diff
@@ -281,12 +272,63 @@ public override byte[] GetDiskId()
...
+            set
+            {
+                if ( IsMachineInfoProviderInitialized )
+                {
+                    throw new InvalidOperationException(nameof(UserMachineInfoProvider) + " can't be re-set.");
```
yaakov-h: This message could confuse a user if they never set it in the first place, but `GetProvider()` was called first. This needs a revised API design.

SteamKit2/SteamKit2/Util/HardwareUtils.cs
```diff
@@ -309,22 +351,22 @@ public void Set3B3( string value )
...
         }
-
-
+        
         static Task<MachineID>? generateTask;
-
-
-        public static void Init()
+        
+        internal static void Init(IMachineInfoProvider provider)
```
yaakov-h: If a user sets a provider, we should kick off the generator task then, not wait for a later point.

SteamKit2/SteamKit2/Util/HardwareUtils.cs
```diff
@@ -309,22 +351,22 @@ public void Set3B3( string value )
                 this.KeyValues["3B3"].Value = value;
             }
 
+            
+            // 333 is some sort of user supplied data and is currently unused
```
yaakov-h: Why duplicate this comment from below? This doesn't seem to add any further value.

SteamKit2/SteamKit2/Util/HardwareUtils.cs
```diff
@@ -281,12 +273,58 @@ public override byte[] GetDiskId()
...
+        {
+            get => machineInfoProvider;
+            private set
+            {
+                if ( isMachineInfoProviderInitialized )
+                {
+                    throw new InvalidOperationException(nameof(MachineInfoProvider) + " can't be (re-)set after its initialization.");
+                }
+
+                Interlocked.Exchange( ref machineInfoProvider, value );
```
yaakov-h: Not sure what this is doing

SteamKit2/SteamKit2/Util/HardwareUtils.cs
```diff
@@ -281,12 +273,58 @@ public override byte[] GetDiskId()
...
+        /// <summary>
+        /// MachineInfoProvider used for this device.
+        /// </summary>
+        /// <exception cref="InvalidOperationException">Occurs when a user tries to (re-)set this property after its first usage.</exception>
+        public static IMachineInfoProvider? MachineInfoProvider
+        {
+            get => machineInfoProvider;
+            private set
+            {
+                if ( isMachineInfoProviderInitialized )
```
yaakov-h: This seems still prone to race conditions.

SteamKit2/SteamKit2/Util/HardwareUtils.cs
```diff
@@ -340,17 +400,19 @@ public static void Init()
                 return null;
             }
 
+            if ( !isMachineInfoProviderInitialized )
```
yaakov-h: This also seems subject to race conditions. Why do we mark it as initialized here, long after the actual initialization?

SteamKit2/Tests/HardwareUtilsFacts.cs
```diff
@@ -0,0 +1,65 @@
...
+
+            public byte[] GetDiskId() => new byte[ 32 ];
+        }
+
+        [ Fact ]
+        public void GetOSProviderReturnsDefaultForOSMachineInfoProvider()
+        {
+            IMachineInfoProvider machineInfoProvider = HardwareUtils.GetOSProvider();
+            Assert.NotNull( machineInfoProvider );
+            switch ( Environment.OSVersion.Platform )
```
yaakov-h: 🤨

SteamKit2/Tests/HardwareUtilsFacts.cs
```diff
@@ -0,0 +1,65 @@
...
+        public class CustomMachineInfoProvider : IMachineInfoProvider
+        {
+            public byte[] GetMachineGuid() => new byte[ 32 ];
+
+            public byte[] GetMacAddress() => new byte[ 32 ];
+
+            public byte[] GetDiskId() => new byte[ 32 ];
+        }
+
+        [ Fact ]
```
yaakov-h: Whitespace in attribute seems off

SteamKit2/Tests/HardwareUtilsFacts.cs
```diff
@@ -0,0 +1,65 @@
...
+        public void InitSetsDefaultMachineInfoProvider()
+        {
+            IMachineInfoProvider machineInfoProvider = HardwareUtils.GetOSProvider();
+            HardwareUtils.Init( machineInfoProvider );
+            Assert.NotNull( HardwareUtils.MachineInfoProvider );
+            Assert.IsAssignableFrom<DefaultInfoProvider>( HardwareUtils.MachineInfoProvider );
+        }
+
```
yaakov-h: excessive blank lines here

SteamKit2/SteamKit2/Steam/SteamClient/SteamClient.cs
```diff
@@ -72,6 +72,11 @@ public SteamClient( SteamConfiguration configuration )
             this.AddHandler( new SteamMatchmaking() );
             this.AddHandler( new SteamNetworking() );
 
+            if ( HardwareUtils.MachineInfoProvider == null )
```
yaakov-h: Leaky abstraction.

Abrynos:
Any update on this one?

yaakov-h:
The PR as it currently stands still has some thread unsafety issues. I'll revisit this myself when I have some time.

yaakov-h:
Closing in favour of #1028.

PR #721, opened 2019-09-24
Adopt C# 8.0 Nullable Reference Types

yaakov-h:
It looks like there are actually a few code paths whereby we can encounter NullReferenceExceptions.

This PR is to annotate code and handle the remaining compiler warnings to make our code safer.

This PR:

- Requires .NET Core 3.0 SDK and/or Visual Studio 2019 v16.3 to build.
- Adds nullability annotations as per C# 8.0 spec. I think I got them all right, I was mostly just following the compiler warnings.
- Removes SonarScanner which was consuming 85% of our build time.
- Updates NetHook2 to use the newer VC++ build toolchain (VS2019)
- Fixes a safety/deprecation warning in NetHook2


.travis.yml
```diff
@@ -2,33 +2,30 @@ dist: xenial
...
 
-addons:
-    sonarcloud:
-        organization: "steamre"
-
 solution: SteamKit2/SteamKit2.sln
 
+before_install:
+    - sudo apt-get update
+    - sudo apt-get install -y dotnet-runtime-2.1
```
xPaw: You changed to `dotnet: 3.0`, but install 2.1 here?

yaakov-h: .NET SDK 3.0 w/ 3.0 runtime.

We still need a .NET 2.x runtime to run our tests which target `netcoreapp2.0`, and we did need it to run SonarScanner which targets `netcoreapp2.1`.

By default, .NET apps don't roll forward on major versions to a different runtime.

PR #722, opened 2019-10-01
Added getting the CellID in the constructor method.

### LINKED ISSUE #690: SteamConfiguration.CellID is not used in login
`SteamUser.LogOn` and `LogOnAnonymous` accept details object which can have a cellid set, I think it would be a good idea to set it by default to `SteamConfiguration.CellID` if it's not provided in `LogOnDetails`.

ADustyOldMuffin:
closes #690 

This is my first contribution so I'm still getting used to the project so any constructive criticism is welcome!

All I did was instantiate a configuration in the `LogOnDetails` and `AnonymousLogOnDetails` constructors and grab the CellID and set it. Let me know if I should've set the default value elsewhere or if there's a better way to do this.

### DIFF
--- a/SteamKit2/SteamKit2/Steam/Handlers/SteamUser/SteamUser.cs
+++ b/SteamKit2/SteamKit2/Steam/Handlers/SteamUser/SteamUser.cs
@@ -119,9 +119,6 @@ namespace SteamKit2
                 AccountInstance = SteamID.DesktopInstance; // use the default pc steam instance
                 AccountID = 0;
 
-                var configuration = SteamConfiguration.Create(b => b.WithProtocolTypes(ProtocolTypes.All));
-                CellID = configuration.CellID;
-
                 ClientOSType = Utils.GetOSType();
                 ClientLanguage = "english";
             }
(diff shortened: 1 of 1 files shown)


yaakov-h:
Creating a new configuration and then immediately using it's values is fairly pointless, as it doesn't give the consumer of the library any chance to configure it.

Inside SteamUser, where we use the values from `LogOnDetails` and `AnonymousLogOnDetails`, we have access to the configuration object that the consumer has already configured through `this.Client.Configuration`.

A better approach would be where we use the values at this line:

```cs
logon.Body.cell_id = details.CellID;
```

to either check for the magic value zero and then use the configuration instead, or to make the value on the `xxxDetails` objects nullable and then perform a null-coalesce like so:

```cs
logon.Body.cell_id = details.CellID ?? Client.Configuration.CellID;
```

ADustyOldMuffin:
I figured it was simpler that what I was doing to get the configuration, embarrassed I missed where to access it. Thanks for the help!

yaakov-h:
Thanks!

Issue #503, opened 2017-12-18
Make DebugLog more SteamClient-personal


JustArchi:
As part of https://github.com/SteamRE/SteamKit/issues/260#issuecomment-352566952 I noticed that `DebugLog` is always static and prints information about any amount of clients taking part in the process.

Sometimes this might be confusing. An example situation would be when we're debugging one particular `SteamClient` and other `SteamClients` working in the background are generating at best useless, and at worst misleading messages that are not related to our main `SteamClient` in any way. Of course you could say "then just use one SteamClient while reproducing the problem", but it's not always possible or a proper solution, since sometimes bugs are triggered exactly when more than one `SteamClient` is operating.

Therefore my suggestion is to somehow include bot identifier in output messages, so we could make use of it for identifying to which `SteamClient` given message should be connected to.

There are several possibilities to achieve this, the way that would be most native for my ASF would be separating general "static" `DebugLog` from `SteamClient`-specific messages that are now in "another" `DebugLog` specific to given `SteamClient`. This way we can be subscribed to any amount of debug log generators, mainly debug logs of specific bots, as well as existing "static" one.

If that's too much work (e.g. making `SteamClient`-specific `DebugLog` is unwanted), we could at least add and make use of an identifier in `SteamClient` and print it in the message. This could be a simple `string` in `SteamClient` object. This way we could map problem of bots ourselves.

```
2017-12-18 21:27:13|ArchiSteamFarm-27083|INFO|nebrassy|Start() Starting...
2017-12-18 21:27:13|ArchiSteamFarm-27083|DEBUG|ASF|WriteLine() CMClient | nebrassy | Initialized new SteamClient with ID nebrassy
2017-12-18 21:27:14|ArchiSteamFarm-27083|INFO|Pyro|Start() Starting...
2017-12-18 21:27:14|ArchiSteamFarm-27083|DEBUG|ASF|WriteLine() CMClient | Pyro | Initialized new SteamClient with ID Pyro
2017-12-18 21:27:14|ArchiSteamFarm-27083|INFO|Scout|Start() Starting...
2017-12-18 21:27:14|ArchiSteamFarm-27083|DEBUG|ASF|WriteLine() CMClient | Scout | Initialized new SteamClient with ID Scout
```

Of course those are only two suggestions, feel free to suggest other ones.

yaakov-h:
[...]

PR #723, opened 2019-10-02
Add a per-SteamClient unique identifier message to DebugLog messages

yaakov-h:
One approach to solving https://github.com/SteamRE/SteamKit/issues/503.

Thoughts?

Personally I don't like having to pass it everywhere - though I think that I would rather pass just a unique ID around than a whole `ILogger` object or similar.

I have considered storing it in an `AsyncLocal<T>` but that could get messy quite quickly.

(cc @JustArchi)

### DIFF
--- a/SteamKit2/SteamKit2/Networking/Steam3/NetFilterEncryption.cs
+++ b/SteamKit2/SteamKit2/Networking/Steam3/NetFilterEncryption.cs
@@ -12,12 +12,14 @@ namespace SteamKit2
     class NetFilterEncryption : INetFilterEncryption
     {
         readonly byte[] sessionKey;
+        readonly LoggerToken loggerToken;
 
-        public NetFilterEncryption( byte[] sessionKey )
+        public NetFilterEncryption( byte[] sessionKey ,LoggerToken loggerToken )
         {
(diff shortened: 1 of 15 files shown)


SteamKit2/SteamKit2/Networking/Steam3/NetFilterEncryption.cs
```diff
@@ -12,12 +12,14 @@ namespace SteamKit2
     class NetFilterEncryption : INetFilterEncryption
     {
         readonly byte[] sessionKey;
+        readonly LoggerToken loggerToken;
 
-        public NetFilterEncryption( byte[] sessionKey )
+        public NetFilterEncryption( byte[] sessionKey ,LoggerToken loggerToken )
```
xPaw:
```suggestion
        public NetFilterEncryption( byte[] sessionKey, LoggerToken loggerToken )
```

xPaw:
This is a breaking change for anyone that implements DebugLog, right?

Also, perhaps CMClient should accept a logger token guid so consumers can set one consistently if need be (same guid between restarts).

yaakov-h:
As it is now, this is a breaking change.

If we want to keep compatibility, we could keep the old methods as overloads and wrap them, version `IDebugLogListener2`, even use DIM on .NET Core 3.0 (but keep a breaking change on other platforms).

yaakov-h:
Abandoning in favour of #777.

PR #726, opened 2019-10-02
Fix CA2101: Specify marshaling for P/Invoke string arguments

xPaw:
https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2101-specify-marshaling-for-p-invoke-string-arguments?view=vs-2019

### DIFF
--- a/SteamKit2/SteamKit2/Util/MacHelpers.cs
+++ b/SteamKit2/SteamKit2/Util/MacHelpers.cs
@@ -4,6 +4,7 @@ using System.Text;
 
 namespace SteamKit2.Util.MacHelpers
 {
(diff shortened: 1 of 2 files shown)


SteamKit2/SteamKit2/Util/MacHelpers.cs
```diff
@@ -71,7 +71,7 @@ static class LibC
     {
         const string LibraryName = "libc";
 
-        [DllImport(LibraryName, CallingConvention = CallingConvention.Cdecl)]
```
yaakov-h: All the APIs in this file deal with regular UTF-8 strings (`char *`) and need to be `CharSet.Ansi`. With `CharSet.Unicode`, SK2 just crashes.

PR #728, opened 2019-10-03
Add LICENSE file, update licencing details

yaakov-h:
This incorporates and supercedes #669, following https://github.com/SteamRE/SteamKit/pull/669#issuecomment-495134608 since nobody had any further comments.


SteamKit2/SteamKit2/SteamKit2.csproj
```diff
@@ -44,14 +42,6 @@
...
     </Content>
-    <Content Include="gpl.txt">
-        <Pack>true</Pack>
-        <PackagePath />
-    </Content>
-    <Content Include="lgpl.txt">
-        <Pack>true</Pack>
-        <PackagePath />
-    </Content>
     <Content Include="license.txt">
```
VoiDeD: We also have this license file, dunno if we want to drop this.

yaakov-h: I kept it for now because it also includes the copyright line. We don't have that anywhere else in the package for now.

PR #734, opened 2019-10-07
Fixing build warnings

### LINKED ISSUE #640: Fix SourceLink build warning
> Microsoft.Build.Tasks.Git.targets(36,5): warning : Submodule 'Resources/Protobufs' doesn't have any commit, the source code won't be available via source link. 

Warning is because submodules are not checked out.

We should run `git submodule update --init --recursive` in the build. For AppVeyor, this should be during the `install` section.

ADustyOldMuffin:
fixes #640 

I've added the suggested changes where I believe it was suggested to add them hoping to fix the build warning.


yaakov-h:
Looks good. Thanks!

Issue #758, opened 2019-10-19
Use the new ip fields in the logon message (IPv6 support)


xPaw:
Now that protogen has been updated, new protos are included.

Valve have been adding ipv6 support in the steam client, so there's new proto fields for it.

`CMsgClientLogon` added `.CMsgIPAddress obfuscated_private_ip` and `.CMsgIPAddress public_ip` fields. Same with `CMsgClientLogonResponse`.

CMsgIPAddress is this:
```
message CMsgIPAddress {
	oneof ip {
		fixed32 v4 = 1;
		bytes v6 = 2;
	}
}
```

See https://github.com/SteamDatabase/Protobufs/commit/a4ba0a9e35a204b8185ca32a8aec697c830cadb2#diff-ae4e2a3dc1bbf60e70ebca0ab4507c8a

See https://github.com/SteamDatabase/Protobufs/search?q=CMsgIPAddress&unscoped_q=CMsgIPAddress for all the messages that use v4/v6 oneof.

xPaw:
See https://github.com/SteamRE/SteamKit/pull/759/commits/7ab4e1a64c1ac573363bf015d8a96b5fdbe6a593 for what code actually needs to be updated in SK2.

yaakov-h:
Should we do something about LoginID, and if so, what?

xPaw:
> Should we do something about LoginID, and if so, what?

Not sure I understand, what's up with LoginID?

yaakov-h:
https://github.com/SteamRE/SteamKit/blob/d7b13702319b1dcd0997772551de6028030c83e3/SteamKit2/SteamKit2/Steam/Handlers/SteamUser/SteamUser.cs#L40-L50

It allows the user to override the "obfustucated" IPv4 address.

JustArchi:
The best way to determine if there is any action needed is to try to connect 2 SteamClients from the same machine (best if the same process).

If you don't get LogonSessionReplaced or something like that during connect of second account, it's no longer needed.

If you do, and setting LoginID of the second session to some arbitrary number resolves the issue, it's still needed.

xPaw:
That's done now.

PR #759, opened 2019-10-19
Update protobufs

xPaw:
- Updated ip field names to use deprecated ones. See #758
- Legacy servicemethod
- `CMsgClientServerList` was removed - https://github.com/SteamDatabase/Protobufs/commit/a4ba0a9e35a204b8185ca32a8aec697c830cadb2#diff-0f28215af4ff6780216ed68eec25ff52


yaakov-h:
So server-list-via-CM is dead now?

Resources/NetHookAnalyzer2/NetHookAnalyzer2/Specializations/CSGO/CSGOSOMultipleObjectsGCSpecialization.cs
```diff
@@ -3,7 +3,8 @@
 using System.IO;
 using ProtoBuf;
 using ProtoBuf.Meta;
-using SteamKit2.GC.CSGO.Internal;
+//using SteamKit2.GC.CSGO.Internal;
+using SteamKit2.GC.Dota.Internal; // CSGO no longer has objects_added and objects_removed
```
yaakov-h: if CS:GO no longer has those fields, should we just delete the code in this class that handles them? Or is data still there but it just has no field in the protos?

xPaw:
I removed `GetServersOfType` which means `CDNClient.FetchServerListAsync` also had to be removed. That method is trivially replaced by `ContentServerDirectoryService.LoadAsync`, like depot downloader already did: https://github.com/SteamRE/DepotDownloader/commit/c6975323d2915a451119260ded45032a47e4fa47

`UFSClient` also used the server list, I changed it to require a server for now.

Resources/NetHookAnalyzer2/NetHookAnalyzer2/Specializations/CSGO/CSGOSOMultipleObjectsGCSpecialization.cs
yaakov-h: Last issue outstanding.

PR #765, opened 2019-10-28
NetHook2 emsg descriptors 'fix' for steam client beta

m4dEngi:
It's still possible to get descriptors after msglist initialization.

### DIFF
--- a/Resources/NetHook2/NetHook2/crypto.cpp
+++ b/Resources/NetHook2/NetHook2/crypto.cpp
@@ -11,18 +11,16 @@
 
 
 SymmetricEncryptChosenIVFn Encrypt_Orig = nullptr;
-bool (__cdecl *GetMessageFn)( int * ) = nullptr;
-
+PchMsgNameFromEMsgFn PchMsgNameFromEMsg = nullptr;
 
 static_assert(sizeof(MsgInfo_t) == 20, "Wrong size of MsgInfo_t");
 static_assert(offsetof(MsgInfo_t, eMsg) == 0, "Wrong offset of MsgInfo_t::eMsg");
-static_assert(offsetof(MsgInfo_t, pchMsgName) == 4, "Wrong offset of MsgInfo_t::pchMsgName");
-static_assert(offsetof(MsgInfo_t, nFlags) == 8, "Wrong offset of MsgInfo_t::nFlags");
-static_assert(offsetof(MsgInfo_t, k_EServerTarget) == 12, "Wrong offset of MsgInfo_t::k_EServerTarget");
-static_assert(offsetof(MsgInfo_t, nUnk1) == 16, "Wrong offset of MsgInfo_t::uUnk1");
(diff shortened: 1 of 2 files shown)


m4dEngi:
It's still a bad piece of code, because we can no longer tell the correct size of emsg list. Maybe it'll be a better idea to use CMessageList::FindMessageInfoForEMsg instead and get names using returned MsgInfo_t structs for incoming/outgoing packets?

WhiteRabbitHS:
Valve does not help
Hello,
I think I got a VAC ban as a result of a mistake.
I request that my account be reviewed.
Please help.
I swear I didn't use any software.
Please listen to me sincerely.
It's a real shame to take the ban on vac.
I never deserved this.
I'm 33, and I've never cheated.
I've been playing CS since 1999.
I always trust my talent.
My computer and internet are bad.
So my rank is low.
And I'm used to it.
You thought any software on my computer was cheating.
I appreciate you fighting the trick.
But it's not fair for me to be punished.
I'm an honest player.
He's an honest salesman.
I bought and sold more than 4000 skins.
I have a beautiful profile.
Good game history.
While CS GO is free.
Why would I risk my own account?
There are 2000 TL skins in the market.
I'm so sorry I didn't play Cs go.
Please help, I'm never the tricks.
Thank you for your interest
Yours truly Ferhat Daşdemir "White Rabbit"


https://steamcommunity.com/id/WhiteRabbitHS/
https://steamcommunity.com/id/WhiteRabbitHS/allcomments
https://steamcommunity.com/id/WhiteRabbitHS/inventory/
https://steamcommunity.com/groups/whiterabbitheadshot
https://www.youtube.com/channel/UCJTStjc8vYemsOjlRL_vmRA?view_as=subscriber
https://www.youtube.com/watch?v=0gowikkuJFU&feature=emb_title

yaakov-h:
Confirmed as working.

Interestingly, instead of `(null)` being the placeholder for "I don't know what this EMsg is", it's now `Unknown`.

Is there anything we can do to preserve `emsg_list.txt`/`emsg_list_detailed.txt`, or do we basically lose the ability to keep track of all EMsgs?

WhiteRabbitHS:
Bu nedir?

6 Mar 2020 Cum 07:12 tarihinde Yaakov <notifications@github.com> şunu yazdı:

m4dEngi:
> 
> Is there anything we can do to preserve `emsg_list.txt`/`emsg_list_detailed.txt`, or do we basically lose the ability to keep track of all EMsgs?

It looks like they changed emsglist class internals. EMsgs are now dynamically pushed on vector. Old EMsg structs are still there, but i'm not sure that whole list can be safely read sequentially, because its size is unknown( on windows list populated in a while loop until it hits record with emsgid 0xFFFFFFFF, on osx it's the same while loop but until it hits next symbol ) and string descriptors are not populated before constructor is called.

The reasons why i switched to using PchMsgNameFromEMsg 
- It's easier to find and it's signature is more or less stable.
- It's guaranteed to call msglist constructor.
- It's easier than re-implementing everything to do lookup properly using msglist interface while emsg list dumps are available @steamdb.

PR #769, opened 2019-10-30
Use UriBuilder in CDNClient

xPaw:
Also removes `DoCommandAsync` as that wasn't used anymore.

I haven't tested this though.


yaakov-h:
Does this change anything, or is it purely a refactor?

xPaw:
Just a refactor.

yaakov-h:
I thought so, just checking.

PR #770, opened 2019-10-30
Implement CMsgIPAddress

xPaw:
- [x] Obfuscation for ipv6 is implemented
- [x] `NetHelpers.TryParseIPEndPoint` doesn't support IPv6 bracket notation
- [ ] `LoginID` in logon just sets v4 - i don't think this a big of an issue
- [ ] `CMsgGSServerType` has `deprecated_game_ip_address`, but there are no other fields for IP

I can't test any of this because there's no way of getting ipv6 cm servers anywhere yet.

### DIFF
--- a/SteamKit2/SteamKit2/Util/Utils.cs
+++ b/SteamKit2/SteamKit2/Util/Utils.cs
@@ -402,8 +402,26 @@ namespace SteamKit2
 
             if ( ipAddr.AddressFamily == AddressFamily.InterNetworkV6 )
             {
-                // TODO: Implement ipv6 obfuscation, it uses MsgClientLogon.ObfuscationMask too
-                // See ObfuscatePrivateIP in steamclient
+                // TODO: See ObfuscatePrivateIP in steamclient, unsure if applied correctly
+                localIp.v6[ 0 ] ^= 0xBA;
+                localIp.v6[ 1 ] ^= 0xAD;
+                localIp.v6[ 2 ] ^= 0xF0;
+                localIp.v6[ 3 ] ^= 0x0D;
+
+                localIp.v6[ 4 ] ^= 0xBA;
+                localIp.v6[ 5 ] ^= 0xAD;
+                localIp.v6[ 6 ] ^= 0xF0;
+                localIp.v6[ 7 ] ^= 0x0D;
+
+                localIp.v6[ 8 ] ^= 0xBA;
+                localIp.v6[ 9 ] ^= 0xAD;
+                localIp.v6[ 10 ] ^= 0xF0;
+                localIp.v6[ 11 ] ^= 0x0D;
+
+                localIp.v6[ 12 ] ^= 0xBA;
+                localIp.v6[ 13 ] ^= 0xAD;
+                localIp.v6[ 14 ] ^= 0xF0;
+                localIp.v6[ 15 ] ^= 0x0D;
             }
             else
             {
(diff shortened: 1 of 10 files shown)


xPaw:
If we bump SK2 to `netstandard2`, then `NetHelpers.TryParseIPEndPoint` can be replaced by `IPEndPoint.TryParse`.

SteamKit2/SteamKit2/Util/Utils.cs
```diff
@@ -363,42 +365,95 @@ public static uint GetIPAddress( IPAddress ipAddr )
             return BitConverter.ToUInt32( addrBytes, 0 );
         }
 
-
-        public static uint EndianSwap( uint input )
+        public static IPAddress GetIPAddress( CMsgIPAddress ipAddr )
```
yaakov-h, line 368: Can we put this in `SteamKit2.Internal` alongside `CMsgIPAddress`? It seems odd having this function cross-namespace.

Also, IMO I think this makes more sense as an extension method.

SteamKit2/SteamKit2/Util/Utils.cs
```diff
@@ -363,42 +365,95 @@ public static uint GetIPAddress( IPAddress ipAddr )
...
+
+                msgIpAddress.v4 = BitConverter.ToUInt32( addrBytes, 0 );
+            }
+
+            return msgIpAddress;
         }
-        public static ushort EndianSwap( ushort input )
+
+        public static CMsgIPAddress ObfuscatePrivateIP( IPAddress ipAddr )
         {
```
yaakov-h, line 400: I'd rather have two methods - one to convert `IPAddress` into `CMsgIPAddress`, and one to apply obfuscation to `CMsgIPAddress`

SteamKit2/SteamKit2/Util/Utils.cs
```diff
@@ -363,42 +365,95 @@ public static uint GetIPAddress( IPAddress ipAddr )
...
-            return ( ushort )IPAddress.NetworkToHostOrder( ( short )input );
+            var localIp = GetMsgIPAddress( ipAddr );
+
+            if ( ipAddr.AddressFamily == AddressFamily.InterNetworkV6 )
+            {
+                // TODO: See ObfuscatePrivateIP in steamclient, unsure if applied correctly
```
yaakov-h, line 405: This looks right to me unless there's something funny about byte order that I've overlooked...

yaakov-h: So using the following sample code, to try mirror the function from steamclient:

```c++
int main()
{
	int8_t data[16];
	memset(data, 0, sizeof(data));

	auto i = 0;
	int32_t* ptr = (int32_t*)data;
	do
	{
		ptr[i] = ptr[i] ^ 0xBADF00D;

		i++;
	} while (i < 4);

	for (auto i = 0; i < sizeof(data); i++)
	{
		std::cout << data << std::endl;
	}
	return 0;
}
```

It looks like we actually need to XOR in reverse, i.e. `0x0D, 0xF0, 0xAD, 0xBA` as the mask. You're using the high byte to obfuscate the low byte and vice-versa.

yaakov-h:
My guess on `CMsgGSServerType` is that they're implementing IPv6 Steam infrastructure before implementing IPv6 game server support.

SteamKit2/SteamKit2/Util/Utils.cs
xPaw: It won't be in `Internal` if protobufs update is merged.

xPaw:
I added method extensions onto `CMsgIPAddress`, didn't do it on `IPAddress`.

PR #771, opened 2019-10-31
Render CMsgIPAddress in NHA

xPaw:
`CMsgIPAddress` is `oneof` so I'm just rendering whichever value is set.

### DIFF
--- a/Resources/NetHookAnalyzer2/NetHookAnalyzer2/TreeNodeObjectExplorer.cs
+++ b/Resources/NetHookAnalyzer2/NetHookAnalyzer2/TreeNodeObjectExplorer.cs
@@ -567,6 +567,23 @@ namespace NetHookAnalyzer2
 					SetValueForDisplay(string.Format("\"{0}\"", kv.Value), kv.Value);
 				}
 			}
+			else if ( value is CMsgIPAddress msgIpAddr )
+			{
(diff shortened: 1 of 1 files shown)


Resources/NetHookAnalyzer2/NetHookAnalyzer2/TreeNodeObjectExplorer.cs
```diff
@@ -567,6 +567,23 @@ void Initialize()
 					SetValueForDisplay(string.Format("\"{0}\"", kv.Value), kv.Value);
 				}
 			}
+			else if ( value is CMsgIPAddress msgIpAddr )
+			{
+				if ( msgIpAddr.ShouldSerializev4() )
+				{
+					byte[] addrBytes = BitConverter.GetBytes( msgIpAddr.v4 );
+					Array.Reverse( addrBytes );
```
yaakov-h: https://docs.microsoft.com/en-us/dotnet/api/system.net.ipaddress.networktohostorder?view=netframework-4.8#System_Net_IPAddress_NetworkToHostOrder_System_Int32_ ?

xPaw: All other SK2 code does it this way, so I kept it.

PR #774, opened 2019-11-02
Remove sent/recv debug messages

xPaw:
DebugNetworkListener can be used for this purpose.

This will actually make DebugLog easier to enable without having to worry about the packet spam.


yaakov-h:
I find these useful. 😢 

Can we _move_ it to DebugNetworkListener?

xPaw:
I added them to `NetHookNetworkListener`, is that good enough?

yaakov-h:
I think so. That should be the EMsg without the high-bit for proto.

xPaw:
It practically logs in the same place as the logs i removed (except for the fact that multis will be split).

PR #777, opened 2019-11-03
Add a per-SteamClient unique identifier message to DebugLog messages (2)

yaakov-h:
A different approach to solving #503.

Thoughts?

This version should not trigger breaking changes in consumers.

cc/ @JustArchi @xPaw

### DIFF
--- a/SteamKit2/SteamKit2/Steam/CMClient.cs
+++ b/SteamKit2/SteamKit2/Steam/CMClient.cs
@@ -23,12 +23,17 @@ namespace SteamKit2.Internal
     /// <summary>
     /// This base client handles the underlying connection to a CM server. This class should not be use directly, but through the <see cref="SteamClient"/> class.
     /// </summary>
-    public abstract class CMClient
+    public abstract class CMClient : ILogContext
     {
         /// <summary>
         /// The configuration for this client.
         /// </summary>
         public SteamConfiguration Configuration { get; }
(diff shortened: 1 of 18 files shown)


SteamKit2/SteamKit2/Steam/CMClient.cs
```diff
@@ -23,13 +23,18 @@ namespace SteamKit2.Internal
...
     {
         /// <summary>
         /// The configuration for this client.
         /// </summary>
         public SteamConfiguration Configuration { get; }
 
+        /// <summary>
+        /// A unique identifier for this client instance.
+        /// </summary>
+        public string ID { get; }
```
xPaw: Allow setter?

JustArchi: All fine but a setter would indeed be nice, see my reasoning in https://github.com/SteamRE/SteamKit/issues/503#issuecomment-549124002

yaakov-h: Should be immutable for the object’s lifetime, so perhaps a constructor parameter?

yaakov-h: See above.

xPaw: That works.

PR #779, opened 2019-11-04
Restore response status code in exception messages

RudeySH:
After updating my SteamKit dependency from 2.1.0 to 2.2.0, I noticed the exception messages from Web API requests were slightly different. Before 2.2.0, an example exception message looked like this:

> Response status code does not indicate success: 403 (Forbidden).

But now after updating to 2.2.0, the same message looks like this:

> Response status code does not indicate success: Forbidden (Forbidden).

This change was caused by #603 and is most likely unintentional. My pull requests restores the original exception messages.

### DIFF
--- a/SteamKit2/SteamKit2/Steam/CDNClient.cs
+++ b/SteamKit2/SteamKit2/Steam/CDNClient.cs
@@ -497,7 +497,7 @@ namespace SteamKit2
 
                     if ( !response.IsSuccessStatusCode )
                     {
-                        throw new SteamKitWebRequestException( $"Response status code does not indicate success: {response.StatusCode} ({response.ReasonPhrase}).", response );
+                        throw new SteamKitWebRequestException( $"Response status code does not indicate success: {(int)response.StatusCode} ({response.ReasonPhrase}).", response );
                     }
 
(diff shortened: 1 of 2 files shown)


SteamKit2/SteamKit2/Steam/CDNClient.cs
```diff
@@ -497,7 +497,7 @@ async Task<byte[]> DoRawCommandAsync( Server server, string command, string? arg
 
                     if ( !response.IsSuccessStatusCode )
                     {
-                        throw new SteamKitWebRequestException( $"Response status code does not indicate success: {response.StatusCode} ({response.ReasonPhrase}).", response );
```
yaakov-h: The more-correct fix here would be `{response.StatusCode:D}`, which calls into `Enum.ToString(string)` as per https://docs.microsoft.com/en-us/dotnet/standard/base-types/enumeration-format-strings?view=netframework-4.8

RudeySH: I wasn't familiar with the "D" format string. I've edited both files.

yaakov-h:
Thanks!

PR #787, opened 2019-11-07
Override duplicate keys in SteamApps handlers instead of throwing

xPaw:
Repro:
```c#
Steam.Apps.PICSGetAccessTokens(new List<uint>() { 440, 440 }, Enumerable.Empty<uint>());
```

This caused Steam to return same appid twice in the response, and `AppTokens.Add` would throw an exception.

### DIFF
--- a/SteamKit2/SteamKit2/Steam/Handlers/SteamApps/Callbacks.cs
+++ b/SteamKit2/SteamKit2/Steam/Handlers/SteamApps/Callbacks.cs
@@ -315,12 +315,12 @@ namespace SteamKit2
 
                 foreach ( var package_token in msg.package_access_tokens )
                 {
-                    PackageTokens.Add( package_token.packageid, package_token.access_token );
+                    PackageTokens[ package_token.packageid ] = package_token.access_token;
                 }
 
                 foreach ( var app_token in msg.app_access_tokens )
                 {
-                    AppTokens.Add( app_token.appid, app_token.access_token );
+                    AppTokens[ app_token.appid ] = app_token.access_token;
                 }
             }
         }
@@ -397,12 +397,12 @@ namespace SteamKit2
 
                 foreach ( var package_change in msg.package_changes )
                 {
-                    PackageChanges.Add( package_change.packageid, new PICSChangeData( package_change ) );
+                    PackageChanges[ package_change.packageid ] = new PICSChangeData( package_change );
                 }
 
                 foreach ( var app_change in msg.app_changes )
                 {
-                    AppChanges.Add( app_change.appid, new PICSChangeData( app_change ) );
+                    AppChanges[ app_change.appid ] = new PICSChangeData( app_change );
                 }
             }
         }
@@ -547,12 +547,12 @@ namespace SteamKit2
 
                 foreach ( var package_info in msg.packages )
                 {
-                    Packages.Add( package_info.packageid, new PICSProductInfo( package_info ) );
+                    Packages[ package_info.packageid ] = new PICSProductInfo( package_info );
                 }
 
                 foreach ( var app_info in msg.apps )
                 {
-                    Apps.Add( app_info.appid, new PICSProductInfo( msg, app_info ) );
+                    Apps[ app_info.appid ] = new PICSProductInfo( msg, app_info );
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


JustArchi:
I'm not confident this is the correct solution, if anything it should be some kind of name value collection if duplicates are allowed, but you definitely shouldn't override entry with given key using the last value spotted.

xPaw:
When do you except Valve to give different responses in PICS for the same app?

JustArchi:
Do you have a guarantee that PICS info won't change between one execution and another? Because I'm pretty confident Valve has no locks when somebody asks for that API.

It's irrelevant if the response is the same or not, if it's returned two times and it's expected scenario, we should put it in the name value collection. If we agree that asking for the same appID/packageID two times is unwanted and out of any valid use case, we should `ToHashSet()` on that input ienumerable and never get down to this situation to begin with.

Current code is correct and throws on invalid usage. Whether we should correct that usage is entirely different thing and not outright obvious in regards to the solution. If you ask me, the correct solution is to either throw on duplicates from input, or ignore them using `ToHashSet()` rather than fixing the dictionary.

JustArchi:
> Send garbage request in, get garbage out 🤷‍♂

Exactly, and to the best of my knowledge those SK2 wrappers exist for the sole reason of doing their best to forbid us from doing shit like that, otherwise you could just craft and parse protos yourself. If you use `SteamUser.LogOn()` with null username, you get an exception, not request sent with invalid data following the logic of "garbage in, garbage out". If you want to send that garbage request, you can always craft it yourself.

Following the same logic, this method should also throw on invalid input, or if we decide that invalid input is not irrecoverable from (since we can just ignore duplicates in this case and be fine), the input should be fixed, not the output.

SteamKit2/SteamKit2/Steam/Handlers/SteamApps/Callbacks.cs
```diff
@@ -315,12 +315,12 @@ internal PICSTokensCallback( JobID jobID, CMsgClientPICSAccessTokenResponse msg
...
                 foreach ( var package_token in msg.package_access_tokens )
                 {
-                    PackageTokens.Add( package_token.packageid, package_token.access_token );
+                    PackageTokens[ package_token.packageid ] = package_token.access_token;
                 }
 
                 foreach ( var app_token in msg.app_access_tokens )
                 {
-                    AppTokens.Add( app_token.appid, app_token.access_token );
+                    AppTokens[ app_token.appid ] app_token.access_token;
```
Ne3tCode: Missing `=`

SteamKit2/SteamKit2/Steam/Handlers/SteamApps/Callbacks.cs
```diff
@@ -397,12 +397,12 @@ internal PICSChangesCallback( JobID jobID, CMsgClientPICSChangesSinceResponse ms
...
                 foreach ( var package_change in msg.package_changes )
                 {
-                    PackageChanges.Add( package_change.packageid, new PICSChangeData( package_change ) );
+                    PackageChanges[ package_change.packageid ] = new PICSChangeData( package_change );
                 }
 
                 foreach ( var app_change in msg.app_changes )
                 {
-                    AppChanges.Add( app_change.appid, new PICSChangeData( app_change ) );
+                    AppChanges.Add[ app_change.appid ] = new PICSChangeData( app_change );
```
Ne3tCode: `.Add`

SteamKit2/SteamKit2/Steam/Handlers/SteamApps/Callbacks.cs
```diff
@@ -547,12 +547,12 @@ internal PICSProductInfoCallback( JobID jobID, CMsgClientPICSProductInfoResponse
 
                 foreach ( var package_info in msg.packages )
                 {
-                    Packages.Add( package_info.packageid, new PICSProductInfo( package_info ) );
+                    Packages[ package_info.packageid ] new PICSProductInfo( package_info );
```
Ne3tCode: Missing `=`

yaakov-h:
@JustArchi my first thought was also just to run a `.Distinct()` over the input, but there's still no guarantee that the response proto will contain unique values by appid. Valve could just decide to send us 50 identical apps in the same proto for no reason at all.

PR #804, opened 2019-12-18
Fixed NetHook2 for latest steam version.

Sen66:
Updated MsgInfo_t struct, 2 members swapped position.
Updated signature for MessageList, strings are stored in HEAP, therefore removed the steamclient.dll range check.

### DIFF
--- a/Resources/NetHook2/NetHook2/crypto.cpp
+++ b/Resources/NetHook2/NetHook2/crypto.cpp
@@ -16,10 +16,10 @@ bool (__cdecl *GetMessageFn)( int * ) = nullptr;
 
 static_assert(sizeof(MsgInfo_t) == 20, "Wrong size of MsgInfo_t");
 static_assert(offsetof(MsgInfo_t, eMsg) == 0, "Wrong offset of MsgInfo_t::eMsg");
-static_assert(offsetof(MsgInfo_t, pchMsgName) == 4, "Wrong offset of MsgInfo_t::pchMsgName");
+static_assert(offsetof(MsgInfo_t, pchMsgName) == 16, "Wrong offset of MsgInfo_t::pchMsgName");
 static_assert(offsetof(MsgInfo_t, nFlags) == 8, "Wrong offset of MsgInfo_t::nFlags");
 static_assert(offsetof(MsgInfo_t, k_EServerTarget) == 12, "Wrong offset of MsgInfo_t::k_EServerTarget");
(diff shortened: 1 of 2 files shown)


yaakov-h:
> strings are stored in HEAP

What do you mean by this bit? They're not longer in the DLL's string table?

Sen66:
> > strings are stored in HEAP
> 
> What do you mean by this bit? They're not longer in the DLL's string table?

exactly. Thats what the table looks like currently:
WcC0GL3 1

m4dEngi:
> > strings are stored in HEAP
> 
> What do you mean by this bit? They're not longer in the DLL's string table?
MessageList constructor populates descriptor fields using info from protobuf. That's why i prefer just to use clients own features for descriptor string lookup, just seem safer. (it'll call constructor and know array boundaries. It also look simpler)

Sen66:
> > > strings are stored in HEAP
> > 
> > 
> > What do you mean by this bit? They're not longer in the DLL's string table?
> 
> MessageList constructor populates descriptor fields using info from protobuf. That's why i prefer just to use clients own features for descriptor string lookup, just seem safer. (it'll call constructor and know array boundaries. It also look simpler)

 It still does the same basically as it did before.
LOKewmp 1
Thats the old GetMessageList compared to the "new" one. They moved the "MsgList" variable into the function(it was inlined most likely). I updated the signature to point to the variable inside the function and the size as well. Also I don't know if its new, but it does some hashing of the strings now inside the function.

m4dEngi:
> > > > strings are stored in HEAP
> > > 
> > > 
> > > What do you mean by this bit? They're not longer in the DLL's string table?
> > 
> > 
> > MessageList constructor populates descriptor fields using info from protobuf. That's why i prefer just to use clients own features for descriptor string lookup, just seem safer. (it'll call constructor and know array boundaries. It also look simpler)
> 
> It still does the same basically as it did before.
> LOKewmp 1
> Thats the old GetMessageList compared to the "new" one. They moved the "MsgList" variable into the function(it was inlined most likely). I updated the signature to point to the variable inside the function and the size as well. Also I don't know if its new, but it does some hashing of the strings now inside the function.

That function is a global MessageList singleton accessor.

Something like 

`CMessageList* GetMessageList() {
    static CMessageList msgList;
    return &msgList;
}
`

yaakov-h:
Closed in favour of #765.

Issue #818, opened 2020-03-12
SteamWorkshop.EnumeratePublishedFiles times out


RyanTT:
```
var allAvailableIds = _steamWorkshop.EnumeratePublishedFiles(new SteamWorkshop.EnumerationDetails
{
    AppID = (uint)appId,
    StartIndex = 0,
    Count = 100
}).GetAwaiter().GetResult();
```

This code worked fine at the start of the year but eventually stopped working (I'm fairly sure I didn't make any changes to the codebase at that time), and I can't get it to work even in a very basic connect, login, call this method example. I've tried seperate Steam accounts which all own the games I'm trying to get the workshop items from. I assume I'm just blindly missing something here?

xPaw:
I just tried updating protobufs, and looks like `CMsgCREEnumeratePublishedFilesResponse` was removed at some point, which is used by `EnumeratePublishedFiles`

RyanTT:
Thanks for the quick response, I've tried that now and simply get a "System.NotSupportedException" when trying to run this code

```csharp
var unifiedMessages = steamClient.GetHandler<SteamUnifiedMessages>();
var pubFileService = unifiedMessages.CreateService<IPublishedFile>();

var a = pubFileService.SendMessage(x => x.QueryFiles(new CPublishedFile_QueryFiles_Request {
    appid = 221100,
    page = 1,
    numperpage = 10,
    query_type = 11,
    filetype = (uint)EWorkshopFileType.GameManagedItem,
})).ToTask().GetAwaiter().GetResult();
```

Tried fiddling around with the parameters a little, omitting query_type and filetype, setting the for given parameters corresponding "xxxSpecified" flags aswell, but no luck.

yaakov-h:
This method has now been removed.

I believe the NotSupportedException is coming from our expression tree parser. If you move the creation of `CPublishedFile_QueryFiles_Request` out of the lambda expression, I believe it should work.

PR #819, opened 2020-03-12
:shipit: publish release 2.3.0

yaakov-h:
add release notes, bump version


xPaw:
A few prs that are probably worth mentioning: #761 #779 #787 #774

yaakov-h:
ugh I forgot to tag those with the milestone

yaakov-h:
Waiting for final enum and proto updates

xPaw:
Updating protos is a bit problematic because it requires rewriting `EnumeratePublishedFiles`, see #818.

Underlords protos also require compiling in a few more protos like usermessages because they added includes.

And something in Steam protos needed a `using` or a different namespace too.

PR #825, opened 2020-03-31
Fixed AyncJob support for GetLobbyListCallback

### LINKED ISSUE #824: AsyncJob from GetLobbyList always times out.
When I await the AsyncJob from SteamMatchmaking.GetLobbyList() it always times out and doesn't receive a GetLobbyListCallback. However the client does receive a GetLobbyListCallback, which I can access by registering a function using the CallbackManager.

Benjamin-Dobell:
Closes #824


Tyaap:
Thank you for fixing this. :)

PR #827, opened 2020-04-03
Update protobufs

xPaw:
1. Skipped Underlords protos because it requires some generation changes.
2. Removed `EnumeratePublishedFiles` because all the protos for it are gone; can't obsolete it because `CMsgCREEnumeratePublishedFilesResponse` was removed and it won't compile.
3. Combined `SteamKit2.Unified.Internal` to `SteamKit2.Internal` because some protos depend on each other and they're in separate namespaces.
    If this change is not acceptable, then protobuf generation needs to be upgraded to allow adding usings.


yaakov-h:
I think this makes sense, but wouldn't mind a second opinion.

What changed need to be made for Underlords?

Resources/ProtobufGen/ProtobufGen/ProtobufGen.csproj
```diff
@@ -2,7 +2,9 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <RollForward>LatestMajor</RollForward>
```
JustArchi, line 6: Isn't this the default option?

yaakov-h: Nope. I believe the default is `LatestMinor`.

JustArchi: https://docs.microsoft.com/en-us/dotnet/core/whats-new/dotnet-core-3-0#runtimesdk

So freakin' complicated, apparently `Minor` is the default, thanks 😁

yaakov-h: Ah yeah that does make more sense...

xPaw:
> What changes need to be made for Underlords?

Needs to bring in a bunch of more proto files like user messages, can't be bothered to deal with it.

VoiDeD:
> I think this makes sense, but wouldn't mind a second opinion.

Most of thisl looks good to me as well - I think I originally suggested merging the internal namespaces on IRC. They're internal anywhere so the implication is that breakage here from time to time is expected.

I could go either way on `EnumeratePublishedFiles` - should it be an Obsolete stub that does a no-op, should it throw an exception, or should it be removed entirely? All three approaches seem fine.

yaakov-h:
An obsolete stub could just cause consumers to leak resources, waiting for an event that never fires.

Throwing an exception sounds good, but removing the method will also cause an exception to be thrown... just far more cryptically.

I think all-up I'd rather keep the method, make it throw an exception, mark as obsolete and optionally hide it with `EditorBrowsableAttribute`.

xPaw:
Anything blocking this? There are proto updates that I'd like to use.

PR #828, opened 2020-04-03
 Handle CMsgClientMMSLobbyData with null metadata 

### LINKED ISSUE #826: Disconnection when calling GetLobbyData() on a deleted lobby.
GetLobbyData() results in a Steam disconnection if called on a lobby that was obtained from GetLobbyList() but has been subsequently deleted. There is currently no way to guarantee that a lobby has not been deleted before calling GetLobbyData(), and therefore avoiding disconnection. Might require major changes to the SteamKit architecture. Further details are discussed in pull request #704.

Benjamin-Dobell:
Closes #826

### DIFF
--- a/SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/Lobby.cs
+++ b/SteamKit2/SteamKit2/Steam/Handlers/SteamMatchmaking/Lobby.cs
@@ -367,9 +367,9 @@ namespace SteamKit2
                 }
             }
 
-            internal static ReadOnlyDictionary<string, string> DecodeMetadata( byte[] buffer )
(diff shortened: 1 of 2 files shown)


azuisleet:
The assertion `cachedLobby != null` in `HandleLobbyData`, is that still correct for a lobby we discover through friend presence data? Will that normally be hit for such a case?

Benjamin-Dobell:
Yeah, I don't believe it was ever accurate actually, it was added in:

https://github.com/SteamRE/SteamKit/commit/c67f2ab5c080114d5c4e1e97c8e5e4e4a4bb38d7#diff-afcd8513c5d05604d1e1610042a9f576R591

I'll add in a commit to this PR removing it.

azuisleet:
Ah yes, the assertion was for the previously non-nullable access to `Members`.

PR #836, opened 2020-04-21
Add email address callback

csandr:


yaakov-h:
Why not expose the other fields on the message as well?

PR #841, opened 2020-05-06
Make DepotManifest constructor public

### LINKED ISSUE #840: Make DepotManifest constructor public
That would allow parsing depotcache manifests on disk.

xPaw:
Fixes #840

I successfuly used it in https://github.com/xPaw/SteamDepotFileVerifier by using reflection.


yaakov-h:
You can probably simplify your “dumb” hack with Activator.CreateInstance.

Id actually rather make this a static method. I’m pretty firmly in the “constructors should do no work” camp and boy is this one doing a lot of work.

PR #845, opened 2020-05-14
Create obfuscated_private_ip object when LoginID is provided

xPaw:


JustArchi:
Maybe it'd be a good idea to somehow calculate v6 from LoginID as well in this case? I mean, it's not any real IP for the expected use case, just a value we use for session identification purposes, we might as well do our best guess and supply a value there as well.

xPaw:
The question is whether user should be able to provide their own v6 loginid or not. If not, it could be simply set into the v6 field as bytes (with 96 bits remaining zeroes).

We opted to wait for Steam to actually have IPv6 servers, then we could decide what to do with it.

JustArchi:
I was thinking about it too, with the initial idea of "let's just add loginIDv6 then" and realized that it defeats the purpose of loginID - we don't really want to provide fake IP addresses there, only allow multiple sessions from the same machine without going through hoops, in fact, I'd appreciate if that was no longer needed, but I checked a few months ago and sadly it was still in-place.

So my idea is we should just use loginID as arbitrary session number, and how it translates to Steam's backend should remain SK2 implementation detail, as to not give user the false impression of what it's used of.

`shady`

yaakov-h:
with IPv6 we could just assign a million different addresses to one machine... 🤔

PR #849, opened 2020-05-14
Remove Max value from enums and update some

xPaw:


yaakov-h:
```
Build FAILED.
Types\SteamID.cs(462,99): error CS0117: 'EAccountType' does not contain a definition for 'Max' [C:\projects\steamkit\SteamKit2\SteamKit2\SteamKit2.csproj]
Types\SteamID.cs(465,101): error CS0117: 'EUniverse' does not contain a definition for 'Max' [C:\projects\steamkit\SteamKit2\SteamKit2\SteamKit2.csproj]
    0 Warning(s)
    2 Error(s)
Time Elapsed 00:00:27.28
```
😢

SteamKit2/SteamKit2/Types/SteamID.cs
```diff
@@ -459,10 +459,10 @@ public bool IsValid
         {
             get
             {
-                if ( this.AccountType <= EAccountType.Invalid || this.AccountType >= EAccountType.Max )
+                if ( this.AccountType <= EAccountType.Invalid || this.AccountType > EAccountType.AnonUser )
```
JustArchi, line 462: This is prone to errors in the future, especially if the enum gets extended, it's probably better to use a combination of `!Enum.IsDefined() || this.AccountType == EAccountType.Invalid` or if you don't want the performance cost with it, explicitly extract it to some place that tracks valid account types.

Same below.

xPaw: If either of these enums change, SteamID code will require changes to handle them regardless.

yaakov-h: Can we add a unit test then that these values are the highest values, with a comment to update the SteamID code if it ever changes?

xPaw: Done.

PR #881, opened 2020-07-05
Fix dependabot (I hope), and just whats needed

yaakov-h:
@xPaw you were right on IRC, Samples and NHA2 aren't needed anymore as they pull in their dependencies implicitly through SteamKit2.csproj, but ProtobufDumper and ProtobufGen are both standalone.

Think its worth adding them like here, or not?

It means that we'll get 3 PRs if protobuf-net has an update, but should only get 1 PR if any other package (e.g. XUnit) has an update.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,7 +1,18 @@
 version: 2
 updates:
 - package-ecosystem: nuget
-  directory: "/"
+  directory: "/SteamKit2"
   schedule:
     interval: daily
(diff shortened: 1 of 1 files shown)


xPaw:
> 1 PR if any other package (e.g. XUnit) has an update.

Don't you need to add Tests project for that?

yaakov-h:
I believe it autodetects the solution, we've had dependabot PRs for XUnit and MS.Test.Net.Sdk before.

PR #891, opened 2020-07-08
Implement GitHub Actions

yaakov-h:
Shall we replace AppVeyor and Travis with this?

Still todo:

- [x] automatic versioning
- [x] create nupkg
- [x] publish to NuGet gallery
- [x] publish to GitHub releases
- [x] OpenCover (do we really care about this?)


yaakov-h:
I'm tempted to reverse the publishing workflow.

Old workflow:
- Create Git tag
- Push Git tag
- CI/CD creates an empty Release
- CI/CD uploads NuGet package to Release
- CI/CD uploads NuGet package to Gallery
- Manually add notes to Release

Reversed workflow:
- Create GitHub Release
- GitHub creates Tag
- Actions builds NuGet package
- Actions uploads NuGet package to Release
- Actions uploads NuGet package to Gallery

i.e. the entrypoint for developers to create a new release becomes the GitHub Releases page, rather than the Git CLI.

yaakov-h:
I've implemented a version of the publishing workflow outlined above, but I can't really test it until we merge this and publish a new release. I'm considering publishing a 2.4.0-Alpha.1 release after merging to (a) test this and (b) publish a version that uses protobuf-net 3.0 so that we can find any issues there if there are any to be found.

I'm also tempted to remove OpenCover/Codecov outright, as it seems that the servers are quite intermittent, the service doesn't always work, and we don't seem to be getting any value from it anyway. We also don't exactly have high coverage in the first place. Thoughts, @VoiDeD?

VoiDeD:
> I'm also tempted to remove OpenCover/Codecov outright, as it seems that the servers are quite intermittent, the service doesn't always work, and we don't seem to be getting any value from it anyway. We also don't exactly have high coverage in the first place. 

No skin off my back - it was a neat idea at the time, but yeah we don't really utilize it enough nor do those metrics drive anything.

Since we added it, I've changed my opinion on what coverage means and how useful of a number it really is (it isn't, and there's lots of arguments on the interwebs about this that probably don't need repeating here 😄).

PR #897, opened 2020-07-14
Enable strong-name signing

### LINKED ISSUE #896: SteamKit2 does not have a strong name
Which is odd because it does have a snk:
https://github.com/SteamRE/SteamKit/blob/80f48437d9955a9323df4244d22a3dfd4961af7f/SteamKit2/SteamKit2/SteamKit2.csproj#L5

> CSC : warning CS8002: Referenced assembly 'SteamKit2, Version=2.3.0.1249, Culture=neutral, PublicKeyToken=null' does not have a strong name.

Maybe it's due to the path that goes up the directory?

yaakov-h:
Fixes #896.


yaakov-h:
This seems to have worked.

Want to verify, @xpaw?

Produced package is [here](https://github.com/SteamRE/SteamKit/suites/912885475/artifacts/11062516).

PR #910, opened 2020-08-27
Update protobufs and EOSType

xPaw:
`appinfo.only_public` was renamed to obsolete, I haven't tested if that option applied at all anymore, but if it doesn't probably worth removing it completely.

EDIT: I tried setting `only_public = true`, but Steam sent me full data anyway, and `OnlyPublic` was false.


yaakov-h (approved):
LGTM.

Confirmed macOS kernel versions for Catalina and Big Sur.

PR #911, opened 2020-08-27
Remove only_public from pics requests

xPaw:
See #910 

It was renamed to `only_public_obsolete` in protos, and I believe it has no effect.

Question: Is it a good idea to make `PICSRequest` a struct?

### DIFF
--- a/SteamKit2/SteamKit2/Steam/Handlers/SteamApps/SteamApps.cs
+++ b/SteamKit2/SteamKit2/Steam/Handlers/SteamApps/SteamApps.cs
@@ -24,7 +24,7 @@ namespace SteamKit2
         /// <summary>
         /// Represents a PICS request used for <see cref="SteamApps.PICSGetProductInfo"/>
         /// </summary>
-        public sealed class PICSRequest
+        public struct PICSRequest
 #pragma warning restore 0419
         {
             /// <summary>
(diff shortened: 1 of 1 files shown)


yaakov-h (commented):
At a high level it looks OK but this is both a bincompat and srccompat breaking change. We'll need to carefully call this out in release notes.

As for making it a struct - seems like a decent idea, I don't see any downsides.

SteamKit2/SteamKit2/Steam/Handlers/SteamApps/SteamApps.cs
```diff
@@ -205,18 +183,17 @@ public AsyncJob<PICSChangesCallback> PICSGetChangesSince( uint lastChangeNumber
         /// </summary>
         /// <param name="app">App id requested.</param>
         /// <param name="package">Package id requested.</param>
-        /// <param name="onlyPublic">Whether to send only public information.</param>
         /// <param name="metaDataOnly">Whether to send only meta data.</param>
         /// <returns>The Job ID of the request. This can be used to find the appropriate <see cref="PICSProductInfoCallback"/>.</returns>
-        public AsyncJobMultiple<PICSProductInfoCallback> PICSGetProductInfo(uint? app, uint? package, bool onlyPublic = true, bool metaDataOnly = false)
```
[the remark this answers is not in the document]
xPaw: I never liked these bools, as overload that takes `IEnumerable<PICSRequest` did not have `onlyPublic` already.

SteamKit2/SteamKit2/Steam/Handlers/SteamApps/SteamApps.cs
```diff
@@ -263,7 +239,7 @@ public AsyncJobMultiple<PICSProductInfoCallback> PICSGetProductInfo( IEnumerable
                 var appinfo = new CMsgClientPICSProductInfoRequest.AppInfo();
                 appinfo.access_token = app_request.AccessToken;
                 appinfo.appid = app_request.ID;
-                appinfo.only_public_obsolete = app_request.Public;
+                appinfo.only_public_obsolete = false;
```
yaakov-h, line 242: can we just omit it entirely here?

xPaw: We can, but I don't know what Valve defaults to on the backend, if at all (since it doesn't seem to have an effect).

PR #912, opened 2020-08-31
Fix unsupported registry on netstandard2.0

js6pak:
Fixes `PlatformNotSupportedException` on Windows and netstandard2.0

Without these changes following exception was thrown
```
 System.AggregateException: One or more errors occurred. (Registry is not supported on this platform.)
  ---> System.PlatformNotSupportedException: Registry is not supported on this platform.
     at Microsoft.Win32.RegistryKey.OpenBaseKey(RegistryHive hKey, RegistryView view)
	 at SteamKit2.WindowsInfoProvider.GetMachineGuid()
     at SteamKit2.HardwareUtils.GenerateMachineID()     at System.Threading.Tasks.Task`1.InnerInvoke()
     at System.Threading.Tasks.Task.<>c.<.cctor>b__274_0(Object obj)
     at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
```

### DIFF
--- a/SteamKit2/SteamKit2/Util/HardwareUtils.cs
+++ b/SteamKit2/SteamKit2/Util/HardwareUtils.cs
@@ -85,9 +85,18 @@ namespace SteamKit2
     {
         public override byte[] GetMachineGuid()
         {
-            RegistryKey localKey = RegistryKey
-                .OpenBaseKey( Microsoft.Win32.RegistryHive.LocalMachine, RegistryView.Registry64 )
-                .OpenSubKey( @"SOFTWARE\Microsoft\Cryptography" );
+            RegistryKey? localKey;
+
+            try
+            {
+                localKey = RegistryKey
+                    .OpenBaseKey( RegistryHive.LocalMachine, RegistryView.Registry64 )
+                    .OpenSubKey( @"SOFTWARE\Microsoft\Cryptography" );
+            }
+            catch ( PlatformNotSupportedException )
+            {
+                return base.GetMachineGuid();
+            }
 
             if ( localKey == null )
             {


yaakov-h:
`PlatformNotSupportedException` in this case indicated programmer error, and is not something that should be ignored.

What application have you built for GitHub Actions, and how is it being published and run?

js6pak:
After futher testing on [clean repo](https://github.com/js6pak/SteamKit.pull912/actions), I can't reproduce [this](https://github.com/SixModLoader/SixModLoader/runs/1051175399) error again, so its probably package incompatibility on my side?

yaakov-h:
It looks like you're trying to publish a [.NET Standard](https://github.com/SixModLoader/SixModLoader/blob/a8e3e7deace14a5a12a959887919fad9de9d10d0/SixModLoader.MSBuild/SixModLoader.MSBuild.csproj#L4) assembly.

That won't work with NuGet resolution where only reference assemblies are provided for netstandard, such as Microsoft.Win32.Registry. You need to publish for a particular runtime (i.e. windows), at which point NuGet will use the assemblies from the `runtimes\win\lib\netstandard2.0` folder (which contains the actual Windows runtime implementation) instead of the `lib\netstandard2.0` folder (which is just a stub assembly).

I don't know what the official guidance is on custom MSBuild tasks that have platform-specific dependencies. Perhaps [dotnet/sdk](https://github.com/dotnet/sdk) can help you there?

js6pak:
Just [using net472](https://github.com/SixModLoader/SixModLoader/blob/master/SixModLoader.MSBuild/SixModLoader.MSBuild.csproj#L4) as target frameworks seems to resolve the issue, but I still think that this PR should be merged as using registry there is not required and working netstandard is more important imo.

VoiDeD:
@js6pak you'll still want to target `netstandard` if you're publishing a library - it's just your msbuild task code that needs to target a specific RID, as yaakov mentioned.

The `WindowsInfoProvider` in SteamKit is gated by OS platform checks already, so this scenario is highly exceptional where a binary packaging misconfiguration on your end is causing the issue.

yaakov-h:
Yep. It's a packaging error, and PNSE is not something that should be caught.

PR #932, opened 2020-11-09
Fix AsyncJob for GameCoordinator.MessageCallback

Vital7:
Fixes JobID for MessageCallback in order to use value from GC message header.

Related: #824


yaakov-h:
How is this related to the lobby system?

Vital7:
It's not related to lobby system, it's related to the incorrect JobID in callbacks. Not a good example though, I agree.

PR #937, opened 2020-11-16
Write empty values (null and no children) as an object when saving to binary

xPaw:
This will match the behaviour of text saving. And looking at Valve's code, they default to type none, so when serializing it should be saved as an object as well.

I believe right now deserializing an empty object, and then serializing back into binary would not be roundtrip in Steamkit.

https://github.com/SteamRE/SteamKit/blob/df7f30f0d2880d8d9ceaf0d6b1e3c5d8f04bf7a7/SteamKit2/SteamKit2/Types/KeyValue.cs#L766-L769


yaakov-h:
> I believe right now deserializing an empty object, and then serializing back into binary would not be roundtrip in Steamkit.

Can you add a unit test for that? 😃

xPaw:
I did?

PR #951, opened 2020-12-08
Updated NetHooks CWebSocketConnection::BBuildAndAsyncSendFrame signature

m4dEngi:
Hope i didn't mess it up...


VoiDeD:
>Hope i didn't mess it up...

Does it work? 😄

m4dEngi:
> > Hope i didn't mess it up...
> 
> Does it work? smile

Of course it does. I was more concerned about text formatting and line breaks.

VoiDeD:
>Of course it does. I was more concerned about text formatting and line breaks.

No worries, we're not too stringent on line breaks (good editors don't care), and the formatting looked fine.

Thanks for the fix!

PR #957, opened 2021-01-09
Update protobufs

xPaw:
I added a script that updates submodules and runs regen.

Not keeping `ClientNoLongerSupported` as deprecated because it never was in a SteamKit release.


yaakov-h:
I don't spy any changes to the generator, so where did the IDE0079 suppression come from?

xPaw:
I'd wager from protobuf library being updated since last run?

PR #965, opened 2021-03-17
Update nethook to protobuf-2.6.1

xPaw:
That's what Steam uses now. I haven't built locally, want to see if actions succeeds.


yaakov-h:
> D:\a\SteamKit\SteamKit\Resources\NetHook2\NetHook2\steammessages_base.pb.h(17,1): fatal error C1189: #error:  This file was generated by an older version of protoc which is [D:\a\SteamKit\SteamKit\Resources\NetHook2\NetHook2\NetHook2.vcxproj]

xPaw:
Yep.

There is detours.lib, libprotobuf.lib, libprotobuf_r.lib, zlib.lib in the repo. Setting up dependencies doesn't actually build these.

These files add up to like 44mb in the repo. I don't know how to fix it so it builds from the pulled source instead of keeping these libs in the repo.

PR #971, opened 2021-03-22
Fix nullability warnings

yaakov-h:


SteamKit2/SteamKit2/Steam/Handlers/SteamUnifiedMessages/Callbacks.cs
```diff
@@ -101,30 +99,26 @@ public string RpcName
...
-            public object? Body { get; private set; }
+            public object Body { get; private set; }
 
 
             internal ServiceMethodNotification( Type messageType, IPacketMsg packetMsg )
             {
                 // Bounce into generic-land.
                 var setupMethod = GetType().GetMethod( nameof(Setup), BindingFlags.Instance | BindingFlags.NonPublic ).MakeGenericMethod( messageType );
-                setupMethod.Invoke( this, new[] { packetMsg } );
+                (MethodName, Body) = ((string, object))setupMethod.Invoke( this, new[] { packetMsg } );
```
yaakov-h, line 114: This cast needs to be tested.

xPaw: I replaced the dll in steamdb and ran `pubfile` command which uses `UnifiedService<IPublishedFile>`, it worked. I am not 100% certain whether it went through this invocation though.

yaakov-h: IIRC I can trigger this by enabling parental controls. Let me check...

yaakov-h: Cast works but reflection returned null, lol

SteamKit2/SteamKit2/Steam/Discovery/BasicServerListProto.cs
```diff
@@ -1,13 +1,11 @@
...
+﻿using ProtoBuf;
 
 namespace SteamKit2.Discovery
 {
     [ProtoContract]
     class BasicServerListProto
     {
         [ProtoMember(1)]
-        [DisallowNull, NotNull]
         public string? Address { get; set; }
```
xPaw, line 9: How can this be null anyway?

yaakov-h: good question. can you deserialize a null value or does it come back as an empty string?

JustArchi: Lack of string value is deserialized as empty string. You can't distinguish between lack of string value and empty string as both are not rendered in proto. `null` value is illegal for proto strings (depending on the lib it'll fail when you attempt to assign it in proto object, or at the very latest during serialization).

JustArchi: So yes, if this property is set straight from proto and never through any other means, it'll never be `null`. It can still be empty though.

xPaw: Considering only SK2 serializes this proto, it should never be empty in the first place.

JustArchi: Then yeah it should be set as non-null, because you can't serialize null value in the proto, at least without resorting to custom hacks such as `var serialized = Address ?? ""`. It should just be a `string` which doesn't accept nulls, exactly the same as previous annotations (which might not be needed anymore with nullability enabled, they're redundant).

SteamKit2/SteamKit2/Steam/Discovery/FileStorageServerListProvider.cs
```diff
@@ -34,11 +34,11 @@ public Task<IEnumerable<ServerRecord>> FetchServerListAsync()
...
                     using (FileStream fileStream = File.OpenRead(filename))
                     {
-                        return Serializer.DeserializeItems<BasicServerListProto>(fileStream, PrefixStyle.Base128, 1)
-                            .Select(item =>
-                            {
-                                return ServerRecord.CreateServer(item.Address, item.Port, item.Protocols);
-                            })
+                        return Serializer.DeserializeItems<BasicServerListProto>(fileStream, PrefixStyle.Base128, 1 )
+                            .Select( item => ServerRecord.CreateServer(
+                                 item.Address ?? throw new InvalidDataException( "Serialized server had null address" ),
```
xPaw, line 39: You can undo this now

PR #974, opened 2021-03-26
High DPI support for NHA2

Vital7:
Enables using high DPI support for NHA2 and removes old workaround disabling `EnableVisualStyles` since it's fixed with .NET runtime 5.0.4.


yaakov-h:
@xPaw IIRC this was your handiwork, can you please verify?

xPaw:
Yeah, that bug was fixed in 5.0.4.

PR #975, opened 2021-03-26
ClientServiceMethodLegacy/Response fixes for NHA2

Vital7:
1. Fixes deserializing `ClientServiceMethodLegacy` message (earlier it was parsed as Dictionary, because `ClientServiceMethodResponseSpecialization` did not have support for this type).
2. Fixes `ArgumentNullException` when NHA2 tries to display `CMsgClientServiceMethodLegacyResponse` with underlying response being `null` (e.g. `NoResponse` type in protobufs).


yaakov-h:
Do we need to be concerned with `CMsgClientServiceMethodLegacyResponse` too?

Vital7:
NHA2 deserializes CMsgClientServiceMethodLegacyResponse properly, the only problem was an exception occuring when message body is null.

PR #980, opened 2021-04-11
NH2: Display build info when initializing

### LINKED ISSUE #979: Add build date and/or git commit information to nethook console logs
People create issues that nethook does not work but do not specify which version they are using.

yaakov-h:
Example (dirty working tree):

image

Example (clean working tree):

image

Format is:

> Built at \<build time\> from \<commit sha\>[/dirty] (\<commit date\>)

Resolves #979.

### DIFF
--- a/Resources/NetHook2/GenerateVersionInfo.ps1
+++ b/Resources/NetHook2/GenerateVersionInfo.ps1
@@ -7,8 +7,8 @@ Function Test-GitAvailable {
         }
     }
     Catch {
-        return $false
     }
+    return $false
 }
 
 $buildDate = Get-Date -Format "yyyy-MM-dd HH:mm:ss K"
@@ -19,6 +19,8 @@ if (Test-GitAvailable) {
 
     git diff --quiet
     $dirty = [bool]($LASTEXITCODE)
+} else {
+    $dirty = $false
 }
 
 $versionFile = 'version.cpp'
--- a/Resources/NetHook2/GenerateVersionInfo.ps1
+++ b/Resources/NetHook2/GenerateVersionInfo.ps1
@@ -1,0 +1,34 @@
+﻿$ErrorActionPreference = 'stop'
+
+Function Test-GitAvailable {
+    Try {
+        If (Get-Command git) {
+            return $true
+        }
+    }
+    Catch {
+        return $false
+    }
+}
+
+$buildDate = Get-Date -Format "yyyy-MM-dd HH:mm:ss K"
+
+if (Test-GitAvailable) {
+    $commitDate = [string](git show -s --format="%ci" HEAD)
+    $sha = [string](git rev-parse --short HEAD)
+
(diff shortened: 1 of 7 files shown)


Resources/NetHook2/GenerateVersionInfo.ps1
```diff
@@ -0,0 +1,34 @@
...
+        If (Get-Command git) {
+            return $true
+        }
+    }
+    Catch {
+        return $false
+    }
+}
+
+$buildDate = Get-Date -Format "yyyy-MM-dd HH:mm:ss K"
```
xPaw, line 14: Add `-AsUTC`?

Resources/NetHook2/GenerateVersionInfo.ps1
```diff
@@ -0,0 +1,34 @@
...
+
+if (Test-GitAvailable) {
```
xPaw, line 16: Don't you want to add default/fallback variables if it happens to be not available?

yaakov-h: the strings should become empty string, I need to test the bool.

Resources/NetHook2/GenerateVersionInfo.ps1
yaakov-h: >Get-Date : A parameter cannot be found that matches parameter name 'AsUtc'.

yaakov-h:
Build info from the DLL produced by GitHub Actions. The build from the PR was from `+ddc051dfbdbb49fedf79b659f72fef59e2de9336:refs/remotes/pull/980/merge`.

image

Resources/NetHook2/GenerateVersionInfo.ps1
xPaw: Needs newer powershell I guess?

https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/get-date?view=powershell-7.1#example-10--return-a-date-value-interpreted-as-utc

yaakov-h: Probably. PowerShell Core adds a bunch of new stuff... but also removes some stuff. It's also not preinstalled with Windows so we can't rely on it existing.

psychonic: `[DateTime]::UtcNow.ToString('yyyy-MM-dd HH:mm:ss K')`

yaakov-h: tbh I’d rather keep it in the system timezone unless you can also figure out how to get the git commit time in utc. I guess we could parse and reformat it?

xPaw: Getting git commit time is pointless when you already have the commit.

xPaw (approved):

PR #983, opened 2021-05-17
Add LogOnDetails.DisablePartnerAutogrants

xPaw:
It should be same as this steam cli flag: `-disablepartnerlicenses - disable automatically granted partner licenses`


xPaw:
So I tested this flag on my account, `disable_partner_autogrants = true` actually removes the licenses from your account. And they are restored if you login again with the flag set to false.

I'm gonna close this PR because this sounds like a nightmare.

PR #995, opened 2021-06-13
WebAPI: Treat the args dictionary as immutable

yaakov-h:
This PR follows on from #992 by avoiding making any changes to the arguments dictionary in the first place.

I've rewritten the flow of the HTTP request builder function so that instead of updating the dictionary and then transforming the dictionary into text, we write out the text values directly, and then copy over the existing dictionary values.

cc/ @JustArchi @xPaw

### DIFF
--- a/SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs
+++ b/SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs
@@ -222,20 +222,13 @@ namespace SteamKit2
                     throw new ArgumentNullException( nameof(func) );
                 }
 
-
-                if ( args == null )
-                {
-                    args = new Dictionary<string, object>();
-                }
-                else if ( args.TryGetValue( "format", out var format ) )
+                if ( args != null && args.TryGetValue( "format", out var format ) )
                 {
                     if ( !( format is string formatText ) || formatText != "vdf" )
                     {
                         throw new ArgumentException( $"{nameof( args )} include unsupported {nameof( format )}: {format}" );
                     }
                 }
-
-                args ??= new Dictionary<string, object>();
 
                 var paramBuilder = new StringBuilder();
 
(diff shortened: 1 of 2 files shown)


SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs
```diff
@@ -222,68 +222,61 @@ public async Task<KeyValue> CallAsync( HttpMethod method, string func, int versi
                     throw new ArgumentNullException( nameof(func) );
                 }
 
-                if ( args == null )
+                if ( args != null && args.TryGetValue( "format", out var format ) )
                 {
-                    args = new Dictionary<string, object>();
+                    if ( !( format is string formatText ) || formatText != "vdf" )
+                    {
+                        throw new ArgumentException( $"{nameof( args )} include unsupported {nameof( format )}: {format}" );
```
xPaw, line 229: Printing {format} seems sketchy when it can be a non string. Perhaps just reword this exception to say that only format=vdf is supported?

SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs
```diff
@@ -222,68 +222,61 @@ public async Task<KeyValue> CallAsync( HttpMethod method, string func, int versi
...
+                    }
                 }
 
-
-                var urlBuilder = new StringBuilder();
                 var paramBuilder = new StringBuilder();
```
xPaw, line 233: Is `HttpValueCollection` only in ASP?

yaakov-h: It seems to be an internal class only in both .NET Framework and Core, and does not appear as an available API anywhere on MS docs site. HttpUtility can parse but not build. FormUrlEncoded content can build but it requires having a full IEnumerable of all the values upfront, and is async.

xPaw: That's lame.

JustArchi (approved):
We must pay special attention to https://github.com/SteamRE/SteamKit/issues/641 when rewriting this, as to not regress when user supplies byte array which causes different encoding to be used. I checked the current code and it looks OK. I also didn't find any problems in regards to this, thanks for following up on this matter @yaakov-h 😎.

VoiDeD (approved):

PR #1006, opened 2021-07-22
Read all available data in CryptoStream.Read()

xPaw:
This appears to be a breaking change in .NET 6 caused by https://github.com/dotnet/runtime/pull/53644

https://github.com/dotnet/runtime/issues/55527#issuecomment-878555536

### DIFF
--- a/SteamKit2/SteamKit2/Util/CryptoHelper.cs
+++ b/SteamKit2/SteamKit2/Util/CryptoHelper.cs
@@ -161,7 +161,7 @@ namespace SteamKit2
                 using ( var ms = new MemoryStream( input ) )
                 using ( var cs = new CryptoStream( ms, aesTransform, CryptoStreamMode.Read ) )
                 {
-                    outLen = ReadAll( cs, plainText );
+                    outLen = cs.ReadAll( plainText );
                 }
 
                 byte[] output = new byte[ outLen ];
@@ -406,7 +406,7 @@ namespace SteamKit2
                     // plaintext is never longer than ciphertext
                     byte[] plaintext = new byte[ cipherText.Length ];
 
-                    int len = ReadAll( cs, plaintext );
+                    int len = cs.ReadAll( plaintext );
 
                     byte[] output = new byte[ len ];
                     Array.Copy( plaintext, 0, output, 0, len );
@@ -536,17 +536,6 @@ namespace SteamKit2
 
                 return block;
             }
-        }
-
-        private static int ReadAll( Stream stream, byte[] buffer )
-        {
-            int bytesRead;
-            int totalRead = 0;
-            while ( ( bytesRead = stream.Read( buffer, totalRead, buffer.Length - totalRead ) ) != 0 )
-            {
-                totalRead += bytesRead;
-            }
-            return totalRead;
         }
     }
 }
--- a/SteamKit2/SteamKit2/Util/CryptoHelper.cs
+++ b/SteamKit2/SteamKit2/Util/CryptoHelper.cs
@@ -161,7 +161,7 @@ namespace SteamKit2
                 using ( var ms = new MemoryStream( input ) )
                 using ( var cs = new CryptoStream( ms, aesTransform, CryptoStreamMode.Read ) )
                 {
-                    outLen = cs.Read( plainText, 0, plainText.Length );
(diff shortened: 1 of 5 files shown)


JustArchi:
My repro from #1007 in case somebody wanted to reliably test the patch:

```csharp
private const string EncryptionKey = "zmkXyWQzRHiameIe";

private static string? Decrypt(string encryptedString) {
	if (string.IsNullOrEmpty(encryptedString)) {
		throw new ArgumentNullException(nameof(encryptedString));
	}

	byte[] key;

	using (SHA256 sha256 = SHA256.Create()) {
		key = sha256.ComputeHash(EncryptionKey);
	}

	byte[] decryptedData = Convert.FromBase64String(encryptedString);
	decryptedData = CryptoHelper.SymmetricDecrypt(decryptedData, key);

	return Encoding.UTF8.GetString(decryptedData);
}

private static string? Encrypt(string decryptedString) {
	if (string.IsNullOrEmpty(decryptedString)) {
		throw new ArgumentNullException(nameof(decryptedString));
	}

	byte[] key;

	using (SHA256 sha256 = SHA256.Create()) {
		key = sha256.ComputeHash(EncryptionKey);
	}

	byte[] encryptedData = Encoding.UTF8.GetBytes(decryptedString);
	encryptedData = CryptoHelper.SymmetricEncrypt(encryptedData, key);

	return Convert.ToBase64String(encryptedData);
}

private static int Main() {
	const string decrypted = "qF7Sr1RfOZ9K9bfTXfl3";

	Console.WriteLine(decrypted);

	var encrypted = Encrypt(decrypted);

	Console.WriteLine(encrypted);

	var decrypted2 = Decrypt(encrypted);

	Console.WriteLine(decrypted2);

	return decrypted == decrypted2 ? 0 : 1;
}```

**Review by JustArchi — APPROVED** [review]
🚢

**Comment by JustArchi** [comment]
Reposting my thoughts from Discord:

https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.read?view=net-5.0

> An implementation is free to return fewer bytes than requested even if the end of the stream has not been reached.

We should include this change even if .NET folks somehow decide to revert this breaking change in .NET 6.0. The MSDN docs explicitly state that the implementation can return fewer bytes than requested, so even if it decides not to, it should be classified as bullet-proofing for the future.

To me this is one of those places where the code was wrong to begin with but it somehow worked, therefore it's IMHO worth it to fix it before everybody gets hit in the head just because .NET guys decided to enforce what the documentation has always stated.

**Comment by xPaw** [comment]
I was making a test case and found this behaviour:

CryptoStream.Read returns full data if the data length is divisible by 16. If it's not, then the last remaining chunk won't be returned and requires a second Read call.

For example for for strings with length of 16, 32, and even 1440 it returns correctly. But If I add one character, it will require the second Read() call to get that remaining character.

File: SteamKit2/SteamKit2/Util/CryptoHelper.cs [hunk 1]
```diff
@@ -538,5 +538,15 @@ public static byte[] GenerateRandomBlock( int size )
             }
         }
 
+        private static int ReadAll( Stream stream, byte[] buffer )
```
yaakov-h, line 541: I'd be tempted to make this an extension method, then we can reuse it in other places like ZipUtil below quite easily.

xPaw: I thought about it, I have no preference one way or the other. Which one you prefer?

yaakov-h: Either extension method, or helper method in Utils that ZipUtil also uses.
I’m not fussed which way you choose but I’d rather the code not be duplicated.

yaakov-h (approved):

JustArchi:
@yaakov-h @azuisleet if it doesn't clash with anything I'd appreciate `Alpha.3` with those changes so we could test .NET 6.0 compatibility on official release, thank you in advance for considering it.

yaakov-h:
I'll have a look at that tonight.

JustArchi:
RIP release https://github.com/SteamRE/SteamKit/actions/runs/1078301158

yaakov-h:
I’ve updated the secret but want to recheck a couple PRs before I resubmit the job.

PR #1008, opened 2021-07-30
Include all the protobufs that were split in last beta client update

xPaw:


yaakov-h:
LGTM. Anything notable, or do I just go for the big green Merge button?

xPaw:
I only diffed appinfo protos, and that was identical, unsure about others.

yaakov-h:
Well the compiler seems happy, so everything we depend upon still exists...

PR #1009, opened 2021-07-30
Remove PICSGetProductInfo overloads that takes uint

xPaw:
@JustArchi had a good suggestion to deprecate these overloads, as you really should be requesting app info while providing a token.

I also added an overload that takes one PICSRequest to make migrating easier.

### DIFF
--- a/SteamKit2/SteamKit2/Steam/Handlers/SteamApps/SteamApps.cs
+++ b/SteamKit2/SteamKit2/Steam/Handlers/SteamApps/SteamApps.cs
@@ -185,6 +185,7 @@ namespace SteamKit2
         /// <param name="package">Package id requested.</param>
         /// <param name="metaDataOnly">Whether to send only meta data.</param>
         /// <returns>The Job ID of the request. This can be used to find the appropriate <see cref="PICSProductInfoCallback"/>.</returns>
+        [Obsolete( "Use an overload that takes PICSRequest instead of uint. In SteamKit 2.3, this overload had four parameters, but onlyPublic was removed. Audit your code." )]
         public AsyncJobMultiple<PICSProductInfoCallback> PICSGetProductInfo(uint? app, uint? package, bool metaDataOnly = false)
         {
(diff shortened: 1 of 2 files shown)


archi:
I suppose you intended to ping @JustArchi? ;-)
If not: Yes, overload the tokens to migrate the PICSRequest for deprecation of the app info :P

JustArchi:
What @archi said 😁

SteamKit2/SteamKit2/Steam/Handlers/SteamApps/SteamApps.cs
```diff
@@ -185,6 +185,7 @@ public AsyncJob<PICSChangesCallback> PICSGetChangesSince( uint lastChangeNumber
         /// <param name="package">Package id requested.</param>
         /// <param name="metaDataOnly">Whether to send only meta data.</param>
         /// <returns>The Job ID of the request. This can be used to find the appropriate <see cref="PICSProductInfoCallback"/>.</returns>
+        [Obsolete( "Use an overload that takes PICSRequest instead of uint. In SteamKit 2.3, this overload had four parameters, but onlyPublic was removed. Audit your code." )]
```
JustArchi, line 188: I *still* believe we should just remove this overload entirely right away, so nobody disregards the warning and end up with broken functionality, but if we really intend to support it then obsoletion mark is still far better than silent breaking change.

JustArchi (approved):

yaakov-h:
Sample 9 needs to be updated so that it doesn't use deprecated APIs.

There are a few different ways to do this:

- If we care about source compatibility then we're pretty much screwed unless we keep the old method and have it throw if you set the flag to true.
- If we care about binary compatibility then we need to keep the old method, but can tag it with [Obsolete] and even [EditorBrowsable(Never)].
- If we don't care about either - which we largely do not - then we can just dump the old methods. The PICSRequest overloads should be pretty much zero-cost anyway, now that PICSRequest is a struct.

PR #1015, opened 2021-08-17
Name threads so they fit on Linux

xPaw:
Thread names on Linux are limited to 15 characters.


JustArchi (approved):

yaakov-h:
What happens at the moment? Do they get truncated, or does it just not get applied?

yaakov-h (approved):

xPaw:
Truncated

PR #1021, opened 2021-09-23
Add WebAPI overload that fetches protobuf instead of vdf

xPaw:
I only changed the response, not the input.

The api accepts `input_protobuf_encoded` which means the overload could be changed to accept a proto request.

I updated `ContentServerDirectoryService` to use this (which by the way, could actually be executed as a service call over CMClient instead).

### DIFF
--- a/SteamKit2/SteamKit2/Steam/WebAPI/ContentServerDirectoryService.cs
+++ b/SteamKit2/SteamKit2/Steam/WebAPI/ContentServerDirectoryService.cs
@@ -5,7 +5,6 @@ using System.Linq;
 using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
-using SteamKit2.Discovery;
 
 namespace SteamKit2
 {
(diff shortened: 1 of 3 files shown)


SteamKit2/SteamKit2/Steam/WebAPI/ContentServerDirectoryService.cs
```diff
@@ -78,53 +77,39 @@ static async Task<IReadOnlyCollection<CDNClient.Server>> LoadCoreAsync( SteamCon
...
-            var response = await directory.CallAsync( HttpMethod.Get, "GetServersForSteamPipe", version: 1, args: args ).ConfigureAwait( false );
-
-            var result = ( EResult )response[ "result" ].AsInteger( ( int )EResult.OK );
-            if ( result != EResult.OK || response["servers"] == KeyValue.Invalid )
-            {
-                throw new InvalidOperationException( string.Format( "Steam Web API returned EResult.{0}", result ) );
-            }
-
-            var serverList = response[ "servers" ];
+            var response = await directory.CallAsync<Internal.CContentServerDirectory_GetServersForSteamPipe_Response>( HttpMethod.Get, "GetServersForSteamPipe", version: 1, args: args ).ConfigureAwait( false );
```
yaakov-h, line 80: nameof?

yaakov-h (commented):
Overall I think it makes sense. Annoyingly that's one more set of generic type arguments to annotate for trimming, but that's the world we live in now.

PR #1027, opened 2021-10-02
Automatically update protos

yaakov-h:
Here's quick late-night hack: use GitHub Actions to automatically update our protos regularly.

(I picked Sunday morning in my time zone, I think.)

The complication here is that the repository's automatic GITHUB_TOKEN can push code and create the pull request, but that [will not trigger subsequent Actions](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token) - i.e. building and testing the new code.

IMO that part is pretty important - we need to know at a glance if a protobuf update contains any breaking changes, we can't blindly merge them if they remove or rename something that we're using.

The recommendation from the various docs I've come across is to use a Personal Access Token (PAT) from either a particular developer (to whom all the PRs will be attributed to), or a new Machine Account.

According to the GitHub user terms, everyone can create at most one actual personal account plus one Machine Account, which is basically a personal account used solely for automation.

Does anyone have a machine account that we can re-use for this, or do I need to create a new one?

### DIFF
--- a/.github/workflows/update-protobufs.yaml
+++ b/.github/workflows/update-protobufs.yaml
@@ -3,8 +3,8 @@ on:
   push:
     branches:
     - yaakov/auto-protos # temporary trigger during development
-  schedule:
-  - cron: '0 17 * * SUN'
+  #schedule:
+  #- cron: '0 17 * * SUN'
 
 jobs:
   update-protobufs:
@@ -27,12 +27,12 @@ jobs:
 
     # TODO: We need a new GitHub Machine Account (or maybe an existing one?) to generate a PAT that we can use as the token here
     # otherwise, our new changes will not trigger further Actions (on:push or on:pull_request, i.e. CI/CD builds).
-    - name: Create Pull Request
-      uses: peter-evans/create-pull-request@v3.10.1
-      with:
-          #token: ${{ secrets.PAT }}
-          commit-message: Update protobufs
-          title: Update protobufs
-          body: |
-            - Update protobufs
-          branch: auto/protobufs
+    #- name: Create Pull Request
+    #  uses: peter-evans/create-pull-request@v3.10.1
+    #  with:
+    #      #token: ${{ secrets.PAT }}
+    #      commit-message: Update protobufs
+    #      title: Update protobufs
+    #      body: |
+    #        - Update protobufs
+    #      branch: auto/protobufs
--- a/.github/workflows/update-protobufs.yaml
+++ b/.github/workflows/update-protobufs.yaml
@@ -1,4 +1,4 @@
-name: Update Dependencies
+name: Update Protobufs
 on:
   push:
     branches:
--- a/.github/workflows/update-protobufs.yaml
+++ b/.github/workflows/update-protobufs.yaml
@@ -27,12 +27,12 @@ jobs:
 
     # TODO: We need a new GitHub Machine Account (or maybe an existing one?) to generate a PAT that we can use as the token here
     # otherwise, our new changes will not trigger further Actions (on:push or on:pull_request, i.e. CI/CD builds).
-    #- name: Create Pull Request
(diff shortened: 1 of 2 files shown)


.github/workflows/update-protobufs.yaml
```diff
@@ -0,0 +1,38 @@
...
+jobs:
+  update-protobufs:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2.3.4
+
+    - name: Fetch latest protobufs
+      run: |
```
xPaw, line 18: There's a script you can call: https://github.com/SteamRE/SteamKit/blob/master/Resources/ProtobufGen/update.ps1

I don't think you need the cd/git pull?

yaakov-h: ooooh I didn't know about `--remote`

.github/workflows/update-protobufs.yaml
```diff
@@ -0,0 +1,33 @@
...
+    - name: Run Update Script
+      shell: pwsh
+      run: |
+        git submodule init
+        Resources/ProtobufGen/update.ps1
+
+    # TODO: We need a new GitHub Machine Account (or maybe an existing one?) to generate a PAT that we can use as the token here
+    # otherwise, our new changes will not trigger further Actions (on:push or on:pull_request, i.e. CI/CD builds).
+    #- name: Create Pull Request
+    #  uses: peter-evans/create-pull-request@v3.10.1
```
VoiDeD, line 26: I haven't played with GH actions much - but what does security look like for a third-party action that is fed a PAT?

Does pinning the action to a specific version guarantee it to be immutable?

VoiDeD: Not sure that we have a machine account around (I definitely don't have one), so we'll want to create one - and share creds among the core team?

The PAT will definitely need to be limited in scope to just creating PRs, if that's doable.

.github/workflows/update-protobufs.yaml
```diff
@@ -0,0 +1,33 @@
...
+    #  with:
+    #      #token: ${{ secrets.PAT }}
+    #      commit-message: Update protobufs
+    #      title: Update protobufs
+    #      body: |
+    #        - Update protobufs
+    #      branch: auto/protobufs
```
VoiDeD, line 33: Configure `branch-suffix` as well? Not sure what happens if the branch already exists and this action tries to create it again.

.github/workflows/update-protobufs.yaml
```diff
@@ -0,0 +1,33 @@
+name: Update Protobufs
+on:
+  push:
+    branches:
+    - yaakov/auto-protos # temporary trigger during development
+  #schedule:
+  #- cron: '0 17 * * SUN'
```
VoiDeD, line 7: Just dropping a comment for this to be uncommented once we're ready to merge.

.github/workflows/update-protobufs.yaml
xPaw: You can provide a full commit hash as the version afaik.

.github/workflows/update-protobufs.yaml
xPaw: Can also add `workflow_dispatch:` to be able to manually trigger this from UI.

.github/workflows/update-protobufs.yaml
xPaw: It might be that pull requests created when run by cron shouldn't require a PAT.

> Note: If you want pull requests created by this action to trigger an on: push or on: pull_request workflow then you cannot use the default GITHUB_TOKEN.

FWIW, adding `permissions` section is worthwhile:
contents: write
pull-requests: write

PR #1028, opened 2021-10-02
Allow consumers to specify a custom Machine Info Provider

### LINKED ISSUE #719: Make MachineInfoProvider class public

yaakov-h:
This allows consumers to specify a custom Machine Info Provider. If they do not specify one, we continue to use a SteamKit-supplied default based on the current operating system.

Whilst this can open the door to shady behaviour that we do not support, it also meets the legitimate case of allowing users to provide implementations on platforms that we do not neccesarily yet support (such as FreeBSD), or cannot easily support from a pure .NET Standard (or `net5.0`/`net6.0`) assembly - such as iOS, Android, or perhaps Blazor.

We retain the existing behaviour of running the generation in the background (in case it is a bit heavy/expensive), and caching the result.

It is the responsibility of the `IMachineInfoProvider` to either return a valid byte array containing a (consistent) value, or null. If the provider returns null, we fall back to `DefaultMachineInfoProvider` (formerly `DefaultInfoProvider`), which is guaranteed to return a non-null value.

It remains the responsibility of SteamKit to hash these values, build the message object, and attach the message object to login requests.

This closes #719 and supersedes #720.

### DIFF
--- a/SteamKit2/SteamKit2/Util/HardwareUtils.cs
+++ b/SteamKit2/SteamKit2/Util/HardwareUtils.cs
@@ -12,46 +12,45 @@ using static SteamKit2.Util.MacHelpers.LibC;
 using static SteamKit2.Util.MacHelpers.CoreFoundation;
 using static SteamKit2.Util.MacHelpers.DiskArbitration;
 using static SteamKit2.Util.MacHelpers.IOKit;
+using System.Runtime.CompilerServices;
 
 namespace SteamKit2
 {
-    abstract class MachineInfoProvider
+    static class MachineInfoProvider
     {
-        public static MachineInfoProvider GetProvider()
+        public static IMachineInfoProvider GetProvider()
         {
             switch ( Environment.OSVersion.Platform )
             {
                 case PlatformID.Win32NT:
                 case PlatformID.Win32Windows:
-                    return new WindowsInfoProvider();
+                    return new WindowsMachineInfoProvider();
 
                 case PlatformID.Unix:
                     if ( Utils.IsMacOS() )
                     {
-                        return new OSXInfoProvider();
+                        return new MacOSMachineInfoProvider();
                     }
                     else
                     {
-                        return new LinuxInfoProvider();
+                        return new LinuxMachineInfoProvider();
                     }
             }
 
-            return new DefaultInfoProvider();
+            return new DefaultMachineInfoProvider();
         }
-
-        public abstract byte[] GetMachineGuid();
-        public abstract byte[] GetMacAddress();
-        public abstract byte[] GetDiskId();
     }
 
-    class DefaultInfoProvider : MachineInfoProvider
+    sealed class DefaultMachineInfoProvider : IMachineInfoProvider
     {
-        public override byte[] GetMachineGuid()
+        public static DefaultMachineInfoProvider Instance { get; } = new DefaultMachineInfoProvider();
+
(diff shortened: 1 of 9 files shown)


JustArchi (approved):

yaakov-h:
TODO:

- [ ] Add tests for background generation
- [ ] Add tests for caching value
- [ ] Add tests for thread-safety of generation/caching
- [ ] Add tests for MessageObject construction(?)

SteamKit2/SteamKit2/Util/HardwareUtils.cs
```diff
@@ -351,19 +358,20 @@ public static void Init()
...
+            var provider = (IMachineInfoProvider)state;
 
-            MachineInfoProvider provider = MachineInfoProvider.GetProvider();
+            var machineId = new MachineID();
 
-            machineId.SetBB3( GetHexString( provider.GetMachineGuid() ) );
-            machineId.SetFF2( GetHexString( provider.GetMacAddress() ) );
-            machineId.Set3B3( GetHexString( provider.GetDiskId() ) );
+            // Custom implementations can fail for any particular field, in which case we fall back to DefaultMachineInfoProvider.
+            machineId.SetBB3( GetHexString( provider.GetMachineGuid() ?? DefaultMachineInfoProvider.Instance.GetMachineGuid() ) );
```
VoiDeD, line 372: Do we want to handle providers throwing exceptions?

With this current code it's looking like this will throw when the task is `Wait`ed, which will manifest itself during `SteamUser/SteamGameServer.Logon`.

Do we want to bubble up that exception there or fall back to the default provider?

yaakov-h: I think for SteamKit internals it is best to play it safe and quietly handle failure, but for consumer-supplied implementations it should be obvious and in their face that their component exploded.

SteamKit2/SteamKit2/Util/HardwareUtils.cs
```diff
@@ -315,29 +323,40 @@ public void Set333( string value )
...
+        static ConditionalWeakTable<IMachineInfoProvider, Task<MachineID>> generationTable = new ConditionalWeakTable<IMachineInfoProvider, Task<MachineID>>();
 
-        static Task<MachineID>? generateTask;
-
-
-        public static void Init()
+        public static void Init(IMachineInfoProvider machineInfoProvider)
         {
-            generateTask = Task.Factory.StartNew( GenerateMachineID );
+            lock (machineInfoProvider)
```
VoiDeD, line 330: Locking on an object that's publicly accessible isn't ideal...

I wonder though - do we need to support initializing multiple providers? Would a single `Lazy<T>` suffice here?

Regardless I think this is fine as-is.

VoiDeD (commented):
Overall LGTM, minor nits for whitespace. The locking is probably fine but I left a comment to have that up for discussion.

SteamKit2/SteamKit2/Util/HardwareUtils.cs
yaakov-h: I did consider asking the provider for a lock object [like ICollection/Array/etc. has](https://docs.microsoft.com/en-us/dotnet/api/system.collections.icollection.syncroot?view=net-5.0) but in the end I figured it is probably unnecessary.

PR #1033, opened 2021-10-08
Add PurchaseResponseCallback and RedeemGuestPassResponseCallback

xPaw:
cc @JustArchi for review.

I didn't pull any info out of the keyvalue to leave it up to the consumer, same as the other callbacks.

Actual methods to do these actions are left as an exercise for the consumer.

### DIFF
--- a/SteamKit2/SteamKit2/Steam/Handlers/SteamApps/Callbacks.cs
+++ b/SteamKit2/SteamKit2/Steam/Handlers/SteamApps/Callbacks.cs
@@ -292,7 +292,7 @@ namespace SteamKit2
                 using ( var ms = new MemoryStream( payload ) )
                 using ( var br = new BinaryReader( ms ) )
                 {
-                    for ( int x = 0 ; x < msg.NumBans ; x++ )
+                    for ( int x = 0; x < msg.NumBans; x++ )
                     {
                         tempList.Add( br.ReadUInt32() );
                     }
@@ -481,7 +481,7 @@ namespace SteamKit2
                 /// </summary>
                 public Uri? HttpUri { get; private set; }
 
-                internal PICSProductInfo( CMsgClientPICSProductInfoResponse parentResponse, CMsgClientPICSProductInfoResponse.AppInfo app_info)
+                internal PICSProductInfo( CMsgClientPICSProductInfoResponse parentResponse, CMsgClientPICSProductInfoResponse.AppInfo app_info )
                 {
                     this.ID = app_info.appid;
                     this.ChangeNumber = app_info.change_number;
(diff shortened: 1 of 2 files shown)


JustArchi:
As those callbacks are directly tied to the requests that fire them, I believe we should also add functions for that.

```csharp
		internal async Task<RedeemGuestPassResponseCallback?> RedeemGuestPass(ulong guestPassID) {
			if (guestPassID == 0) {
				throw new ArgumentOutOfRangeException(nameof(guestPassID));
			}

			if (Client == null) {
				throw new InvalidOperationException(nameof(Client));
			}

			if (!Client.IsConnected) {
				return null;
			}

			ClientMsgProtobuf<CMsgClientRedeemGuestPass> request = new(EMsg.ClientRedeemGuestPass) {
				SourceJobID = Client.GetNextJobID(),
				Body = { guest_pass_id = guestPassID }
			};

			Client.Send(request);

			try {
				return await new AsyncJob<RedeemGuestPassResponseCallback>(Client, request.SourceJobID).ToLongRunningTask().ConfigureAwait(false);
			} catch (Exception e) {
				ArchiLogger.LogGenericException(e);

				return null;
			}
		}

		internal async Task<PurchaseResponseCallback?> RedeemKey(string key) {
			if (string.IsNullOrEmpty(key)) {
				throw new ArgumentNullException(nameof(key));
			}

			if (Client == null) {
				throw new InvalidOperationException(nameof(Client));
			}

			if (!Client.IsConnected) {
				return null;
			}

			ClientMsgProtobuf<CMsgClientRegisterKey> request = new(EMsg.ClientRegisterKey) {
				SourceJobID = Client.GetNextJobID(),
				Body = { key = key }
			};

			Client.Send(request);

			try {
				return await new AsyncJob<PurchaseResponseCallback>(Client, request.SourceJobID).ToLongRunningTask().ConfigureAwait(false);
			} catch (Exception e) {
				ArchiLogger.LogGenericException(e);

				return null;
			}
		}
```

My code from ASF. Slight modifications for SK2 would be needed.

yaakov-h:
I'm conflicted on adding the methods to kick these flows off.

On one hand we don't want to make it easy for people to come in and start brute-forcing things, or malware that redeems keys that it finds. e.g. there was something I read earlier this week about malware that scrapes people's inboxes and looks for gift cards...

On the other hand, any time somebody asks, they usually land up in ASF anyway with working sample code, so the gatekeeping is minimal at best.

We could add it with `[EditorBrowsable(Never)]` so that it is there but not suggested by Intellisense, but that might be just as pointless.

Really this should be protected on the server side anyway, which @JustArchi indicates above that Valve already seem to be doing (yay). IIRC there are also IClient interfaces for this so people could also just do this with a legitimate first-part client...

@VoiDeD what are your thoughts?

xPaw:
FYI you get rate limited for an hour after 10 (or was it 5?) invalid keys.

VoiDeD:
Just callbacks - if anything I still believe in having _some_ barrier to entry.

Issue #909, opened 2020-08-27
Use new service method emsg


xPaw:
> 19:36 <@xPaw> steamkit is using legacy ServiceMethods in unified messages
> 19:36 <@xPaw> steam uses Msg: ServiceMethodCallFromClient (151)
> 05:35 <@Netshroud> do we just need to switch ServiceMethod/ServiceMethodResponse to ServiceMethodCallFromClient/ServiceMethodSendToClient?
05:36 <@Netshroud> oh no we're using ClientServiceMethodLegacy and friends, ServiceMethod is VAC
14:46 <+rgp> ServiceMethodCallFromClient, ServiceMethodSendToClient, ServiceMethodResponse, ServiceMethod
14:47 <+rgp> and the body is now relevant proto based on target_job_name etc

xPaw:
I was looking at my gists, and saw this diff, could be useful:
```diff
diff --git a/SteamKit2/SteamKit2/Base/ClientMsg.cs b/SteamKit2/SteamKit2/Base/ClientMsg.cs
index 9d9b8b99..f4c39f69 100644
--- a/SteamKit2/SteamKit2/Base/ClientMsg.cs
+++ b/SteamKit2/SteamKit2/Base/ClientMsg.cs
@@ -72,6 +72,17 @@ namespace SteamKit2
             set => ProtoHeader.jobid_target = value ?? throw new ArgumentNullException( nameof(value) );
         }
         /// <summary>
+        /// Gets or sets the target job name for this client message.
+        /// </summary>
+        /// <value>
+        /// The target job name.
+        /// </value>
+        public string TargetJobName
+        {
+            get => ProtoHeader.target_job_name;
+            set => ProtoHeader.target_job_name = value ?? throw new ArgumentNullException( nameof( value ) );
+        }
+        /// <summary>
         /// Gets or sets the source job id for this client message.
         /// </summary>
         /// <value>
@@ -148,7 +159,7 @@ namespace SteamKit2
         /// <summary>
         /// Gets the body structure of this message.
         /// </summary>
-        public BodyType Body { get; private set; }
+        public BodyType Body { get; set; }
 
 
         /// <summary>
diff --git a/SteamKit2/SteamKit2/Steam/Handlers/SteamUnifiedMessages/SteamUnifiedMessages.cs b/SteamKit2/SteamKit2/Steam/Handlers/SteamUnifiedMessages/SteamUnifiedMessages.cs
index bfa3ce74..0acd1e79 100644
[...]

PR #1036, opened 2021-10-09
Use non-legacy service methods (and some refactoring of messages/packets handling)

xPaw:
Ref: #909

This appears to work, but the implementation is questionable at best. I haven't tested ServiceMethodSendToClient.

`GetDeserializedResponse` could be removed if responses were handled similarly to how they're done for notifications here: https://github.com/xPaw/SteamKit/blob/c3964e89564575083dfd30c354394ec13131301c/SteamKit2/SteamKit2/Steam/Handlers/SteamUnifiedMessages/SteamUnifiedMessages.cs#L182

However, it will not work if consumers have their own protobufs that SK doesn't have.

### DIFF
--- a/SteamKit2/SteamKit2/Base/ClientMsg.cs
+++ b/SteamKit2/SteamKit2/Base/ClientMsg.cs
@@ -191,8 +191,9 @@ namespace SteamKit2
 
             Header = packetMsg.Header;
 
-            using MemoryStream ms = new MemoryStream( packetMsg.GetData() );
-            ms.Seek( packetMsg.BodyOffset, SeekOrigin.Begin );
+            var data = packetMsg.GetData();
+            var offset = (int)packetMsg.BodyOffset;
+            using MemoryStream ms = new MemoryStream( data, offset, data.Length - offset );
 
             Body = Serializer.Deserialize<TBody>( ms );
 
@@ -353,8 +354,9 @@ namespace SteamKit2
 
             Header = packetMsg.Header;
 
-            using MemoryStream ms = new MemoryStream( packetMsg.GetData() );
-            ms.Seek( packetMsg.BodyOffset, SeekOrigin.Begin );
+            var data = packetMsg.GetData();
+            var offset = (int)packetMsg.BodyOffset;
+            using MemoryStream ms = new MemoryStream( data, offset, data.Length - offset );
 
             Body.Deserialize( ms );
 
@@ -506,8 +508,9 @@ namespace SteamKit2
 
             Header = packetMsg.Header;
 
-            using MemoryStream ms = new MemoryStream( packetMsg.GetData() );
-            ms.Seek( packetMsg.BodyOffset, SeekOrigin.Begin );
+            var data = packetMsg.GetData();
+            var offset = (int)packetMsg.BodyOffset;
+            using MemoryStream ms = new MemoryStream( data, offset, data.Length - offset );
 
             Body.Deserialize( ms );
 
--- a/SteamKit2/SteamKit2/Base/ClientMsg.cs
+++ b/SteamKit2/SteamKit2/Base/ClientMsg.cs
@@ -150,7 +150,7 @@ namespace SteamKit2
         /// <summary>
         /// Gets the body structure of this message.
         /// </summary>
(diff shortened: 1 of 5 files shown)


xPaw:
FYI you linked #909, but that issue talks about GC messages which this PR doesn't touch.

SteamKit2/SteamKit2/Base/PacketBase.cs
```diff
@@ -95,15 +102,28 @@ public sealed class PacketClientMsgProtobuf : IPacketMsg
...
-        public ulong SourceJobID { get; }
-
+        public ulong SourceJobID => Header.Proto.jobid_source;
+        /// <summary>
+        /// Gets the header for this packet message.
+        /// </summary>
+        /// <value>
+        /// The header.
+        /// </value>
+        public MsgHdrProtoBuf Header { get; }
```
yaakov-h, line 119: I'm wondering if this should be `internal`. Previously immutable properties are now mutable through this.

SteamKit2/SteamKit2/Base/ClientMsg.cs
```diff
@@ -515,7 +499,26 @@ public Msg( IPacketMsg msg )
...
-            Deserialize( msg.GetData() );
+            if ( !( msg is PacketMsg packetMsg ) )
+            {
+                throw new InvalidDataException( $"ClientMsg<{typeof( TBody ).FullName}> used for proto message!" );
+            }
+
+            Header = packetMsg.Header;
+
+            using MemoryStream ms = new MemoryStream( packetMsg.GetData() );
+            ms.Seek( packetMsg.BodyOffset, SeekOrigin.Begin );
```
yaakov-h, line 510: There's an overload of the `MemoryStream` constructor that takes an offset.

SteamKit2/SteamKit2/Base/PacketBase.cs
```diff
@@ -173,14 +189,28 @@ public sealed class PacketClientMsg : IPacketMsg
...
         /// </value>
-        public ulong SourceJobID { get; }
+        public ulong SourceJobID => Header.SourceJobID;
+        /// <summary>
+        /// Gets the header for this packet message.
+        /// </summary>
+        /// <value>
+        /// The header.
+        /// </value>
+        public ExtendedClientMsgHdr Header { get; }
```
yaakov-h, line 206: As above, this now makes certain properties mutable.

SteamKit2/SteamKit2/Base/PacketBase.cs
```diff
@@ -250,14 +276,28 @@ public sealed class PacketMsg : IPacketMsg
...
         /// </value>
-        public ulong SourceJobID { get; }
+        public ulong SourceJobID => Header.SourceJobID;
+        /// <summary>
+        /// Gets the header for this packet message.
+        /// </summary>
+        /// <value>
+        /// The header.
+        /// </value>
+        public MsgHdr Header { get; }
```
yaakov-h, line 293: As above, this makes certain properties mutable.

SteamKit2/SteamKit2/Base/ClientMsg.cs
```diff
@@ -190,9 +184,26 @@ public ClientMsgProtobuf( EMsg eMsg, MsgBase<MsgHdrProtoBuf> msg, int payloadRes
...
+            if ( !( msg is PacketClientMsgProtobuf packetMsg ) )
+            {
+                throw new InvalidDataException( $"ClientMsgProtobuf<{typeof(TBody).FullName}> used for non-proto message!" );
+            }
+
+            Header = packetMsg.Header;
 
-            Deserialize( msg.GetData() );
+            using MemoryStream ms = new MemoryStream( packetMsg.GetData() );
+            ms.Seek( packetMsg.BodyOffset, SeekOrigin.Begin );
```
yaakov-h, line 195: There's an overload of the MemoryStream constructor that takes an offset.

yaakov-h (commented):
Confirmed that service methods and service notifications are working with this.

Not sure I like making the header object public though, see comments on that. Perhaps they should be `internal`.

PR #1037, opened 2021-10-10
Remove some of the old messages

xPaw:
MsgClientLoggedOff still happens if you connect and don't login after a minute. (result is Fail).

OGS messages, guest pass list, chat messages, MsgClientServerUnavailable, MsgClientMarketingMessageUpdate2 are still used.

SK handles `MsgClientLogOnResponse`, but I don't know when that can happen.


yaakov-h:
I don't know if there were any non-proto GC messages, from memory (quite hazy over 10+ years)I think GC was introduced around the same time as the protobuf migration. Perhaps in the very original TF2 implementation?

It might be worth checking steamclient to see if it [still] has any code to handle non-proto GC messages.

PR #1047, opened 2021-11-10
Catch BeginConnect exceptions in TcpConnection

xPaw:
https://docs.microsoft.com/en-us/dotnet/api/system.net.sockets.socket.beginconnect

I've had my program crash with SocketException (albeit with no useful stacktrace directly pointing to this line), and looking at the other connection failures above it appears to be this.

### DIFF
--- a/SteamKit2/SteamKit2/Networking/Steam3/TcpConnection.cs
+++ b/SteamKit2/SteamKit2/Networking/Steam3/TcpConnection.cs
@@ -104,19 +104,19 @@ namespace SteamKit2
             // If we have no cancellation token source, we were already Release()'ed
             if (cancellationToken?.IsCancellationRequested ?? true)
             {
-                log.LogDebug("TcpConnection", "Connection request to {0} was cancelled", CurrentEndPoint);
+                log.LogDebug( nameof( TcpConnection ), "Connection request to {0} was cancelled", CurrentEndPoint );
                 if (success) Shutdown();
                 Release( userRequestedDisconnect: true );
                 return;
             }
             else if (!success)
             {
-                log.LogDebug( "TcpConnection", "Timed out while connecting to {0}", CurrentEndPoint);
+                log.LogDebug( nameof( TcpConnection ), "Failed connecting to {0}", CurrentEndPoint );
                 Release( userRequestedDisconnect: false );
                 return;
             }
 
-            log.LogDebug( "TcpConnection", "Connected to {0}", CurrentEndPoint);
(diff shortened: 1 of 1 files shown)


yaakov-h:
> .NET Framework and .NET 5 and earlier only: An error occurred when attempting to access the socket.

You could fix it by upgrading to .NET 6 😂

SteamKit2/SteamKit2/Networking/Steam3/TcpConnection.cs
```diff
@@ -152,12 +152,24 @@ private void TryConnect(object sender)
...
+
+            IAsyncResult asyncResult;
+
+            try
+            {
+                asyncResult = socket!.BeginConnect( CurrentEndPoint, null, null );
+            }
+            catch ( Exception ex )
+            {
+                log.LogDebug( nameof( TcpConnection ), "Socket exception while beginning connection request to {0}: {1}", CurrentEndPoint, ex );
```
yaakov-h, line 168: We're catching any old `Exception` here, so it might not necessarily be a `"Socket exception..."`.

xPaw:
> You could fix it by upgrading to .NET 6 😂

This was on NET 6

xPaw:
After brief digging, this appears to be a .NET6 regression. So this PR likely won't fix it, but it's still useful as BeginConnect can throw other exceptions.

Here's the full trace:
```
Unhandled exception. System.AggregateException: A Task's exception(s) were not observed either by Waiting on the Task or accessing its Exception property. As a result, the unobserved exception was rethrown by the finalizer thread. (Operation canceled) (Operation canceled)
 ---> System.Net.Sockets.SocketException (125): Operation canceled
   at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.ThrowException(SocketError error, CancellationToken cancellationToken)
   at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.System.Threading.Tasks.Sources.IValueTaskSource.GetResult(Int16 token)
   at System.Threading.Tasks.ValueTask.ValueTaskSourceAsTask.<>c.<.cctor>b__4_0(Object state)
   --- End of inner exception stack trace ---
   at System.AggregateException.Handle(Func`2 predicate)
   at SteamDatabaseBackend.Bootstrapper.OnUnobservedTaskException(Object sender, UnobservedTaskExceptionEventArgs args) in /home/steamdb/deploy/Bootstrapper.cs:line 149
   at System.Threading.Tasks.TaskScheduler.PublishUnobservedTaskException(Object sender, UnobservedTaskExceptionEventArgs ueea)
   at System.Threading.Tasks.TaskExceptionHolder.Finalize()
```

PR #1050, opened 2021-11-13
Fix unobserved task exception under .NET 6 runtime

yaakov-h:
Fixes bug identified by #1047 / https://github.com/dotnet/runtime/issues/61411 in what seems to be the only practical approach that doesn't require targeting the .NET 6.0 TFM.

### DIFF
--- a/SteamKit2/SteamKit2/Networking/Steam3/TcpConnection.cs
+++ b/SteamKit2/SteamKit2/Networking/Steam3/TcpConnection.cs
@@ -159,7 +159,10 @@ namespace SteamKit2
 
             try
             {
-                using ( cancellationToken.Token.Register( s => ( ( Socket )s ).Dispose(), socket ) )
+                using var timeoutTokenSource = new CancellationTokenSource( timeout );
+                using var connectCancellation = CancellationTokenSource.CreateLinkedTokenSource( cancellationToken.Token, timeoutTokenSource.Token );
+
+                using ( connectCancellation.Token.Register( s => ( ( Socket )s ).Dispose(), socket ) )
                 {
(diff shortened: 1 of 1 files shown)


SteamKit2/SteamKit2/Networking/Steam3/TcpConnection.cs
```diff
@@ -157,36 +157,29 @@ private void TryConnect(object sender)
...
-                catch ( Exception ex )
-                {
-                    log.LogDebug( nameof( TcpConnection ), "Exception while completing connection request to {0}: {1}", CurrentEndPoint, ex );
-                    ConnectCompleted( false );
-                }
-            }
-            else
+            catch ( Exception ex )
             {
+                log.LogDebug( nameof( TcpConnection ), "Exception while beginning connection request to {0}: {1}", CurrentEndPoint, ex );
```
xPaw, line 177:
```suggestion
                log.LogDebug( nameof( TcpConnection ), "Exception while connecting to {0}: {1}", CurrentEndPoint, ex );

```

SteamKit2/SteamKit2/Networking/Steam3/TcpConnection.cs
```diff
@@ -157,36 +157,29 @@ private void TryConnect(object sender)
...
                 ConnectCompleted( false );
+                return;
             }
+
+            ConnectCompleted( true );
```
xPaw, line 182:
```suggestion
            ConnectCompleted( socket.Connected );
```

Would this be saner/more correct?

xPaw:
While you're at it, remove `int timeout = (int)sender;` cast

xPaw (approved):

PR #1053, opened 2021-11-14
Fix up userInitiated in various disconnect calls

xPaw:


SteamKit2/SteamKit2/Steam/CMClient.cs
```diff
@@ -233,7 +233,7 @@ public void Connect( ServerRecord? cmServer = null )
         /// </summary>
         public void Disconnect() => Disconnect( userInitiated: true ); 
 
-        void Disconnect( bool userInitiated)
+        internal void Disconnect( bool userInitiated )
```
yaakov-h, line 236: Does `protected` work here instead?

xPaw: It does, but then anyone extending CMClient can see it, do you actually want that?

yaakov-h: hmm, how about “private protected” then?

PR #1055, opened 2021-11-14
Use UriBuilder for websocket uris; fix ipv6

xPaw:
IPv6 addresses would produce: `[[::1]:1337]:1337`.


yaakov-h:
Are there v6 CMs that I can test this against?

Not that my home internet has IPv6 in the first place...

xPaw:
I don't think Steam does IPv6 yet, and even if it did it likely would be with a dns name.

yaakov-h:
well that's some handy magic.

PR #1075, opened 2022-01-06
Fix sending service notifications

xPaw:
- Fix notification servicemethods not working
- Add SendNotification and deprecate SendMessage(..., true)

This was broken in #1036, `ServiceMethodSendToClient` was not the correct call. Notifications are sent by Steam by not setting the source jobid.

This logic can be verified by looking at the new chat's javascript. it has `SendMsg`/`SendNotification` on service transport. `SendMsg` uses `SendMsgAndAwaitResponse` which sets the `jobid_source`.

@Netshroud which notification message did you test?

### DIFF
--- a/SteamKit2/SteamKit2/Steam/Handlers/SteamUnifiedMessages/SteamUnifiedMessages.cs
+++ b/SteamKit2/SteamKit2/Steam/Handlers/SteamUnifiedMessages/SteamUnifiedMessages.cs
@@ -64,7 +64,7 @@ namespace SteamKit2
             /// <param name="isNotification">Whether this message is a notification or not.</param>
             /// <returns>The JobID of the request. This can be used to find the appropriate <see cref="ServiceMethodResponse"/>.</returns>
             [Obsolete( "Use SendNotification() instead of passing 'true' bool in SendMessage. SendMessage incorrectly returned AsyncJob for notifications, they have no response by design." )]
-            public AsyncJob<ServiceMethodResponse>? SendMessage<TResponse>( Expression<Func<TService, TResponse>> expr, bool isNotification = false )
+            public AsyncJob<ServiceMethodResponse>? SendMessage<TResponse>( Expression<Func<TService, TResponse>> expr, bool isNotification )
             {
                 return SendMessageOrNotification( expr, isNotification );
             }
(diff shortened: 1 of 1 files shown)


JustArchi:
I helped.

JustArchi (approved):
Tested `SendNotification()` with below code as of 84f4049598ed34af9f923b5c1f4708c4913e9607 - works.

```csharp
		CFriendMessages_AckMessage_Notification request = new() {
			steamid_partner = steamID,
			timestamp = timestamp
		};

		UnifiedFriendMessagesService.SendNotification(x => x.AckMessage(request));
```

JustArchi:
Also tested `SendMessage()` that was `isNotification: false`, also works.

SteamKit2/SteamKit2/Steam/Handlers/SteamUnifiedMessages/SteamUnifiedMessages.cs
```diff
@@ -64,14 +94,24 @@ public AsyncJob<ServiceMethodResponse> SendMessage<TResponse>( Expression<Func<T
...
+                {
+                    var notification = typeof( SteamUnifiedMessages ).GetMethod( nameof( SteamUnifiedMessages.SendNotification ) )!.MakeGenericMethod( message.GetType() );
+                    notification.Invoke( this.steamUnifiedMessages, new[] { rpcName, message } );
+                    return null;
+                }
+
+                // When the obsolete overload is removed, then this code be used, instead of GetMethods().
+                // var method = typeof( SteamUnifiedMessages ).GetMethod( nameof( SteamUnifiedMessages.SendMessage ) )!.MakeGenericMethod( message.GetType() );
+
+                var method = typeof( SteamUnifiedMessages ).GetMethods().First( x => x.Name == nameof( SteamUnifiedMessages.SendMessage ) ).MakeGenericMethod( message.GetType() );
```
yaakov-h, line 113: `First` isn't necessarily safe here.

@JustArchi can you please retest with this?:

```csharp
var method = sendMessageMethod.MakeGenericMethod( message.GetType() );
```

and then outside the method definition:

```csharp
static readonly MethodInfo sendMessageMethod = typeof( SteamUnifiedMessages )
    .GetMethods( BindingFlags.Public | BindingFlags.Instance )
    .Single( m => m is { Name: nameof( SteamUnifiedMessages.SendMessage ) } && !Attribute.IsDefined( m, typeof( ObsoleteAttribute ) ) );
```

xPaw: Changed it to this, it works.

PR #1082, opened 2022-04-01
TuxHook2: cross-platform NetHook2 port

m4dEngi:
Here be dragons


JustArchi (approved):
This is the best thing that happened this year so far :sunglasses:

Resources/TuxHook2/inject.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
JustArchi, line 1: `#!/usr/bin/env sh`, from what I see you don't use any bash extensions. POSIX compatibility ftw!

I'd also append `set -eu`.

Resources/TuxHook2/inject.sh
```diff
@@ -0,0 +1,10 @@
...
+STEAM_PID=$(ps -C steam -o pid= | head -1)
```
JustArchi, line 2: Why not `pidof steam`?

m4dEngi:
Won't pass checks, because deps are excluded from ALL.

Resources/TuxHook2/inject.sh
m4dEngi: Dunno, i'm no bash expert.

m4dEngi: I also forgot to add passing PID by argument...

xPaw:
Can you add a build to https://github.com/SteamRE/SteamKit/blob/master/.github/workflows/steamkit2-build.yaml

It also seems that CI for Nethook2 broke.

Resources/TuxHook2/readme.md
```diff
@@ -0,0 +1,30 @@
...
+2. cmake ../
+3. make deps
+4. make
+5. chmod +x ./inject.sh
+
+## Usage
+### Windows
+~~Net~~TuxHook is capable of self injecting and ejecting from running instances of Steam, so there's no requirement to use a separate loader such as winject.
+### Linux
+Use supplied inject.sh to inject ~~Net~~TuxHook into running steam process.
```
xPaw, line 30: Should probably mention that gdb is required.

yaakov-h:
Initial thoughts:
- This should probably be an extension of NetHook2 rather than a clone. I don't want two to maintain. It also makes it hard to see what changes you've made between NetHook2 and TuxHook.
- This should be included in CI.
- Can we use LD_PRELOAD instead of / as well as GDB?
- Can we detect GDB or LLDB and use the appropriate one?
- Does this work on macOS? From what I can tell the hooks library you're using should support x86_64, but Apple have done some extra security work on aarch64/arm64 to prevent this.

Resources/TuxHook2/inject.sh
yaakov-h: And other processes (srcds et al)?

m4dEngi:
> Initial thoughts:
> 
> * This should probably be an extension of NetHook2 rather than a clone. I don't want two to maintain. It also makes it hard to see what changes you've made between NetHook2 and TuxHook.
> * This should be included in CI.
> * Can we use LD_PRELOAD instead of / as well as GDB?
> * Can we detect GDB or LLDB and use the appropriate one?
> * Does this work on macOS? From what I can tell the hooks library you're using should support x86_64, but Apple have done some extra security work on aarch64/arm64 to prevent this.
> * Can this work on 64-bit Windows too?

- And miss opportunity to make a linux related pun on 1st of april?!
- I have no idea how to work with github actions
- We can, it'll just be a little bit more messy and require bit more work than injection via gdb
- If lldb is capable to call __libc_dlopen in target process i don't see why not
- Will not work on arm/aarch64 due to platform limitations as stated in funchook library description.
- It can, but will require separate implementation like linux version for obvious reasons.

xPaw:
Here's CI config, took me a while to get it working:
```yaml
  nethook2:
    name: NetHook2 on ${{ matrix.os }}

    strategy:
      matrix:
        os: [ windows-latest, ubuntu-latest ]
      fail-fast: false

    runs-on: ${{ matrix.os }}

    steps:
    - name: Checkout
      uses: actions/checkout@v2

    - name: Setup developer command prompt (Windows)
      if: matrix.os == 'windows-latest'
      uses: ilammy/msvc-dev-cmd@v1
      with:
        arch: win32

    - name: Install dependencies (Linux)
      if: matrix.os == 'ubuntu-latest'
      run: |
        sudo apt update
        sudo apt install -y g++-multilib libc++-11-dev libc++abi-11-dev

    - name: Create build directory
      run: mkdir Resources/TuxHook2/build

    - name: Generate make file (Windows)
      if: matrix.os == 'windows-latest'
      working-directory: Resources/TuxHook2/build/
      run: cmake -A Win32 ../

    - name: Generate make file (Linux)
      if: matrix.os == 'ubuntu-latest'
      working-directory: Resources/TuxHook2/build/
      run: cmake ../

    - name: Build NetHook2
      working-directory: Resources/TuxHook2/build/
      run: |
        cmake --build . --target deps --config Release
        cmake --build . --config Release

    - name: Tar files (Linux)
      if: matrix.os == 'ubuntu-latest'
      working-directory: Resources/TuxHook2/build/
      run: tar -cvf NetHook2-Linux.tar inject.sh *.so

    - name: Upload NetHook2 (Linux)
      uses: actions/upload-artifact@v3
      if: matrix.os == 'ubuntu-latest'
      with:
        name: NetHook2-Linux
        path: Resources/TuxHook2/build/NetHook2-Linux.tar

    - name: Upload NetHook2 (Windows)
      uses: actions/upload-artifact@v3
      if: matrix.os == 'windows-latest'
      with:
        name: NetHook2
        path: 'Resources/TuxHook2/build/Release/*.dll'
```

things to do:
- `git update-index --chmod=+x inject.sh`
- ldr on linux doesn't build with cmake for some reason
- do not build protobuf compiler
- warning C4530: C++ exception handler used, but unwind semantics are not enabled. Specify /EHsc

PR #1102, opened 2022-05-20
Add extra asserts connection logic

xPaw:


SteamKit2/SteamKit2/Steam/CMClient.cs
```diff
@@ -212,26 +212,32 @@ public void Connect( ServerRecord? cmServer = null )
...
-              {
+                {
                     if ( t.IsFaulted )
                     {
-                      LogDebug( nameof( CMClient ), "Unhandled exception when attempting to connect to Steam: {0}", t.Exception );
+                        LogDebug( nameof( CMClient ), "Unhandled exception when attempting to connect to Steam: {0}", t.Exception );
                         OnClientDisconnected( userInitiated: false );
                     }
+
+                    connectionSetupTask = null;
```
xPaw, line 232: I think that's the right place to put it? We don't want this task reference to linger after it's done.

azuisleet: This can't be done as Disconnect may be trying to access it.

xPaw: This would be nulled after the task is already finished its work, does it matter if disconnect accesses it?

SteamKit2/SteamKit2/Steam/CMClient.cs
```diff
@@ -212,26 +212,35 @@ public void Connect( ServerRecord? cmServer = null )
...
-              {
+                {
                     if ( t.IsFaulted )
                     {
-                      LogDebug( nameof( CMClient ), "Unhandled exception when attempting to connect to Steam: {0}", t.Exception );
+                        LogDebug( nameof( CMClient ), "Unhandled exception when attempting to connect to Steam: {0}", t.Exception );
                         OnClientDisconnected( userInitiated: false );
                     }
+
+                    lock ( connectionLock )
```
azuisleet, line 232: This will result in a deadlock with Disconnect, as it holds connectionLock to await this Task

PR #1109, opened 2022-06-20
Add PlayingSessionStateCallback

xPaw:


SteamKit2/SteamKit2/Steam/Handlers/SteamUser/Callbacks.cs
```diff
@@ -548,5 +548,28 @@ internal MarketingMessageCallback( MsgClientMarketingMessageUpdate2 body, byte[]
                 Messages = new ReadOnlyCollection<Message>( msgList );
             }
         }
+
+        /// <summary>
+        /// This callback is received when another client starts or stops playing a game.
+        /// While blocked, sending ClientGamesPlayed message will log you off with LoggedInElsewhere result.
```
yaakov-h, line 554: blocked?

JustArchi: `PlayingBlocked`

JustArchi (approved):

JustArchi (approved):
Using it in ASF myself, LGTM.

SteamKit2/SteamKit2/Steam/Handlers/SteamUser/Callbacks.cs
yaakov-h: but what does that _mean_?

JustArchi: @yaakov-h when one Steam client starts a game, it holds something I like to call "playing lock" which is transmitted in this callback as `PlayingBlocked`.

If another Steam client (connected to the same account of course) attempts to start a game in this state, it'll get forcefully disconnected with `LoggedInElsewhere` error. This callback therefore is a way for SK2 bots to know whether or not they can start playing a game without getting forcefully disconnected in case they can't.

Hope this is the explanation you were looking for. If not let me know and I'll try to elaborate further.

yaakov-h (approved):

PR #1124, opened 2022-08-26
Replace CA2101 suppression with disabling BestFitMapping

yaakov-h:
Instead of suppressing the rule, fix the undesired behaviour as suggested here: https://github.com/dotnet/roslyn-analyzers/issues/2886#issuecomment-1079426722


yaakov-h:
ugh, unfortunately this still triggers the rule.

xPaw:
Can anything be done here now that we target .NET 8?

PR #1126, opened 2022-08-28
Send client hello; support non-authed service methods

xPaw:
Hello is required for non-authed service methods to work (otherwise it drops connection).

Non authed service method is required for CAuthentication messages.


xPaw:
Sending client hello with the protocol version may have changed some semantics of messages like log offs or disconnects, because the were non-proto specifically due to the server not knowing the version.

psychonic (approved):

PR #1129, opened 2022-08-31
Add SteamAuthentication to handle new Steam login flow sessions

### LINKED ISSUE #1125: New Steam login flow
Making this issue for collecting observations.

Implementation in JS: https://github.com/DoctorMcKay/node-steamcommunity/pull/292

Even Steam client uses this flow, and it no longer uses login keys for remembering password, instead it uses new `access_token` field in the login message. There's also `refresh_token`, but I don't know how will the client decide when it needs to refresh the token.

Their tokens (and cookies) are JWT tokens, you can base64 decode it and see its expiration.

For web, the jwt token expires in a day, and when it does it will redirect to login.steampowered.com which has a separate JWT token cookie set (if remember password, it expires in 207 days).

When auth has confirmation_type 6, it will do a call to /jwt/checkdevice. When login. subdomain has steamMachineAuth cookie set, this will return true and it will avoid asking for an email code.

The protos: https://github.com/SteamDatabase/Protobufs/blob/master/steam/steammessages_auth.steamclient.proto

xPaw:
Fixes #1125.

This is a work in progress. When beginning an auth session it returns a new object which some methods to do stuff on the session.

It adds a new IAuthenticator class to provide 2FA codes (with an example implementation that just reads from the console).

What I am not entirely certain about is the polling design (right now it's just a loop with task.delay).


Samples/1a.Authentication/Program.cs
```diff
@@ -0,0 +1,108 @@
...
+
+    Console.WriteLine( $"QR Link: {authSession.ChallengeURL}" );
+    */
+
+    // Begin authenticating via credentials
+    var authSession = await auth.BeginAuthSessionViaCredentials( new SteamAuthentication.AuthSessionDetails
+    {
+        Username = user,
+        Password = pass,
+        Persistence = SteamKit2.Internal.ESessionPersistence.k_ESessionPersistence_Ephemeral,
```
yaakov-h, line 64: It doesn't seem right that a fairly core logon API needs to use `SteamKit2.Internal`

SteamKit2/SteamKit2/Steam/Handlers/SteamAuthentication/SteamAuthentication.cs
```diff
@@ -0,0 +1,519 @@
...
+        /// <param name="packetMsg">The packet message that contains the data.</param>
+        public override void HandleMsg( IPacketMsg packetMsg )
+        {
+            // not used
+        }
+
+        private static List<CAuthentication_AllowedConfirmation> SortConfirmations( List<CAuthentication_AllowedConfirmation> confirmations )
+        {
+            /*
+            valve's preferred order:
```
yaakov-h, line 487: This comment isn't helping me understand what this function does.

Samples/1a.Authentication/Program.cs
xPaw: I changed it to a bool.

SteamKit2/SteamKit2/Steam/Handlers/SteamAuthentication/SteamAuthentication.cs
```diff
@@ -40,24 +41,31 @@ public sealed class AuthSessionDetails
...
+
             /// <summary>
             /// Gets or sets the session persistence.
             /// </summary>
             /// <value>The persistence.</value>
-            public bool IsPersistentSession { get; set; } = true;
+            public bool IsPersistentSession { get; set; } = false;
 
             /// <summary>
             /// Gets or sets the website id that the login will be performed for. (EMachineAuthWebDomain)
```
yaakov-h, line 64: Can we not declare this as an EMachineAuthWebDomain?

xPaw:
Logging in as SteamClient requires to do it over CM. I tried over webapi, but it throws InvalidParam, and using other types will give InvalidPassword when trying to logon.

This means that CM will kick you in 60 seconds if you don't auth. Steam client also has this issue, but it will reconnect and just continue polling.

I don't want to deal with graceful reconnections in the scope of this PR.

JustArchi (approved):
I tested this implementation in ASF, works good enough in my opinion to be released as part of SK2 alpha release: https://github.com/JustArchiNET/ArchiSteamFarm/pull/2843

xPaw:
Current code should good enough to release an alpha version and get this tested by consumers. We can improve the implementation and specifics in the future.

Adios22:
When alpha version will be available to download?

psychonic (approved):

yaakov-h (changes requested):

SteamKit2/SteamKit2/Steam/Handlers/SteamAuthentication/IAuthenticator.cs
```diff
@@ -0,0 +1,34 @@
...
+    /// 
+    /// </summary>
+    public interface IAuthenticator
+    {
+        /// <summary>
+        /// This method is called when the account being logged into requires 2-factor authentication using the authenticator app.
+        /// </summary>
+        /// <param name="previousCodeWasIncorrect">True when previously provided code was incorrect.</param>
+        /// <returns>The 2-factor auth code used to login. This is the code that can be received from the authenticator app.</returns>
+        public Task<string> ProvideDeviceCode( bool previousCodeWasIncorrect );
```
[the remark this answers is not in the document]
xPaw: Reporting an error to the user, or handle whatever else they need (e.g. ASF needed this because they handle totp)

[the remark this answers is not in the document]
xPaw: I didn't pass in cancellation because whoever starts authentication can pass the cancellation token into their own authenticator directly

SteamKit2/SteamKit2/Steam/Handlers/SteamAuthentication/SteamAuthentication.cs
```diff
@@ -0,0 +1,569 @@
...
+            /// </summary>
+            public ulong ClientID { get; set; }
+            /// <summary>
+            /// Unique request ID to be presented by requestor at poll time.
+            /// </summary>
+            public byte[] RequestID { get; set; }
+            /// <summary>
+            /// Confirmation types that will be able to confirm the request.
+            /// </summary>
+            public List<CAuthentication_AllowedConfirmation> AllowedConfirmations { get; set; }
```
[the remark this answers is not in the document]
xPaw: This can be made internal, but I wanted to expose it for any consumers if they wanted to do something custom.

SteamKit2/SteamKit2/Steam/Handlers/SteamAuthentication/SteamAuthentication.cs
```diff
@@ -0,0 +1,569 @@
...
+            return authResponse;
+        }
+
+        /// <summary>
+        /// Handles a client message. This should not be called directly.
+        /// </summary>
+        /// <param name="packetMsg">The packet message that contains the data.</param>
+        public override void HandleMsg( IPacketMsg packetMsg )
+        {
+            // not used
```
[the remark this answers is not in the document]
xPaw: Mostly to get it from the handler with client instance. Do you have a better idea?

SteamKit2/SteamKit2/Steam/Handlers/SteamAuthentication/AuthenticationException.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+
+namespace SteamKit2
+{
+    /// <summary>
+    /// Thrown when <see cref="SteamAuthentication"/> fails to authenticate.
+    /// </summary>
+    public class AuthenticationException : Exception
+    {
```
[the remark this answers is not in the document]
xPaw: Nothing in SK does this.

SteamKit2/SteamKit2/Steam/Handlers/SteamAuthentication/SteamAuthentication.cs
```diff
@@ -0,0 +1,569 @@
...
+            /// </summary>
+            /// <value>The guard data.</value>
+            public string? GuardData { get; set; }
+
+            /// <summary>
+            /// Authenticator object which will be used to handle 2-factor authentication if necessary.
+            /// Use <see cref="UserConsoleAuthenticator"/> for a default implementation.
+            /// </summary>
+            /// <value>The authenticator object.</value>
+            public IAuthenticator? Authenticator { get; set; }
```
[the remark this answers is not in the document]
xPaw: I don't know, it seems fine for me here.

SteamKit2/SteamKit2/Steam/Handlers/SteamAuthentication/SteamAuthentication.cs
```diff
@@ -0,0 +1,569 @@
...
+            var publicKey = await GetPasswordRSAPublicKey( details.Username!, authenticationService );
+            var rsaParameters = new RSAParameters
+            {
+                Modulus = Utils.DecodeHexString( publicKey.publickey_mod ),
+                Exponent = Utils.DecodeHexString( publicKey.publickey_exp ),
+            };
+
+            using var rsa = RSA.Create();
+            rsa.ImportParameters( rsaParameters );
+            var encryptedPassword = rsa.Encrypt( Encoding.UTF8.GetBytes( details.Password ), RSAEncryptionPadding.Pkcs1 );
```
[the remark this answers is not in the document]
xPaw: No.

SteamKit2/SteamKit2/Steam/Handlers/SteamAuthentication/SteamAuthentication.cs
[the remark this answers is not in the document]
xPaw: Ok I changed it to `var auth = new SteamAuthentication( steamClient );`, and moved it into its own namespace.

SteamKit2/SteamKit2/Steam/Handlers/SteamAuthentication/IAuthenticator.cs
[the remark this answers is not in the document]
JustArchi: Case in point, we have 2FA tokens generator, this way I can come to conclusion that I've used credentials and the code was incorrect - so in my case, I reset the time back to Steam servers, other option might include telling user that his 2FA credentials are wrong for example.

SteamKit2/SteamKit2/Steam/Handlers/SteamAuthentication/AuthenticationException.cs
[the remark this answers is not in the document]
yaakov-h: `BerDecodeException` does. :D

Created #1207 to sort this out for the whole project.

SteamKit2/SteamKit2/Steam/Handlers/SteamAuthentication/SteamAuthentication.cs
```diff
@@ -117,31 +117,33 @@ internal AuthPollResult( CAuthentication_PollAuthSessionStatus_Response response
         /// </summary>
         public class AuthSession
         {
-            internal SteamUnifiedMessages.UnifiedService<IAuthentication> AuthenticationService { get; private set; }
+            internal SteamAuthentication Authentication { get; }
```
yaakov-h: Can/should this be public?

SteamKit2/SteamKit2/Steam/Handlers/SteamAuthentication/SteamAuthentication.cs
```diff
@@ -169,7 +171,7 @@ public async Task<AuthPollResult> StartPolling( CancellationToken? cancellationT
                 // simply poll until confirmation is accepted, or whether they want to fallback to the next preferred confirmation type.
                 if ( Authenticator != null && preferredConfirmation.confirmation_type == EAuthSessionGuardType.k_EAuthSessionGuardType_DeviceConfirmation )
                 {
-                    var prefersToPollForConfirmation = await Authenticator.AcceptDeviceConfirmation();
+                    var prefersToPollForConfirmation = await Authenticator.AcceptDeviceConfirmationAsync();
```
yaakov-h: All of these `await` calls should have `.ConfigureAwait(false)` so that .NET Framework consumers don't end up deadlocking on a UI thread or ASP.NET execution context.

Not sure if/how it works on WinForms/WPF Core or MAUI.

SteamKit2/SteamKit2/Steam/Handlers/SteamAuthentication/SteamAuthentication.cs
```diff
@@ -155,7 +157,7 @@ internal AuthSession( SteamUnifiedMessages.UnifiedService<IAuthentication> authe
             /// <returns>An object containing tokens which can be used to login to Steam.</returns>
             /// <exception cref="InvalidOperationException">Thrown when an invalid state occurs, such as no supported confirmation methods are available.</exception>
             /// <exception cref="AuthenticationException">Thrown when polling fails.</exception>
-            public async Task<AuthPollResult> StartPolling( CancellationToken? cancellationToken = null )
+            public async Task<AuthPollResult> PollingWaitForResultAsync( CancellationToken? cancellationToken = null )
```
yaakov-h: The standard convention is `CancellationToken cancellationToken = default`.

`default` is the same as `CancellationToken.None`.

This removes the need for null-checks further down.

SteamKit2/SteamKit2/Steam/Handlers/SteamAuthentication/SteamAuthentication.cs
```diff
@@ -281,13 +283,16 @@ public async Task<AuthPollResult> StartPolling( CancellationToken? cancellationT
 
                 while ( true )
                 {
-                    cancellationToken?.ThrowIfCancellationRequested();
-
-                    await Task.Delay( PollingInterval );
-
-                    cancellationToken?.ThrowIfCancellationRequested();
+                    if( cancellationToken is CancellationToken nonNullCancellationToken )
```
yaakov-h: This should fade away after making CancellationToken non-nullable as mentioned above.

SteamKit2/SteamKit2/Steam/Handlers/SteamAuthentication/SteamAuthentication.cs
```diff
@@ -297,19 +302,19 @@ public async Task<AuthPollResult> StartPolling( CancellationToken? cancellationT
...
+            public async Task<AuthPollResult?> PollAuthSessionStatusAsync()
             {
                 var request = new CAuthentication_PollAuthSessionStatus_Request
                 {
                     client_id = ClientID,
                     request_id = RequestID,
                 };
 
-                var message = await AuthenticationService.SendMessage( api => api.PollAuthSessionStatus( request ) );
+                var message = await Authentication.AuthenticationService!.SendMessage( api => api.PollAuthSessionStatus( request ) );
```
yaakov-h: It seems odd that this could/would be null.

SteamKit2/SteamKit2/Steam/Handlers/SteamAuthentication/SteamAuthentication.cs
```diff
@@ -337,6 +333,14 @@ public async Task<AuthPollResult> StartPolling( CancellationToken? cancellationT
 
                 return null;
             }
+
+            internal virtual void HandlePollAuthSessionStatusResponse( CAuthentication_PollAuthSessionStatus_Response response)
```
yaakov-h: why not `protected`?

SteamKit2/SteamKit2/Steam/Handlers/SteamAuthentication/SteamAuthentication.cs
```diff
@@ -437,6 +453,9 @@ private async Task<CAuthentication_GetPasswordRSAPublicKey_Response> GetPassword
         /// <param name="details">The details to use for logging on.</param>
         public async Task<QrAuthSession> BeginAuthSessionViaQR( AuthSessionDetails details )
         {
+            var unifiedMessages = Client.GetHandler<SteamUnifiedMessages>()!;
+            AuthenticationService = unifiedMessages.CreateService<IAuthentication>();
```
yaakov-h: Can we push this to the constructor?

yaakov-h (changes requested):
There are also still obsoletion warnings in the build that need to be taken care of (probably just suppressed).

SteamKit2/SteamKit2/Steam/Handlers/SteamAuthentication/SteamAuthentication.cs
xPaw: I don't think there is a good reason to expose this.

yaakov-h: It seems like it could be private/protected then, rather than internal?

PR #1145, opened 2022-10-20
Fix nethook builds not to break on new zlib releases

xPaw:
`tar` is available on Windows since Build 17063, unfortunately there is no zip archive in the fossils folder.


m4dEngi:
Maybe it's better to use latest version ( [1.2.13](https://zlib.net/zlib1213.zip) ) instead since it was just a CVE/Bugfix release?

xPaw:
Sure if it works, I only fixed the build so if doesn't break. But future updates should also use that URL.

yaakov-h:
Agreed - both are good individually, but better to not have the build break when a new version is published.

If only Dependant could detect this one 😄

yaakov-h:
17063 seems to be about 4.5 years old so I don't mind taking a dependency on that. As best I can tell, all versions of Windows older than that are out of mainstream support.

PR #1146, opened 2022-10-20
Cache NetHook native dependencies in actions

xPaw:


xPaw:
With no caching configure dependencies step takes 3 minutes, with caching it takes 22 seconds (although I kind of expected it to skip any building completely, but now it simply says that all outputs are up to date).

https://github.com/SteamRE/SteamKit/actions/runs/3296636475/jobs/5436524183

PR #1147, opened 2022-10-28
Implement decrypting remote client (LAN broadcast) messages

xPaw:
image


psychonic (approved):

Resources/NetHookAnalyzer2/NetHookAnalyzer2/MainForm.cs
```diff
@@ -105,7 +106,7 @@ string GetLatestNethookDumpDirectory()
 			return Path.Combine(nethookDirectory, latestDump);
 		}
 
-		NetHookDump Dump { get; set; }
+		internal static NetHookDump Dump { get; private set; }
```
yaakov-h, line 109: `internal` I can live with, but I really don't like `static` here.

Resources/NetHookAnalyzer2/NetHookAnalyzer2/NetHookDump.cs
```diff
@@ -23,26 +26,34 @@ public void LoadFromDirectory(string directory)
...
-			var fileInfo = new FileInfo(path);
 			var item = new NetHookItem();
 			if (!item.LoadFromFile(fileInfo))
 			{
 				return null;
 			}
 
 			items.Add(item);
+
+			if (item.EMsg == SteamKit2.EMsg.ServiceMethodResponse && item.InnerMessageName == "Credentials.GetAccountAuthSecret#1")
```
yaakov-h, line 47: So this relies upon us parsing the dump in chronological order?

Resources/NetHookAnalyzer2/NetHookAnalyzer2/MainForm.cs
xPaw: What's the alternative? I don't want to pass the dump all over the place.

Resources/NetHookAnalyzer2/NetHookAnalyzer2/NetHookDump.cs
xPaw: When it loads this secrets payload, earlier dumps will also decrypt because the decryption happens when you click on the item. But you do need to have this payload to be dumped at some point for this to work.

xPaw:
Also implemented specialization for broadcast payloads.

image

Resources/NetHookAnalyzer2/NetHookAnalyzer2/NetHookDump.cs
yaakov-h: Should we require the payload to be in the _same_ dump, or could we make `accountAuthSecrets` the static object instead?

PR #1148, opened 2022-10-28
Generate some webui protobufs for NHA2 to use

xPaw:
They exist only in JS code, but Steam client library uses them via service methods.


psychonic (approved):

yaakov-h:
not a fan of needing the `..\`s, but if it works it works 😄

xPaw:
Yeah it was easier to do ../ because I didn't want to copy paste the generation script.

PR #1150, opened 2022-10-29
Add button to directly open latest folder

xPaw:
This avoids the file picker dialog, so less steps.

image


Resources/NetHookAnalyzer2/NetHookAnalyzer2/MainForm.cs
```diff
@@ -179,24 +179,39 @@ void OnOpenToolStripMenuItemClick( object sender, EventArgs e )
...
+
+		}
+
+		void OnOpenLatestFolderToolStripMenuItemClick( object sender, EventArgs e )
+		{
+			var latestNethookDir = GetLatestNethookDumpDirectory();
+			if ( latestNethookDir != null )
+			{
+				OpenDirectory( latestNethookDir );
+			}
```
yaakov-h, line 192: If it is `null` then we should display an error rather than no-oping

psychonic (approved):

PR #1167, opened 2022-12-16
Fix machine_id on windows to be like the steam client

LelouBil:
First, the first available physical network interface is found. Using registry keys to differentiate the physical ones.

Then, instead of hashing only the MAC address, we hash 16 bytes where the first 8 are the MAC address.

That is because the steam client on windows hashes the first 16 bytes of a struct containing the mac address at the start and only zeroes afterwards.

### DIFF
--- a/SteamKit2/SteamKit2/Util/HardwareUtils.cs
+++ b/SteamKit2/SteamKit2/Util/HardwareUtils.cs
@@ -9,6 +9,7 @@ using System.Text;
 using SteamKit2.Util;
 using SteamKit2.Util.MacHelpers;
 using Microsoft.Win32;
+
 using static SteamKit2.Util.MacHelpers.LibC;
 using static SteamKit2.Util.MacHelpers.CoreFoundation;
 using static SteamKit2.Util.MacHelpers.DiskArbitration;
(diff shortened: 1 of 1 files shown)


LelouBil:
Okay, after investigating with a friend, there's a bit more to it. The bytes afterwards are not always zeroes. I'm on it right now, and I'll update the code.

LelouBil:
After investigating, the structure that is hashed is like this :
6 bytes of Mac address of the *first* physical network interface.
2 bytes of 0s
6 bytes of Mac address of the *second* physical network interface.
2 bytes of 0s


My computer is a desktop, and I had only one network adapter so the last 8 bytes were always zeroes, after reverse engineering steam a bit more on a friend's laptop we found this. He had Ethernet and wifi as different network adapters, so he had the second part filled.

I'll update the code tomorrow

xPaw:
@LelouBil can you add  `using` to the registry opens please? Then we can merge.

xPaw:
Merged in 74a32bb14e968b01f6d2e592092fecc26417776a

xPaw:
I am currently looking at the linux version, and it feels similar, however I have question:

What happens if you have more than 2 interfaces? Are we sure it blindly takes the first two it finds, or there's a missing sort by the mac address?

PR #1177, opened 2023-01-24
Create ContextMenuStrip on demand

xPaw:
Opening big payloads creates a bunch of objects, this PR changes to only create them on right click. Previous code already did this for elements in the context menu, but `ContextMenuStrip` itself consumes plenty of memory.

Before:
before

After:
after


JustArchi:
In particular, it all started from trying to open `ClienLicenseList` callback in NHA2. With big enough of libraries, especially going into 20k+ packages, `System.OutOfMemoryException` was to be expected.

JustArchi (approved):

PR #1183, opened 2023-02-20
Add machine name field to logon

xPaw:
One of (?) places it shows up is the profile games page:

20230220-PS8g79gN

### DIFF
--- a/SteamKit2/SteamKit2/Steam/Handlers/SteamUser/SteamUser.cs
+++ b/SteamKit2/SteamKit2/Steam/Handlers/SteamUser/SteamUser.cs
@@ -114,7 +114,7 @@ namespace SteamKit2
             /// Gets or sets the machine name.
             /// </summary>
             /// <value>The machine name.</value>
-            public string? MachineName { get; set; } = nameof( SteamKit2 );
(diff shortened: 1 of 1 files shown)


psychonic (approved):

SteamKit2/SteamKit2/Steam/Handlers/SteamUser/SteamUser.cs
```diff
@@ -110,6 +110,11 @@ public sealed class LogOnDetails
             /// </summary>
             /// <value>The client language.</value>
             public string ClientLanguage { get; set; }
+            /// <summary>
+            /// Gets or sets the machine name.
+            /// </summary>
+            /// <value>The machine name.</value>
+            public string? MachineName { get; set; } = nameof( SteamKit2 );
```
yaakov-h, line 117: I'm not sure I like settings this as a default - and definitely not as a `nameof(...)` since it's not the name of a particular C# construct we want.

I would have expected one of two defaults:
1. `null`, for backwards compatibility
2. `Environment.MachineName`, for default correctness, or perhaps a derivative thereof (e.g. `$"{Environment.MachineName} (SteamKit2)"`).

PR #1184, opened 2023-02-20
Update protobufs

xPaw:
Also updated gc.proto and EMsg.


psychonic (approved):

Resources/NetHookAnalyzer2/NetHookAnalyzer2/Generated/.gitattributes
```diff
@@ -0,0 +1 @@
+*.cs text eol=crlf -diff
```
yaakov-h, line 1: why not let autocrlf take care of it?

xPaw: It was mostly for -diff, and I copied it from steamkits generated folder.

yaakov-h: why not diff some nice text files?

FWIW GitHub seems to be ignoring it anyway

xPaw: I removed it.

yaakov-h (approved):

PR #1199, opened 2023-03-17
Add WebAPI `CallProtobufAsync` that accepts protobuf request message

xPaw:
Example:
```csharp
var service = Client.Configuration.GetAsyncWebAPIInterface( "IAuthenticationService" );
var response = await service.CallProtobufAsync<CAuthentication_BeginAuthSessionViaQR_Response, CAuthentication_BeginAuthSessionViaQR_Request>( HttpMethod.Post, "BeginAuthSessionViaQR", request );

response.Result == EResult.OK
response.Body // proto
```

WebAPIResponse could be an interesting breaking change because this would allow any other API calls to also access EResult.

### DIFF
--- a/SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs
+++ b/SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs
@@ -4,7 +4,6 @@ using System.Collections.Generic;
 using System.Dynamic;
 using System.IO;
 using System.Linq;
-using System.Net;
 using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Runtime.ExceptionServices;
@@ -24,7 +23,7 @@ namespace SteamKit2
         /// <summary>
         /// Represents a response to a WebAPI request.
         /// </summary>
-        public class WebAPIResponse<TResponse>
+        public sealed class WebAPIResponse<TResponse>
         {
             /// <summary>
             /// Gets the result of the request.
@@ -39,9 +38,9 @@ namespace SteamKit2
             /// <summary>
             /// Constructs <see cref="WebAPIResponse{TResponse}"/> object.
             /// </summary>
-            /// <param name="body">The result.</param>
-            /// <param name="result">The body.</param>
-            internal WebAPIResponse( TResponse body, EResult result )
+            /// <param name="body">The response body.</param>
+            /// <param name="result">The result.</param>
+            internal WebAPIResponse( EResult result, TResponse body )
             {
                 Result = result;
                 Body = body;
@@ -267,10 +266,25 @@ namespace SteamKit2
                 where TResponse : IExtensible, new()
                 where TRequest : IExtensible, new()
             {
-                using var input_payload = new MemoryStream();
-                Serializer.Serialize<TRequest>( input_payload, request );
+                if ( method == null )
+                {
+                    throw new ArgumentNullException( nameof( method ) );
+                }
 
-                var base64 = Convert.ToBase64String( input_payload.ToArray() );
+                if ( func == null )
+                {
+                    throw new ArgumentNullException( nameof( func ) );
+                }
+
+                if ( request == null )
+                {
+                    throw new ArgumentNullException( nameof( request ) );
+                }
+
+                using var inputPayload = new MemoryStream();
+                Serializer.Serialize<TRequest>( inputPayload, request );
+
+                var base64 = Convert.ToBase64String( inputPayload.ToArray() );
                 var args = new Dictionary<string, object?>
                 {
                     { "input_protobuf_encoded", base64 },
@@ -287,9 +301,9 @@ namespace SteamKit2
                 }
 
                 var body = Serializer.Deserialize<TResponse>( stream );
-                var api_response = new WebAPIResponse<TResponse>( body, eresult );
+                var apiResponse = new WebAPIResponse<TResponse>( eresult, body );
 
-                return api_response;
+                return apiResponse;
             }
 
             /// <summary>
--- a/SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs
+++ b/SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs
@@ -22,6 +22,33 @@ namespace SteamKit2
     public sealed class WebAPI
     {
         /// <summary>
+        /// Represents a response to a WebAPI request.
+        /// </summary>
+        public class WebAPIResponse<TResponse>
+        {
+            /// <summary>
+            /// Gets the result of the request.
+            /// </summary>
+            public EResult Result { get; }
+
+            /// <summary>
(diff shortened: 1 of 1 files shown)


SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs
```diff
@@ -21,6 +21,33 @@ namespace SteamKit2
     /// </summary>
     public sealed class WebAPI
     {
+        /// <summary>
+        /// Represents a response to a WebAPI request.
+        /// </summary>
+        public class WebAPIResponse<TResponse>
```
yaakov-h, line 27:
```suggestion
        public sealed class WebAPIResponse<TResponse>
```

SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs
```diff
@@ -21,6 +21,33 @@ namespace SteamKit2
...
+            /// Gets the body of the request.
+            /// </summary>
+            public TResponse Body { get; }
+
+            /// <summary>
+            /// Constructs <see cref="WebAPIResponse{TResponse}"/> object.
+            /// </summary>
+            /// <param name="body">The result.</param>
+            /// <param name="result">The body.</param>
+            internal WebAPIResponse( TResponse body, EResult result )
```
yaakov-h, line 44: I would reverse these two parameters, also noting that everywhere else they are already in opposite order with the result coming before the body, even in the constructor body.

Also, do we want to have any null guarantees or checks here?

SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs
```diff
@@ -222,6 +249,49 @@ public async Task<T> CallProtobufAsync<T>( HttpMethod method, string func, int v
...
+            /// <returns>A <see cref="Task{T}"/> that contains object representing the results of the Web API call.</returns>
+            /// <exception cref="ArgumentNullException">The function name or request method provided were <c>null</c>.</exception>
+            /// <exception cref="HttpRequestException">An network error occurred when performing the request.</exception>
+            /// <exception cref="WebAPIRequestException">A network error occurred when performing the request.</exception>
+            /// <exception cref="ProtoException">An error occured when parsing the response from the WebAPI.</exception>
+            public async Task<WebAPIResponse<TResponse>> CallProtobufAsync<TResponse, TRequest>( HttpMethod method, string func, TRequest request, int version = 1 )
+                where TResponse : IExtensible, new()
+                where TRequest : IExtensible, new()
+            {
+                using var input_payload = new MemoryStream();
```
yaakov-h, line 270: `inputPayload` - variables are camel case, not train/snake case.

SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs
```diff
@@ -222,6 +249,49 @@ public async Task<T> CallProtobufAsync<T>( HttpMethod method, string func, int v
...
+                Serializer.Serialize<TRequest>( input_payload, request );
+
+                var base64 = Convert.ToBase64String( input_payload.ToArray() );
```
yaakov-h, line 273: Perhaps use `ToBase64String (byte[] inArray, int offset, int length);` with the MemoryStream's underlying buffer, to avoid an additional array copy.

SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs
```diff
@@ -222,6 +249,49 @@ public async Task<T> CallProtobufAsync<T>( HttpMethod method, string func, int v
...
+                {
+                    { "input_protobuf_encoded", base64 },
+                };
+
+                var response = await CallAsyncInternal( method, func, version, args, "protobuf_raw" );
+                var eresult = EResult.Invalid;
+
+                using var stream = await response.Content.ReadAsStreamAsync().ConfigureAwait( false );
+
+                if ( response.Headers.TryGetValues( "x-eresult", out var result ) && int.TryParse( result.First(), out var resultInt ) )
```
yaakov-h, line 284: Should the `EResult` perhaps be nullable, i.e. null if it is not in the HttpResponseMessage, rather than defaulting to Invalid? Or does this cause significantly more typically-unneccesary work for consumers/

SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs
```diff
@@ -222,6 +249,49 @@ public async Task<T> CallProtobufAsync<T>( HttpMethod method, string func, int v
...
+                {
+                    eresult = ( EResult )resultInt;
+                }
+
+                var body = Serializer.Deserialize<TResponse>( stream );
+                var api_response = new WebAPIResponse<TResponse>( body, eresult );
```
yaakov-h, line 290: casing here too

yaakov-h (changes requested):
What do you mean by an interesting **_breaking_** change?

xPaw:
> What do you mean by an interesting breaking change?

Not related to this PR, I was thinking the existing overloads could return this object which exposes the EResult for you.

SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs
xPaw: I don't think null check is necessary here, deserialize will fail.

SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs
xPaw: Yeah I think nullable isn't particularly nice here. EResult.Invalid is 0.

SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs
```diff
@@ -21,6 +20,33 @@ namespace SteamKit2
     /// </summary>
     public sealed class WebAPI
     {
+        /// <summary>
+        /// Represents a response to a WebAPI request.
+        /// </summary>
+        public sealed class WebAPIResponse<TResponse>
+        {
+            /// <summary>
+            /// Gets the result of the request.
```
yaakov-h, line 29:
```suggestion
            /// Gets the result of the response.
```

SteamKit2/SteamKit2/Steam/WebAPI/WebAPI.cs
```diff
@@ -21,6 +20,33 @@ namespace SteamKit2
...
+            /// </summary>
+            public EResult Result { get; }
+
+            /// <summary>
+            /// Gets the body of the request.
```
yaakov-h, line 34:
```suggestion
            /// Gets the body of the response.
```

yaakov-h (changes requested):
LGTM minus the suggested changes for doc comments

PR #1206, opened 2023-03-21
Push packages regularly to GitHub Packages

yaakov-h:
This allows for easier testing an experimentation without relying on a contributor to push tags and create (pre)releases.


xPaw:
Why did you enable submodules checkout? There's only a submodule in Resources/Protobufs, and that shouldn't affect publishing.

yaakov-h:
That may have been to shut up the SourceLink warnings about incomplete submodules.

PR #1210, opened 2023-03-28
Dispose objects where possible

xPaw:
One big object is `MsgBase` which holds a memory stream and binary readers, but making that disposable requires a bit of a refactor.

### DIFF
--- a/SteamKit2/SteamKit2/Steam/SteamClient/CallbackMgr/CallbackMgr.cs
+++ b/SteamKit2/SteamKit2/Steam/SteamClient/CallbackMgr/CallbackMgr.cs
@@ -7,7 +7,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Reflection;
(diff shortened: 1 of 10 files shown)


SteamKit2/SteamKit2/Steam/SteamClient/CallbackMgr/CallbackMgr.cs
```diff
@@ -112,7 +111,9 @@ public IDisposable Subscribe<TCallback>( JobID jobID, Action<TCallback> callback
                 throw new ArgumentNullException( nameof(callbackFunc) );
             }
 
+#pragma warning disable CA2000 // Not implicitly disposed
```
yaakov-h, line 114: I suspect this is warning because the `Subscription` constructor could throw, and this remains undisposed.

yaakov-h (approved):

PR #1212, opened 2023-04-03
Allow `DuplicateRequest` in `SendSteamGuardCodeAsync`

xPaw:
`DuplicateRequest` happens when accepting the prompt in the mobile app, and then trying to send guard code here, do not throw on it here because authentication will succeed on the next poll.

This makes it easier for users because even if 2FA code is asked, Steam will send a confirmation prompt to the app regardless.

cc @JustArchi

### DIFF
--- a/Samples/1.Logon/Program.cs
+++ b/Samples/1.Logon/Program.cs
@@ -15,6 +15,12 @@ using SteamKit2;
 // callback type to a function, whilst CallbackMgr routes the callback objects to the functions that
 // you have specified. a Callback<T> is bound to a specific callback manager.
 //
+//
+// WARNING!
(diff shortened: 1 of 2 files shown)


yaakov-h:
Seems reasonable to me.

What happens if authentication doesn't succeed on the next poll, but does something else (e.g. fails)? Does that just get handed as normal?

xPaw:
Polling will throw for any eresult that isn't OK, yeah.

JustArchi (changes requested):
We need some way to indicate that the next poll should finish the flow with either success or error, we might have a problem if Steam tells us duplicate for any stupid reason and then will continue giving us "wait" in polls.

JustArchi (approved):
obraz

PR #1232, opened 2023-06-09
Fix CWebSocketConnection::BBuildAndAsyncSendFrame pattern

jessy-lua:
In the Steam beta client 2 functions with the same pattern. Extended by 2 additional bytes. Working on current and beta updates.


voided:
Thanks!

PR #1235, opened 2023-07-06
Add 64-bit support to NetHook2

sapphyrus:
This makes it possible to use NetHook2 in 64-bit-only situations, for example a CS2 dedicated server.

Some work is missing, namely the 64-bit Visual Studio configurations and AppVeyor configuration.
(Un)fortunately I don't use VS so I will leave creating these up to somebody else (I tested using cmake).

If you need any other changes lmk.


Resources/NetHook2/NetHook2/crypto.cpp
```diff
@@ -13,41 +14,53 @@
 SymmetricEncryptChosenIVFn Encrypt_Orig = nullptr;
 PchMsgNameFromEMsgFn PchMsgNameFromEMsg = nullptr;
 
+#ifndef __x86_64__
```
psychonic, line 17: Would prefer to leave this in, just checking for `sizeof(MsgInfo_t) == 24` on `__x86_64`. The other asserts should pass without issue on x64.

Resources/NetHook2/NetHook2/detours64.lib
psychonic, line 1: Where was this build sourced from?

psychonic (changes requested):
Thanks. Looks good overall. I added some comments.

Resources/NetHook2/NetHook2/detours64.lib
sapphyrus: It's an old build I had laying around since I was too lazy to compile it myself, I don't think there are any official builds for it either.

psychonic (approved):

yaakov-h:
I would have merged #1242 first so that we don't have an extra .lib that we immediately delete (I hope...?)

psychonic:
It's minimal risk due to not being included in the project file. One would really have to opt-in to use it, and I've rebased and removed it from that PR now.

PR #1246, opened 2023-08-19
NetHook: Add 64-bit project config and CI build

psychonic:
This adds an x64 solution/project configuration for NetHook, as well as adding these builds to the existing CI flow.


yaakov-h:
I think the artifact uploads from each configuration might be stomping over each-other?

psychonic:
> I think the artifact uploads from each configuration might be stomping over each-other?

🤦Indeed they were. That's been corrected now.

xPaw (approved):

PR #1270, opened 2023-10-18
Deprecate WebAPIUserNonce in logon, and machine auth callback

xPaw:
Ref: https://github.com/SteamRE/SteamKit/discussions/1258
Ref: https://github.com/JustArchiNET/ArchiSteamFarm/issues/3029


psychonic:
Looks like Sample 5 still uses this.

xPaw:
So does sample 1 kind of. But sample 1a covers the new auth, might need to create a new sample for guard data handling at some point.

PR #1274, opened 2023-10-20
Set rollForward to latestMajor in global.json

xPaw:
Readme says:

> To compile SteamKit2, the following is required: .NET 6.0 SDK or higher.

which is not true.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
     "version": "6.0.100",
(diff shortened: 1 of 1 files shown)


yaakov-h:
I'd rather change the readme then, `latestMajor` means we could have all sorts of inconsistencies or issues moving forwards, such as when dotnet CLI args change

xPaw:
I don't like that it forces me to have this version locally too.

PR #1277, opened 2023-10-27
Switch to MSTest, run tests in parallel

xPaw:
Reduces test time from ~30s to <1s.

xunit does not allow parallelizing within a test class. mstest has a ` [DoNotParallelize]` which I set on the debug log facts.

I changed async job tests to call `CancelTimedoutJobs` directly instead of relying on the scheduled timer to fire, which can have very inconsistent timings on CI runs, even when timer delay is reduced to like 50ms.

~~However the async job tests still have `Task.Delay` and that still introduces randomness (especially on Windows builds) causing random failures.~~


yaakov-h:
I'm personally not a fan of MSTest, and even .NET itself is tested using Xunit rather than MSTest.

Personally I'd prefer NUnit, but lately they've been in the habit of doing stupid things with .NET runtime detection, which means every November it starts crashing unless they ship an upgrade _and_ you upgrade to it, so I can't exactly give it a whole-hearted endorsement.

In terms of time-based tests, I would like to keep them time-based, but switch to `TimeProvider` (coming in .NET 8 with down-level support via NuGet) and use [Microsoft.Extensions.TimeProvider.Testing](https://www.nuget.org/packages/Microsoft.Extensions.TimeProvider.Testing) to instead simulate the flow of time.

xPaw:
The tests stayed time based, I am just calling cancel manually instead of waiting for the scheduled timer to fire, which I don't think the extension you're showing will solve.

PR #1279, opened 2023-10-27
Enable package lock and dotnet setup cache

xPaw:
This makes builds repeatable: https://devblogs.microsoft.com/nuget/enable-repeatable-package-restores-using-a-lock-file/

EDIT: Doesn't work cross platform: https://github.com/NuGet/Home/issues/9195


yaakov-h:
> EDIT: Doesn't work cross platform:

RIP.

Personally I've found NuGet's lock files to be fairly immature and laden with subtle problems and vastly increased complexity and maintenance, so I'd rather avoid them until... if I can be honest, until they actually work.

PR #1280, opened 2023-10-29
Make xunit tests run faster

xPaw:
Is this better? I don't think FakeTimeProvider would help us test async job timeouts because it would still have to heavily rely on the Threading.Timer to fire consistently, calling `CancelTimedoutJobs` directly avoids this problem.


yaakov-h (approved):
FakeTimeProvider should help if we pass it into SK2 and call `TImeProvider.CreateTimer` on it, however that would incur an additional dependency, and at the moment that's still prerelease anyway.

Once .NET 8 lands - in a couple of weeks - we can look at updating to .NET 8 and adopting TimeProvider, but for now this looks fine to me.

The only thing I don't like is `internal void CancelTimedoutJobs` but I would hope that can be made private again once we're on TimeProvider.

xPaw:
> The only thing I don't like is internal void CancelTimedoutJobs but I would hope that can be made private again once we're on TimeProvider.

If that dependency works out, sure.

You want to drop net framework support with net8?

yaakov-h:
We could #ifdef it, though at this point I wouldn't mind dropping Framework support.
Then I could also squeeze some Spans and other optimizations in more easily.

PR #1281, opened 2023-10-30
Use Stopwatch instead of wall clock for async job timeouts

xPaw:
Seems like a better thing to do, should also hopefully improve randomness in tests.

### DIFF
--- a/SteamKit2/SteamKit2/Types/JobID.cs
+++ b/SteamKit2/SteamKit2/Types/JobID.cs
@@ -6,10 +6,9 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
-using System.Linq;
 using System.Runtime.CompilerServices;
-using System.Text;
 using System.Threading.Tasks;
+using SteamKit2.Util;
 
 namespace SteamKit2
 {
@@ -94,7 +93,7 @@ namespace SteamKit2
     /// </summary>
     public abstract class AsyncJob
     {
-        DateTime jobStart;
+        ValueStopwatch jobStart;
 
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


SteamKit2/SteamKit2/Util/ValueStopwatch.cs
```diff
@@ -0,0 +1,32 @@
...
+        public TimeSpan GetElapsedTime()
+        {
+            if ( _startTimestamp == 0 )
+            {
+                throw new InvalidOperationException( "An uninitialized, or 'default', ValueStopwatch cannot be used to get elapsed time." );
+            }
+
+            long end = Stopwatch.GetTimestamp();
+            long timestampDelta = end - _startTimestamp;
+            long ticks = ( long )( s_timestampToTicks * timestampDelta );
```
yaakov-h, line 28: isn't the timestamp supposed to already be in Ticks?

xPaw: This is straight from .NET runtime: https://source.dot.net/#Microsoft.Extensions.Http/ValueStopwatch.cs,510774ca21949635

SteamKit2/SteamKit2/Types/JobID.cs
```diff
@@ -6,10 +6,12 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
+using System.Diagnostics;
```
yaakov-h, line 9: [where] is this needed?

xPaw: its not. gonna need to do a usings cleanup pass, because a lot of files have unused ones.

PR #1284, opened 2023-10-31
Target .NET 8

xPaw:
Target for next major SteamKit version (3.0.0). This drops support for .NET framework.


JustArchi:
Did you just kill .NET Framework my bro?!

yaakov-h:
It’s been four years since https://devblogs.microsoft.com/dotnet/net-core-is-the-future-of-net/

Unless we have some exotic enterprise customer, .NET 8 LTS seems like a great time to make the jump.

Many other frameworks like asp.net core and powershell dropped Framework support years ago…

yaakov-h (commented):
Tentative approval for a v3.0.0 release after .NET 8 lands in 2 weeks time. We should probably bump the version in this PR too.

We should deal with the new Analyzer warnings as well. From discussions on dotnet/csharplang, I think the lowercase identifiers warning is one we should fix rather than suppress.

xPaw:
>  I think the lowercase identifiers warning is one we should fix rather than suppress.

One of the warnings comes from generate proto.

yaakov-h:
we control the proto generation, so we can tweak it as neccesary.

and yeah, I'm not sure what's with SizeConst either, I'd have to read up on it

xPaw:
It's defined as `DISK_EXTENT Extents[ANYSIZE_ARRAY];`, ANYSIZE_ARRAY is 1.

https://devblogs.microsoft.com/oldnewthing/20040826-00/?p=38043

xPaw:
> we control the proto generation, so we can tweak it as neccesary.

So protobuf-net already has escaping code for all known language identifiers, If I modify that escape function to capitalize if the identifier is all lowercase, it will capitalize all fields. If I remove `names=original` option, that will also capitalize everything.

xPaw:
`AdditionalParameters` in the next struct defined it as 1 too, and its `BYTE                AdditionalParameters[1];`

voided (commented):
LGTM. Agree with @yaakov-h on waiting for .NET 8 GA.

I posed this question on the SteamDB discord, but do we want to take this major version bump opportunity to make more sweeping changes like any wishlisted tech debt cleanup, project structure reorg, etc.

xPaw:
> take this major version bump opportunity to make more sweeping changes?

You can if there's anything that needs doing, make issues for these? I have PR open to remove obsoletes. Yaakov was looking at switching to google protobufs for AOT stuff.

yaakov-h:
Yep, as far as I can tell Google Protobuf is already trimmable and AOT-ready, and it has better support within the .NET ecosystem than protobuf-net does.

(protobuf-net has been blocking us with https://github.com/protobuf-net/protobuf-net/pull/817 on not being trimmable for over two years, and I believe we could need to generate our own serialisation code to support AOT)

I'm not making any commitment to switching, but it's something I'm vaguely interested in and would also make sense for a v3.0 bump.

Potentially something else we could do is replace the `KeyValue` class with the VRF lib? 🙂

xPaw:
> Potentially something else we could do is replace the KeyValue class with the VRF lib? 🙂

That indeed would be good, but https://github.com/ValveResourceFormat/ValveKeyValue/issues/30 needs solving first, ideally somehow making KeyValue a non-breaking change for consumers.

xPaw:
NET8 is GA, rip framework.

PR #1298, opened 2023-11-16
Switch to `System.IO.Hashing.Crc32`

xPaw:
Tested TCP connection, that still works.

This library has a pre-calculated table, and uses intrinstics when available (although only on arm if I looked at the right source code).


yaakov-h (approved):
I'm not the biggest fan of adding more package dependencies but this seems reasonable.

xPaw:
I'm not sure why this package is not shipped by default in the SDK.

PR #1317, opened 2023-12-27
update Utils.cs

ilyfairy:


yaakov-h (commented):
This looks generally OK to me but I'm a little bit concerned about the long/ulong conversions.

ilyfairy:
The timestamp is usually not larger than long.MaxValue, if it is, then there will be a problem in the previous logic too

yaakov-h:
ah, I didn't realise that the old code had an implicit cast to `double` in there.

thanks.

yaakov-h (approved):

xPaw:
This PR introduced a bug, ToUnixTimeSeconds is applying local timezone offset, so unix timestamps no longer round trip correctly.

PR #1318, opened 2023-12-28
Reduce memory usage during decryption

ilyfairy:


SteamKit2/SteamKit2/Util/CryptoHelper.cs
```diff
@@ -202,14 +202,13 @@ public static byte[] SymmetricEncryptWithHMACIV( byte[] input, byte[] key, byte[
             var random = GenerateRandomBlock( 3 );
             Array.Copy( random, 0, iv, iv.Length - random.Length, random.Length );
 
-            using ( var hmac = new HMACSHA1( hmacSecret ) )
-            using ( var ms = new MemoryStream() )
+            using ( var ms = new MemoryStream( random.Length + input.Length ) )
```
xPaw, line 205: I feel like ArrayPool should be used in these methods.

ilyfairy: I found that if it is a small array, ArrayPool will consume more performance, but stackalloc is worth trying

xPaw: Are you sure that will hold true for long running processes?

ilyfairy: image
I think it's fine, the memory will always be collect in time

yaakov-h: Can you benchmark the old against the new please, including memory stats? I’d expect the change in allocations to reflect a change in CPU time, but it would be good to have actual figures.

ilyfairy: image

yaakov-h: I ran a couple benchmarks on my laptop too, just to confirm.

This PR drops execution time about 10-30%. If we use `ArrayPool<byte>.Shared` instead of `new MemoryStream` then we can get a further 15-25% savings.

However, considering that on my laptop these calls are all measured in microseconds, it's hardly worth the effort to argue over it.

image

I'm happy to merge this if you are @xPaw.

ilyfairy: However, its memory usage is much reduced compared to the original, especially in multi-threading. Multiple threads download chunkdata and decrypt it at the same time, which can reduce the memory by more than 200MB (500MB->300MB).

xPaw: Okay I looked at this code again, a temporary array is not even required.

Make an instance of HMACSHA1, and then use TransformBlock / TransformFinalBlock if I remember correctly.

But this code also allocates other random arrays like `iv` and `GenerateRandomBlock`, this could also be refactored (and I would use ArrayPool).

If the entire encryption flow was refactored, I'm almost positive it's possible to get it down to zero allocations (besides arraypool where it really is needed).

xPaw:
@ilyfairy are you interested in doing the refactor as I wrote in my last comment?

ilyfairy:
> @ilyfairy are you interested in doing the refactor as I wrote in my last comment?

It is public, refactoring will change the signature, which may cause code incompatibility. It is up to you to refactor.

xPaw:
Breaking changes are fine.

xPaw:
Closing in favor of #1381 and #1388  (and the other PRs I opened)

PR #1113, opened 2022-07-17
Add functionality for serializing depot manifests

NicknineTheEagle:
This adds functionality for serializing depot manifests into Steam-compatible protobuf files. The output matches .manifest files found in Steam's depotcache folder unless there are non-ASCII filenames present. It seems that Steam handles filename sorting slightly differently, not quite sure what's going on there.

I'm intending to use this as a stepping stone towards migrating DepotDownloader to using Steam protobuf manifests.

PR #1337, opened 2024-02-22
Optimize DepotManifest Serialization

skylayer:
This PR builds on PR #1113, introducing a `HashSet` with `ChunkIdComparer` to manage unique chunks in `DepotManifest` serialization, significantly reducing processing time.

[My comment](https://github.com/SteamRE/SteamKit/pull/1113#issuecomment-1958688449)

### Changes:
- Replaced `List<byte[]>` with `HashSet<byte[]>` for unique chunks.
- Added `ChunkIdComparer` for efficient byte array comparison.

### Impact:
- **Before**: Serialization took up to 10 minutes.
- **After**: Reduced to ~0.61 seconds.

This optimization streamlines serialization, especially for large manifests, enhancing performance without data integrity loss.

### DIFF
--- a/SteamKit2/SteamKit2/Types/DepotManifest.cs
+++ b/SteamKit2/SteamKit2/Types/DepotManifest.cs
@@ -220,25 +220,6 @@ namespace SteamKit2
         }
 
         /// <summary>
-        /// Serializes depot manifest and saves the output to a file.
-        /// </summary>
-        /// <param name="filename">Output file name.</param>
-        /// <returns><c>true</c> if serialization was successful; otherwise, <c>false</c>.</returns>
-        public bool SaveToFile( string filename )
-        {
-            try
-            {
-                using var fs = File.Open( filename, FileMode.Create );
-                Serialize( fs ); // Directly pass the FileStream to the Serialize method
-                return true; // If serialization completes without throwing an exception, return true
-            }
-            catch ( Exception )
-            {
-                return false; // Return false if an error occurs
-            }
-        }
-
-        /// <summary>
         /// Loads binary manifest from a file and deserializes it.
         /// </summary>
         /// <param name="filename">Input file name.</param>
--- a/SteamKit2/SteamKit2/Types/DepotManifest.cs
+++ b/SteamKit2/SteamKit2/Types/DepotManifest.cs
@@ -226,16 +226,16 @@ namespace SteamKit2
         /// <returns><c>true</c> if serialization was successful; otherwise, <c>false</c>.</returns>
         public bool SaveToFile( string filename )
         {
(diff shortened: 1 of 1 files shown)


xPaw:
Why does it even need to check for duplicates?

skylayer:
BTW, I'm thinking of making the `Serialize()` function in `DepotManifest` public. Right now, we've only got `SaveToFile` accessible, which kinda boxes us in. If `Serialize()` were public, we could directly handle more scenarios, like different formats or custom storage solutions.

NicknineTheEagle:
> Why does it even need to check for duplicates?

Because Valve's manifest format stores the number of unique chunks.

NicknineTheEagle:
> BTW, I'm thinking of making the `Serialize()` function in `DepotManifest` public. Right now, we've only got `SaveToFile` accessible, which kinda boxes us in. If `Serialize()` were public, we could directly handle more scenarios, like different formats or custom storage solutions.

I checked the original PR and I don't know why I didn't make Serialize public, to be honest.

SteamKit2/SteamKit2/Types/DepotManifest.cs
```diff
@@ -226,16 +226,16 @@ public bool DecryptFilenames(byte[] encryptionKey)
...
-            if ( data != null )
+            try
             {
-                bw.Write( data );
-                return true;
+                using var fs = File.Open( filename, FileMode.Create );
+                Serialize( fs ); // Directly pass the FileStream to the Serialize method
+                return true; // If serialization completes without throwing an exception, return true
+            }
+            catch ( Exception )
```
xPaw, line 235: Are we expecting any exceptions, why the catch?

xPaw (approved):

SteamKit2/SteamKit2/Types/DepotManifest.cs
xPaw: I think these should just throw.

yaakov-h: `catch (Exception)` is also bad practice: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1031

Either we should catch certain known exceptions, or let them leak out in actually-exceptional scenarios.

yaakov-h: I also can't find a way in which the old method would ever return `false` because `Serialize()` seems to always return a non-null value.

yaakov-h (commented):
I'm not sold on the exception handling but the core of the change looks good.

yaakov-h:
btw congrats on getting PR number 1337 😉

Issue #1057, opened 2021-11-15
Provide trimming annotations for self-contained apps


yaakov-h:
.NET 5 previewed the ability to trim unused code from applications and from the runtime, and .NET 6 now makes that generally available.

Relevant docs are here: https://docs.microsoft.com/en-us/dotnet/core/deploying/trimming/trim-self-contained

Considering how many protobuf messages we ship that most apps won't use, this is quite valuable. DepotDownloader can trim about 4.5-5 MB away from SteamKit2.dll and a fully self-contained trimmed build of DD is about 15-20MB IIRC. 

Related issues:
- https://github.com/dotnet/linker/issues/2365 (blocking)
- https://github.com/dotnet/linker/issues/2366 (non-blocking)
- https://github.com/protobuf-net/protobuf-net/pull/817 (probably non-blocking)

Some initial prototyping has already been done in this pull request:
- https://github.com/SteamRE/SteamKit/pull/1002

yaakov-h:
We may have to switch to the Google.Protobuf library which isn't reflection-based in order to (a) get this working properly, and (b) be able to run in an AOT environment such as what is being planned at the moment for .NET 7.

🤔

xPaw:
Unified messages are no longer a blocker for this, since they are fully generated without reflection.

xPaw:
With updated protobuf-net, there are still issues with properties being trimmed away:

`LoggedOnCallback` has this: `this.ParentalSettings = Serializer.Deserialize<ParentalSettings>( ms );`

`ParentalSettings` has `public ParentalPlaytimeRestrictions playtime_restrictions { get; set; }`, but after trimming only an empty class `ParentalPlaytimeRestrictions` is produced so it fails to deserialize it.

yaakov-h:
I did have fun problems previously with trimming and nested classes.

That said, what does "fails to deserialise" look like?

xPaw:
It's trying to call an empty constructor which does not exist.

mizantrop777:
trimming can be disabled via

`<TrimmerRootDescriptor Include="ILLink.xml" />`

ILLink.xml:
````
<?xml version="1.0" encoding="utf-8" ?>
<linker>
  <!-- Пример сохранения всех типов в сборке -->
  <assembly fullname="Updater" preserve="all" />
  <assembly fullname="SteamKit2" preserve="all" />
  <assembly fullname="protobuf-net" preserve="all" />
[...]

PR #1342, opened 2024-03-18
Move webui protos to SteamKit2.WebUI.Internal

xPaw:
Also added a few new ones. They are useful to have, but the way webui dumping works, I'm not confident in it long term.

~~I did not generate webui/common, because that ends up being a 1mb file, and we can reuse Steam common protos.~~

Dumper code lives here: https://github.com/SteamDatabase/SteamTracking/blob/master/dump_javascript_protobufs.mjs


JustArchi:
Could you consider adding https://github.com/SteamDatabase/Protobufs/blob/master/webui/service_familygroups.proto too? Would help me over in ASF, thank you in advance :pray:

xPaw:
Family groups exist in Steam protos, so generated here: 530f8e4674f5bcfcd87e877abdcab93f2e88606f

ezhevita:
Please consider adding [service_loyaltyrewards.proto](https://github.com/SteamDatabase/Protobufs/blob/master/webui/service_loyaltyrewards.proto) (used by ASF) and [service_accountcart.proto](https://github.com/SteamDatabase/Protobufs/blob/master/webui/service_accountcart.proto) with [service_checkout.proto](https://github.com/SteamDatabase/Protobufs/blob/master/webui/service_checkout.proto) for generation

xPaw:
The annoying thing about cart and checkout protos is that they actually require webui/common, but the generator does not really support only generating the messages that are used. I don't want to bring 1mb of common protos in steamkit.

If the generator somehow could just copy stuff it needs from common, that would solve this issue.

xPaw:
After some discussion, as these protos are useful to have, and trying to solve common protos is a bit of a pain, i'm gonna include them.

This should be solved by fixing trimming instead: https://github.com/SteamRE/SteamKit/issues/1057

PR #1347, opened 2024-04-04
Remove Artifact and Underlords protobufs

xPaw:
These games are effectively dead.

https://store.steampowered.com/news/app/583950/view/3047218819080842820

If people still need these, they can generate these protos themselves.


psychonic (approved):

voided (approved):
LGTM.

I don't feel too strongly about this, but perhaps we should generate those protos in NHA2 so at least the dissectors can keep working - but I have no idea how many people are looking at historical GC messages for these games.. Probably nobody?

PR #1348, opened 2024-04-07
Enable deterministic builds

xPaw:
@JustArchi Archi do we need to add any other magic to improve the nuget?

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -17,11 +17,17 @@
         <InformationalVersion Condition=" '$(GitHubActionsRef)' == ''">$(SteamKitVersion) - Development</InformationalVersion>
         <InformationalVersion Condition=" '$(GitHubActionsRef)' != ''">$(SteamKitVersion) - CI (GitHub, ref: $(GitHubActionsRef))</InformationalVersion>
     </PropertyGroup>
+
     <PropertyGroup>
         <GenerateAssemblyTitleAttribute>false</GenerateAssemblyTitleAttribute>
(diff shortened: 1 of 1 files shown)


Directory.Build.props
```diff
@@ -17,11 +17,17 @@
...
         <GenerateAssemblyConfigurationAttribute>false</GenerateAssemblyConfigurationAttribute>
         <GenerateAssemblyCompanyAttribute>false</GenerateAssemblyCompanyAttribute>
         <GenerateAssemblyProductAttribute>false</GenerateAssemblyProductAttribute>
     </PropertyGroup>
-</Project>
\ No newline at end of file
+
+    <PropertyGroup>
+        <Deterministic>true</Deterministic>
+        <ContinuousIntegrationBuild Condition="'$(CI)' == 'true'">true</ContinuousIntegrationBuild>
```
yaakov-h, line 31: Isn’t this set automatically based on env cars these days?

xPaw: I don't think it is, I checked other packages that had Deterministic set, but nuget analyzer still complained about this not being set.

JustArchi:
@xPaw I only set deterministic to true in csproj, and add `-p:ContinuousIntegrationBuild=true` when building within CI, so pretty much what you did, but I prefer to not rely on `CI` being set in the env of the runners.

xPaw:
CI env is documented here: https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables

It's also commonly available on other systems.

PR #1352, opened 2024-04-15
Added `ChatMode` property to the `LogOnDetails` class

Rei-sen:
This lets the user enable new Steam group chat

### DIFF
--- a/SteamKit2/SteamKit2/Steam/Handlers/SteamUser/SteamUser.cs
+++ b/SteamKit2/SteamKit2/Steam/Handlers/SteamUser/SteamUser.cs
@@ -111,6 +111,12 @@ namespace SteamKit2
             /// </summary>
             /// <value>The machine name.</value>
             public string? MachineName { get; set; } = $"{Environment.MachineName} (SteamKit2)";
+            /// <summary>
+            /// Gets or sets the chat mode.
+            /// Setting this field to 2 enables support for new Steam group chat.
+            /// </summary>
+            /// <value>The chat mode.</value>
+            public uint ChatMode { get; set; } = 0;
 
             /// <summary>
             /// Initializes a new instance of the <see cref="LogOnDetails"/> class.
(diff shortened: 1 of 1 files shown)


SteamKit2/SteamKit2/Steam/Handlers/SteamUser/SteamUser.cs
```diff
@@ -111,6 +111,12 @@ public sealed class LogOnDetails
             /// </summary>
             /// <value>The machine name.</value>
             public string? MachineName { get; set; } = $"{Environment.MachineName} (SteamKit2)";
+            /// <summary>
+            /// Gets or sets the chat mode.
+            /// Setting this field to 2 enables support for new Steam group chat.
```
yaakov-h, line 116: This sounds more like an enum than a uint.

JustArchi (approved):

JustArchi:
Looks much cleaner than announcing chat mode after login, which is what I'm doing in ASF right now.

Any objections to merge this?

SteamKit2/SteamKit2/Steam/Handlers/SteamUser/SteamUser.cs
xPaw: You can remove this string now that it's actually an enum.

SteamKit2/SteamKit2/Steam/Handlers/SteamUser/SteamUser.cs
```diff
@@ -253,6 +275,8 @@ public void LogOn( LogOnDetails details )
             logon.Body.machine_name = details.MachineName;
             logon.Body.machine_id = HardwareUtils.GetMachineID( Client.Configuration.MachineInfoProvider );
 
+            logon.Body.chat_mode = ( uint )details.ChatMode;
```
xPaw, line 278: Perhaps wrap this into an if chatmode is not default?

PR #1022, opened 2021-09-25
Refactor CDN client

xPaw:
- Moved it to `CDN` namespace, with classes being split.
- Removed `AuthenticateDepot` and CDN auth tokens
- Added `GetServersForSteamPipe` which goes over CM instead of WebAPI
- Added support for manifest request codes:
Is `SteamContent` a good name for this? There's other methods I'd like to implement from it later:
https://github.com/SteamRE/SteamKit/blob/b5d8502afa01385c7d4265f0a49dc3a56c03df23/SteamKit2/SteamKit2/Base/Generated/SteamMsgContentSystem.cs#L465-L466

I made the function async, because AsyncJob can't deal with unified messages nicely as they get routed to `SteamUnifiedMessages`. Perhaps (in future) it could be expanded so it would support unified messages natively for proper callback handling?

PR #1372, opened 2024-06-30
Add support for CDN auth

oureveryday:
Some Steam CDNs in China requires a CDN auth to download.

### DIFF
--- a/SteamKit2/SteamKit2/Steam/CDN/Client.cs
+++ b/SteamKit2/SteamKit2/Steam/CDN/Client.cs
@@ -1,4 +1,4 @@
-﻿/*
+/*
  * This file is subject to the terms and conditions defined in
  * file 'license.txt', which is part of this source code package.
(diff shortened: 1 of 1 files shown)


yaakov-h:
This looks like it just passes `null` around. Where does the actual auth happen?

oureveryday:
You need to call CMsgClientGetCDNAuthTokenResponse and get the CDN auth.
Only in Chinese CDNs it will return a non-empty CDN auth string.

xPaw:
For reference, CDN auth code was removed in #1022.

oureveryday:
You still need to use a CDN auth to download in Chinese CDNs.

xPaw:
I know, I am providing a reference as to where it was removed, and you can take a look at how it was implemented before.

You should add a xmldoc for the auth argument and how to get one (GetCDNAuthToken).

PR #1378, opened 2024-07-19
Make CryptoHelper internal and remove unused methods

xPaw:
This is a prerequisite cleanup for further optimizations (#1318), I don't think these crypto helpers need to be exposed as part of the library.

I've also split classes in Utils into their own files, and deleted WebHelpers as it is unused.

~~DepotDownloader uses CryptoHelper.SymmetricDecryptECB, but that's easy to copy over.~~ (done)

~~EDIT: Argh NetHookAnalyzer2 uses CryptoHelper.SymmetricDecrypt~~ (fixed)


xPaw:
I am currently moving HMAC methods to netfilter, and the only remaining method in CryptoHelper is SymmetricDecrypt, maybe I should leave it public (I've noticed steamdb uses it to decrypt file names).

PR #1379, opened 2024-07-19
Less allocations in non-utf8 string reads, fix non-utf8 string writes, optimize both with stack / pool alloc

kristiker:
* MsgBase string writes and reads default to `Encoding.UTF8` instead of `Encoding.GetEncoding( 0 )`
* Optimize allocations in WriteNullTermString, ReadNullTermString (non-utf8).
* Fix WriteNullTermString logic for non-utf8 encoding (doing ReadNullTermString back would not previously work).

### DIFF
--- a/SteamKit2/SteamKit2/Util/StreamHelpers.cs
+++ b/SteamKit2/SteamKit2/Util/StreamHelpers.cs
@@ -113,12 +113,32 @@ namespace SteamKit2
 
         public static void WriteNullTermString( this Stream stream, string value, Encoding encoding )
         {
-            var dataLength = encoding.GetByteCount( value );
-            var data = new byte[ dataLength + 1 ];
-            encoding.GetBytes( value, 0, value.Length, data, 0 );
-            data[ dataLength ] = 0x00; // '\0'
+            const string NullTerm = "\0";
+            value ??= string.Empty;
 
-            stream.Write( data, 0, data.Length );
+            var stringByteCount = encoding.GetByteCount( value );
+            var nullTermByteCount = encoding.GetByteCount( NullTerm );
+            var totalByteCount = stringByteCount + nullTermByteCount;
+
+            var isLargeBuffer = totalByteCount > 256;
+            var rented = isLargeBuffer ? ArrayPool<byte>.Shared.Rent( totalByteCount ) : null;
+
+            try
+            {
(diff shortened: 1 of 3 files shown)


xPaw:
GetEncoding( 0 ) should be whatever encoding is currently used by the machine, which is a really questionable choice, it probably should be UTF8.

SteamKit2/SteamKit2/Util/StreamHelpers.cs
```diff
@@ -130,12 +132,31 @@ private static string ReadNullTermUtf8String( Stream stream )
...
+            var stringByteCount = encoding.GetByteCount( value );
+            var nullTermByteCount = encoding.GetByteCount( NullTerminator );
+            var totalByteCount = stringByteCount + nullTermByteCount;
+
+            var isLargeBuffer = totalByteCount > 256;
+            var rented = isLargeBuffer ? ArrayPool<byte>.Shared.Rent( totalByteCount ) : null;
+
+            try
+            {
+                Span<byte> encodedSpan = isLargeBuffer ? rented.AsSpan( 0, totalByteCount ) : stackalloc byte[ totalByteCount ];
```
kristiker, line 146: Maybe can always use pooled array for less complexity.

xPaw:
Realistically looking at the code base, after changing `GetEncoding( 0 )` to UTF8, all code paths use UTF8, so I would just completely drop the code for the multi byte encodings (the utf8 optimized function can also be used for ASCII without problems).

SteamKit2/SteamKit2/Util/StreamHelpers.cs
```diff
@@ -64,28 +63,31 @@ public static float ReadFloat( this Stream stream )
...
 
             while ( true )
             {
-                byte[] data = new byte[ characterSize ];
-                stream.Read( data, 0, characterSize );
+                data.Clear();
+                stream.Read( data );
 
-                if ( encoding.GetString( data, 0, characterSize ) == "\0" )
+                if ( encoding.GetString( data ) == NullTerminator )
```
yaakov-h, line 85: Does this allocate a new string every time?

xPaw: Yeah. As I said above though, all the paths use UTF8 now so this won't even be hit.

PR #1381, opened 2024-07-22
Improve encryption allocations

### LINKED ISSUE #552: Crypto performance improvements

xPaw:
Supercedes #1378 (based on this pr)
Supercedes #1318
Supercedes #683
Fixes #552

This removes all public methods from `CryptoHelper` except for `SymmetricDecrypt` which is used by NHA and some other projects that use it for decrypting depots manifests/filenames.

I deleted `NetFilterEncryption` (non HMAC version) which does not appear to be used by Steam (I looked at node-steam-user code, they always skip the first 16 bytes in encrypted packets, so it must always be present).

I also changed envelope send to use arraypool instead of temporarily allocating a byte array for the encrypted message.


SteamKit2/SteamKit2/Networking/Steam3/WebSocketContext.cs
```diff
@@ -88,9 +88,8 @@ async Task RunCore(TimeSpan connectionTimeout, CancellationToken cancellationTok
                 }
             }
 
-            public async Task SendAsync(byte[] data)
+            public async Task SendAsync(ArraySegment<byte> segment)
             {
-                var segment = new ArraySegment<byte>(data, 0, data.Length);
                 try
                 {
                     await socket.SendAsync(segment, WebSocketMessageType.Binary, true, cts.Token).ConfigureAwait(false);
```
xPaw, line 95: This api does not have a Span<> overload forcing the use of ArraySegment everywhere, which I don't like. Any ideas?

SteamKit2/SteamKit2/Networking/Steam3/UdpConnection.cs
```diff
@@ -175,13 +175,13 @@ public void Disconnect( bool userInitiated )
...
-        /// <param name="data">The data to send to the server</param>
-        public void Send( byte[] data )
+        /// <param name="segment">The data to send to the server</param>
+        public void Send( ArraySegment<byte> segment )
         {
             if ( state != (int)State.Connected )
                 return;
 
-            using var ms = new MemoryStream( data );
+            using var ms = new MemoryStream( segment.ToArray() ); // TODO: Remove memorystream
```
xPaw, line 184: TODO

SteamKit2/SteamKit2/Networking/Steam3/EnvelopeEncryptedConnection.cs
```diff
@@ -47,14 +48,26 @@ public void Disconnect( bool userInitiated )
...
+                inner.Send( data );
+                return;
             }
 
-            inner.Send( data );
+            var buffer = ArrayPool<byte>.Shared.Rent( encryption!.CalculateMaxEncryptedDataLength( data.Length ) );
+            try
+            {
+                var length = encryption!.ProcessOutgoing( data.Span, buffer );
+                var segment = new Memory<byte>( buffer, 0, length );
```
yaakov-h, line 63:
```suggestion
                var segment = buffer.AsMemory();
```

xPaw: Need to use the length.

xPaw:
Related reading: https://github.com/dotnet/runtime/issues/55601

PR #1382, opened 2024-07-22
Optimize gzip in HandleMulti

xPaw:
An attempt to optimize it by not copying the entire decompressed buffer into byte array, but instead consume it directly.

### DIFF
--- a/SteamKit2/SteamKit2/Steam/CMClient.cs
+++ b/SteamKit2/SteamKit2/Steam/CMClient.cs
@@ -532,12 +532,12 @@ namespace SteamKit2.Internal
             {
                 Span<byte> length = stackalloc byte[ sizeof( int ) ];
 
-                while ( stream.Read( length ) == length.Length )
+                while ( stream.ReadAll( length ) > 0 )
                 {
                     var subSize = BitConverter.ToInt32( length );
                     var subData = new byte[ subSize ];
 
-                    var read = stream.ReadAll( subData );
+                    stream.ReadAll( subData );
 
                     if ( !OnClientMsgReceived( GetPacketMsg( subData, this ) ) )
                     {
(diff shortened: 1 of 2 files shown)


SteamKit2/SteamKit2/Steam/CMClient.cs
```diff
@@ -520,40 +520,33 @@ void HandleMulti( IPacketMsg packetMsg )
...
-            using var br = new BinaryReader( ms );
-            while ( ( ms.Length - ms.Position ) != 0 )
+            using ( stream )
             {
-                int subSize = br.ReadInt32();
-                byte[] subData = br.ReadBytes( subSize );
+                Span<byte> length = stackalloc byte[ sizeof( int ) ];
 
-                if ( !OnClientMsgReceived( GetPacketMsg( subData, this ) ) )
+                while ( stream.Read( length ) == length.Length )
```
xPaw, line 535: This is slightly sketchy, I don't know if there's a better way to know if gzip stream has reached the end.

yaakov-h: Isn't the standard contract `while (bytes_read > 0)`?

I think here it would be something like:

```cs
while ( stream.Read( length ) is > 0 and var dataLength )
{
    var data = length.Slice(0, dataLength);
    // ..
}
```

The problem there is you might not get the full set of length bytes in one read.

You might need to loop over it, subtracting any bytes read in a previous partial read?

SteamKit2/SteamKit2/Steam/CMClient.cs
```diff
@@ -520,40 +520,33 @@ void HandleMulti( IPacketMsg packetMsg )
...
                 {
-                    break;
+                    var subSize = BitConverter.ToInt32( length );
+                    var subData = new byte[ subSize ];
+
+                    var read = stream.ReadAll( subData );
```
yaakov-h, line 540: I'm slightly concerned about this unused variable too.

xPaw:
@yaakov-h Changed. I've noticed that StreamHelpers methods should likely use the `ReadAll` method too?

PR #1383, opened 2024-07-22
Use ArrayPool for reading websocket messages

xPaw:


SteamKit2/SteamKit2/Networking/Steam3/WebSocketContext.cs
```diff
@@ -117,65 +129,87 @@ public void Dispose()
...
             }
 
-            async Task<byte[]?> ReadMessageAsync(CancellationToken cancellationToken)
+            async Task<byte[]?> ReadMessageAsync( CancellationToken cancellationToken )
             {
-                using var ms = new MemoryStream();
-                var buffer = new byte[ 1024 ];
-                var segment = new ArraySegment<byte>( buffer );
+                var outputBuffer = ArrayPool<byte>.Shared.Rent( 1024 );
+                var readBuffer = ArrayPool<byte>.Shared.Rent( 1024 );
```
xPaw, line 135: Should we just use `stackalloc byte[1024]` here?

PR #1391, opened 2024-08-02
Make client msg handlers a simple list

xPaw:


SteamKit2/SteamKit2/Steam/SteamClient/SteamClient.cs
```diff
@@ -172,12 +174,7 @@ public void RemoveHandler( ClientMsgHandler handler )
         {
             Type type = typeof( T );
 
-            if ( handlers.Contains( type ) )
-            {
-                return handlers[ type ] as T;
-            }
-
-            return null;
+            return handlers.Find( h => h.GetType() == type ) as T;
```
xPaw, line 177: The `GetHandler` call is the only place where it realistically suffers, but I don't see why you need to be calling this method a lot.

SteamKit2/SteamKit2/Steam/SteamClient/SteamClient.cs
```diff
@@ -376,24 +373,21 @@ protected override bool OnClientMsgReceived( IPacketMsg? packetMsg )
             }
 
             // pass along the clientmsg to all registered handlers
-            foreach ( DictionaryEntry kvp in handlers )
+            foreach ( var value in handlers )
```
xPaw, line 376: This loop is the important code which is called a lot, and looking at OrderedDictionary implementation of enumerators, I don't think it's very efficient.

PR #1397, opened 2024-08-03
Use `BufferBlock` for the callbacks queue

xPaw:
Breaking change: this removes `freeLast` and `FreeLastCallback`, which silently changes the behaviour of overloads that had no parameters, perhaps they should be renamed.

I removed `GetAllCallbacks` from `SteamClient` and added `WaitForCallbackAsync(cts)`.

With `RunWaitCallbackAsync` consumers should be able to spin up tasks in TPL that run this in a loop without requering a thread for every manager (or having to loop over them).

@JustArchi

### DIFF
--- a/SteamKit2/SteamKit2/Steam/SteamClient/CallbackMgr/CallbackMgr.cs
+++ b/SteamKit2/SteamKit2/Steam/SteamClient/CallbackMgr/CallbackMgr.cs
@@ -44,7 +44,7 @@ namespace SteamKit2
         /// </summary>
         public void RunCallbacks()
         {
-            var call = client.GetCallback( true );
+            var call = client.GetCallback();
 
             if ( call == null )
                 return;
@@ -58,7 +58,7 @@ namespace SteamKit2
         /// <param name="timeout">The length of time to block.</param>
         public void RunWaitCallbacks( TimeSpan timeout )
         {
-            var call = client.WaitForCallback( true, timeout );
+            var call = client.WaitForCallback( timeout );
 
             if ( call == null )
                 return;
@@ -72,7 +72,7 @@ namespace SteamKit2
         /// <param name="timeout">The length of time to block.</param>
         public void RunWaitAllCallbacks( TimeSpan timeout )
         {
-            var calls = client.GetAllCallbacks( true, timeout );
+            var calls = client.GetAllCallbacks( timeout );
             foreach ( var call in calls )
             {
                 Handle( call );
--- a/SteamKit2/SteamKit2/Steam/SteamClient/CallbackMgr/CallbackMgr.cs
+++ b/SteamKit2/SteamKit2/Steam/SteamClient/CallbackMgr/CallbackMgr.cs
@@ -84,7 +84,12 @@ namespace SteamKit2
         /// </summary>
         public void RunWaitCallbacks()
         {
-            RunWaitCallbacks( TimeSpan.FromMilliseconds( -1 ) );
+            var call = client.WaitForCallback();
+
+            if ( call == null )
+                return;
+
+            Handle( call );
         }
 
         /// <summary>
--- a/SteamKit2/SteamKit2/Steam/SteamClient/CallbackMgr/CallbackMgr.cs
+++ b/SteamKit2/SteamKit2/Steam/SteamClient/CallbackMgr/CallbackMgr.cs
@@ -101,9 +101,9 @@ namespace SteamKit2
         }
         /// <summary>
         /// Blocks the current thread to run a single queued callback.
-        /// If no callback is queued, the method will block until one is posted.
+        /// If no callback is queued, the method will asynchronously await until one is posted.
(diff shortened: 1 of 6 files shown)


JustArchi (approved):
LGTM, after second thought, we don't need a single handling loop for all Steam clients - as long as waiting is asynchronous, and it is with buffer-block, then the effective number of threads busy with handling messages will only hit max when we have maximum overload, that is, more messages that we can handle in `while (true)`, which is not realistically possible.

Therefore, I expect at most a few threads busy for over a hundred of clients in practice.

SteamKit2/SteamKit2/Steam/SteamClient/CallbackMgr/CallbackMgr.cs
```diff
@@ -72,10 +76,14 @@ public void RunWaitCallbacks( TimeSpan timeout )
...
         {
-            var calls = client.GetAllCallbacks( true, timeout );
-            foreach ( var call in calls )
+            if ( !RunWaitCallbacks( timeout ) )
             {
-                Handle( call );
+                return;
+            }
+
+            while ( RunCallbacks() )
```
voided, line 84: This seems like a potentially surprising behavior change. IIRC the intent with the timeout was "run as many callbacks as we can within a short time period, then return".

I don't at all remember the use case, but I'm guessing it was for windows message loops, game frame loops, etc...

So previously you could get a rough time slice (ignoring all the quirks about OS thread scheduling) to process callbacks, but now this seems like it could infinitely block if the queue happens to grow at the same rate the callbacks are posted.

voided (approved):
LGTM other than my one comment.

I don't feel too strongly about it since we're introducing breaking changes anyway, but then we should update the xml docs IMO.

SteamKit2/SteamKit2/Steam/SteamClient/CallbackMgr/CallbackMgr.cs
xPaw: Look at the old implementation of GetAllCallbacks, that's how it worked too.

Wait once, then read whats in the queue. RunCallbacks is non blocking (it uses TryReceive).

xPaw:
@voided can you leave comments on which xmldocs need updating

SteamKit2/SteamKit2/Steam/SteamClient/CallbackMgr/CallbackMgr.cs
voided: The previous impl would wait up to the timeout to get a chunk of calls, then it would process them and return.

In the new impl, even if `RunCallbacks` is non-blocking, if the callback queue grows faster than this while loop can process - wouldn't it infinitely loop regardless of the timeout?

xPaw: Are you sure? It only did `Monitor.Wait` on the callback lock if the queue is empty.

```
if ( callbackQueue.Count == 0 )
if ( !Monitor.Wait( callbackLock, timeout ) )
```

voided: I think the old code was a less-than-ideal way to implement the concept in my OP.

If the queue is empty: `RunWaitAllCallbacks` waits until something arrives, and then likely just gives you a chunk of 1 callback, it processes it and returns.

Assuming the queue still had more callbacks, the next call would immediately hand you the next chunk without needing the timeout at all, we'd process that chunk and return.

The timeout was to ensure you didn't spend too much time waiting for callbacks.

I think we could have improved the throughput of the old code by pulsing the `callbackLock` after a batch of callbacks is inserted in the queue rather than pulsing on each individual insert...

voided:
> @voided can you leave comments on which xmldocs need updating

Just `RunWaitAllCallbacks`, depending on what we decide here.

SteamKit2/SteamKit2/Steam/SteamClient/CallbackMgr/CallbackMgr.cs
xPaw: Well as it stands, the new code should have the same behaviour unless I am missing something?

This is partially why I removed SteamClient.GetAllCallbacks, because the impl felt subpar to me. I wouldn't mind removing RunWaitAllCallbacks either.

SteamKit2/Tests/SteamGameServerFacts.cs
```diff
@@ -13,7 +13,7 @@ public void LogOnPostsLoggedOnCallbackWhenNoConnection()
                 Token = "SuperSecretToken"
             });
 
-            var callback = SteamClient.GetCallback( freeLast: true );
+            var callback = SteamClient.GetCallback( );
             Assert.NotNull( callback );
```
yaakov-h, line 17: Are these assertions still neccesary?

PR #1400, opened 2024-08-06
Apisurface

xPaw:
Copied from ValveKeyValue, do we want this?


xPaw:
@yaakov-h do we want this? if so, how do we make updating it easier without having to manually copy the file.

PR #1401, opened 2024-08-07
Use stackalloc in NetHelpers

ilyfairy:


xPaw:
Can NetworkToHostOrder/HostToNetworkOrder be used?

SteamKit2/SteamKit2/Util/NetHelpers.cs
```diff
@@ -9,36 +9,37 @@
...
 
         public static IPAddress GetIPAddress( uint ipAddr )
         {
-            byte[] addrBytes = BitConverter.GetBytes( ipAddr );
-            Array.Reverse( addrBytes );
-
+            Span<byte> addrBytes = stackalloc byte[ 4 ];
+            BitConverter.TryWriteBytes( addrBytes, IPAddress.NetworkToHostOrder( Unsafe.BitCast<uint, int>( ipAddr ) ) );
+            
             return new IPAddress( addrBytes );
```
xPaw, line 34: can't you construct from the int now without the byte array?

ilyfairy: has been changed to new IPAddress(long address)

yaakov-h:
Do we know if NetworkToHostOrder is correct here when running on arm64?

ilyfairy:
> Do we know if NetworkToHostOrder is correct here when running on arm64?


image

yaakov-h:
Do the Steam servers get the same value when they receive and parse the `CMsgIPAddress` though?

I care more about that host than the one running SK2.

xPaw:
We have tests for this and one of the CI runners is macos-14 arm64 and it passes, so probably fine.

xPaw:
Well this broke something:

> Unexpected exception propogated back to NetLoop: System.ArgumentOutOfRangeException: newAddress ('1844xxxxxxxxxxxxxx') must be less than or equal to '4294967295'. (Parameter 'newAddress')

at SteamKit2.NetHelpers.GetIPAddress(UInt32 ipAddr) in NetHelpers.cs:line 31

PR #1404, opened 2024-08-12
Optimize downloading depot chunks and manifests

xPaw:
This avoids one temporary allocation of the zip, I fixed up deserializing the manifest to not require stream seeking.

Ideally the http response would be directly streamed into deflate, but that's not possible because the zip metadata is at the end of the stream.

Maybe there's a way to read from DoRawCommandAsync into ArrayPool, that would improve things at least (hoping that steam cdns give correct content-length?).


SteamKit2/SteamKit2/Types/Manifest.cs
```diff
@@ -110,33 +110,17 @@ internal void Deserialize( BinaryReader ds )
...
-            using var ms = new MemoryStream( data );
-            using var br = new BinaryReader( ms );
-            Deserialize( br );
-        }
-
-        void Deserialize( BinaryReader ds )
+        internal void Deserialize( BinaryReader ds )
         {
+            /*
+            // The magic is verified by DepotManifest.InternalDeserialize, not checked here to avoid seeking
```
xPaw, line 116: This class is not public, and deserialize is only invoked from one code path, so it's fine.

SteamKit2/SteamKit2/Steam/CDN/Client.cs
```diff
@@ -83,9 +84,19 @@ public async Task<DepotManifest> DownloadManifestAsync( uint depotId, ulong mani
...
 
-            var depotManifest = new DepotManifest( manifestData );
+            DebugLog.Assert( entries.Count == 1, nameof( ZipUtil ), "Expected the zip to contain only one file" );
+
+            DepotManifest depotManifest;
+
+            using ( var zipEntryStream = entries[ 0 ].Open() )
+            {
+                depotManifest = DepotManifest.Deserialize( zipEntryStream );
+            }
```
yaakov-h, line 99: Could we simplify this into a helper function, something like:

```csharp
var depotManifest = ZipUtil.Decompress( manifestData, static zipEntryStream => DepotManifest.Deserialize( zipEntryStream ) );
```
?

SteamKit2/SteamKit2/Steam/CDN/Client.cs
```diff
@@ -83,9 +84,19 @@ public async Task<DepotManifest> DownloadManifestAsync( uint depotId, ulong mani
...
             var manifestData = await DoRawCommandAsync( server, url, proxyServer ).ConfigureAwait( false );
 
-            manifestData = ZipUtil.Decompress( manifestData );
+            // Decompress the zipped manifest data
+            using var ms = new MemoryStream( manifestData );
+            using var zip = new ZipArchive( ms );
+            var entries = zip.Entries;
 
-            var depotManifest = new DepotManifest( manifestData );
+            DebugLog.Assert( entries.Count == 1, nameof( ZipUtil ), "Expected the zip to contain only one file" );
```
yaakov-h, line 92: Nit: This refers to ZipUtil but we're in CDN.Client

SteamKit2/SteamKit2/Steam/CDN/Client.cs
xPaw: It would have to return DepotManifest from that lambda, seems wack.

yaakov-h: ZipUtil.Decompress<T>(..., Func<Stream, T>)

?

xPaw: I won't be able to check CRC anyway, is it really worth the couple lines? Also I'm going to push a commit that streams from http into a rented buffer, that unfortunately has some copypasta, so want your thoughts on the approach.

SteamKit2/SteamKit2/Steam/CDN/Client.cs
```diff
@@ -81,11 +83,60 @@ public async Task<DepotManifest> DownloadManifestAsync( uint depotId, ulong mani
...
 
-            manifestData = ZipUtil.Decompress( manifestData );
+            using var cts = new CancellationTokenSource();
+            cts.CancelAfter( RequestTimeout );
+
+            DepotManifest depotManifest;
+
+            try
+            {
+                using var response = await httpClient.SendAsync( request, HttpCompletionOption.ResponseHeadersRead, cts.Token ).ConfigureAwait( false );
```
xPaw, line 95: Because the cts and response are local, I can't just return the response.Content. The only other code path using DoRawCommandAsync is the chunk download, and that code should also stream.

The chunk download has other stupid things like the vzip decompress using LZMA which allocates a 1mb buffer whenever the decoder properties are set.

And the rented buffer would have to be copied if chunk is downloaded without giving a depot key.

SteamKit2/SteamKit2/Steam/CDN/Client.cs
```diff
@@ -81,11 +83,60 @@ public async Task<DepotManifest> DownloadManifestAsync( uint depotId, ulong mani
...
+                {
+                    using var ms = new MemoryStream( buffer, 0, contentLength );
+
+                    // Stream the http response into the rented buffer
+                    await response.Content.CopyToAsync( ms, cts.Token );
+
+                    ms.Position = 0;
+
+                    // Decompress the zipped manifest data
+                    using var zip = new ZipArchive( ms );
```
xPaw, line 122: ZipArchive needs a seekable stream fyi (otherwise it literally copies the stream into its own temp MemoryStream).

xPaw:
I've eliminated a bunch of allocations when downloading chunks (I had to modify the LZMA code because it allocates a 1mb buffer for every chunk).

I'm thinking of adding an overload to DownloadDepotChunkAsync that accepts a `byte[]` to write to, because the ChunkData already contains the uncompressed length for us, but we don't technically use it when decompressing the zip.

As I predict most use cases of downloading chunks is writing straight to disk, consumer could call it with a pooled buffer.

xPaw:
I've refactored downloading chunks to use a buffer provided by the consumer (which they can rent), as it's common to download a chunk and write it to disk straight away.

master:
image

this pr:
image

SteamKit2/SteamKit2/Util/VZipUtil.cs
```diff
@@ -45,26 +48,41 @@ public static byte[] Decompress(byte[] buffer)
...
+                using MemoryStream outStream = new MemoryStream( destination );
+                decoder.Code( ms, outStream, sizeCompressed, sizeDecompressed, null );
+            }
+            finally
+            {
+                ArrayPool<byte>.Shared.Return( windowBuffer );
+            }
 
-            if ( Crc32.HashToUInt32( outData ) != outputCRC )
+            if ( verifyChecksum && Crc32.HashToUInt32( destination.AsSpan()[ ..sizeDecompressed ] ) != outputCRC )
```
xPaw, line 77: The depot chunk itself has a checksum, so I don't see a reason to verify it here.

SteamKit2/SteamKit2/Util/VZipUtil.cs
```diff
@@ -45,26 +48,41 @@ public static byte[] Decompress(byte[] buffer)
...
+            }
+
             // jump back to the beginning of the compressed data
             ms.Position = compressedBytesOffset;
 
             SevenZip.Compression.LZMA.Decoder decoder = new SevenZip.Compression.LZMA.Decoder();
 
-            decoder.SetDecoderProperties( properties );
+            // If the value of dictionary size in properties is smaller than (1 << 12),
+            // the LZMA decoder must set the dictionary size variable to (1 << 12).
```
xPaw, line 62: This is not strictly necessary because a correct encoder on Valve's side should produce a valid dict size.

SteamKit2/SteamKit2/Util/LZMA/Compress/LZ/LzOutWindow.cs
```diff
@@ -14,6 +14,14 @@ class OutWindow
 
 		public uint TrainSize = 0;
 
+		internal void SteamKitSetBuffer(byte[] buffer, uint windowSize) // Added by SteamKit to avoid allocating a byte array
```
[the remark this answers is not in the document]
xPaw: lmza sdk hasn't been updated in a decade (and i don't think it will), and it already contains a modification for SK to allow invalid starts.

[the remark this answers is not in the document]
xPaw: Just for reference, after the potential bug I found: https://sourceforge.net/p/sevenzip/bugs/2502/

> C# and Java code of LZMA is old (2005-2008). And it was not updated from that time. [..] I'm not sure that I want to change old C# and Java now.

xPaw:
Here's how it looks like on depot downloader with this PR:

before:
image

after:
image

Looks like ValidateSteam3FileChecksums in DD also needs optimizing.

yaakov-h:
What do those memory graphs look like with [Server GC](https://learn.microsoft.com/en-us/dotnet/core/runtime-config/garbage-collector)?

xPaw:
downloading 5.88gb

before: (seems to have allocated 26gb total)
image

after:
image

xPaw:
@yaakov-h any objections on merging?

yaakov-h:
My only concern is that we seem to now be taking a hard dependency on the `Content-Length` header. If that ever disappears we now throw instead of degrading gracefully.

Other than that, if it works then :shipit:.

xPaw:
Well lets hope it doesn't lol. We can add a fallback later if it ever comes to be a problem.

PR #1406, opened 2024-08-26
Support downloading manifests/chunks with missing Content-Length

xPaw:
Chunks require CompressedLength to be set, which it should be.

### DIFF
--- a/SteamKit2/SteamKit2/Steam/CDN/Client.cs
+++ b/SteamKit2/SteamKit2/Steam/CDN/Client.cs
@@ -224,17 +224,25 @@ namespace SteamKit2.CDN
                     throw new SteamKitWebRequestException( $"Response status code does not indicate success: {response.StatusCode:D} ({response.ReasonPhrase}).", response );
                 }
 
-                if ( !response.Content.Headers.ContentLength.HasValue )
+                var contentLength = ( int )chunk.CompressedLength;
+
+                if ( response.Content.Headers.ContentLength.HasValue )
                 {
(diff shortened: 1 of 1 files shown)


SteamKit2/SteamKit2/Steam/CDN/Client.cs
```diff
@@ -100,30 +100,45 @@ public async Task<DepotManifest> DownloadManifestAsync( uint depotId, ulong mani
...
+                        {
+                            throw new InvalidDataException( $"Length mismatch after downloading depot manifest! (was {ms.Position}, but should be {contentLength})" );
+                        }
+
+                        ms.Position = 0;
+                    }
+                    else
                     {
-                        throw new InvalidDataException( $"Length mismatch after downloading depot manifest! (was {ms.Position}, but should be {contentLength})" );
+                        var data = await response.Content.ReadAsByteArrayAsync();
```
yaakov-h, line 138: Are we avoiding `ReadAsStreamAsync` because we need to be able to seek?

xPaw: ZipArchive(Stream) is a lie because it will copy the stream into an internal MemoryStream if stream is not seekable: https://source.dot.net/#System.IO.Compression/System/IO/Compression/ZipArchive.cs,138

yaakov-h: lol

yaakov-h (approved):

PR #1409, opened 2024-08-30
Use `LibraryImport` and CsWin32

xPaw:
I added a test that checks for non-default values in the real providers, so hoping the CI will catch issues. I will verify against real Steam client on Windows in a bit.

I also want to move these classes into separate files in a separate PR.


xPaw:
Well I guess CI machines don't have enough hardware to pass it 😄

SteamKit2/SteamKit2/Util/MacHelpers.cs
```diff
@@ -89,12 +89,14 @@ public enum CFStringEncoding : uint
             kCFStringEncodingASCII = 0x0600
         }
 
-        [DllImport(LibraryName, CallingConvention = CallingConvention.Cdecl)]
-        public static extern void CFRelease(IntPtr cf);
+        [LibraryImport( LibraryName )]
+        [UnmanagedCallConv( CallConvs = [ typeof( System.Runtime.CompilerServices.CallConvCdecl ) ] )]
```
yaakov-h, line 93: Can we move this namespace into a `using`?

SteamKit2/SteamKit2/Util/Win32Helpers.cs
```diff
@@ -165,24 +166,25 @@ protected override bool ReleaseHandle()
...
 		{
 			#region CreateFile
 
 			public static uint FILE_SHARE_READ = 1;
 			public static uint FILE_SHARE_WRITE = 2;
 			public static uint OPEN_EXISTING = 3;
 
-			[DllImport( "kernel32.dll", SetLastError = true, CharSet = CharSet.Unicode )]
-			public static extern FileSafeHandle CreateFile( string lpFileName, uint dwDesiredAccess, uint dwShareMode, IntPtr lpSecurityAttributes, uint dwCreationDisposition, uint dwFlagsAndAttributes, IntPtr hTemplateFile );
+			[LibraryImport( "kernel32.dll", EntryPoint = "CreateFileW", SetLastError = true, StringMarshalling = StringMarshalling.Utf16 )]
```
yaakov-h, line 177: The Windows APIs should probably use [CsWin32](https://github.com/microsoft/CsWin32) rather than having us hand-roll the P/Invokes.

xPaw: If only this was a thing for mac

yaakov-h: I've been poking at the edges of it for macOS and libc but need to figure out how they get the definitions. I might open an issue.

SteamKit2/SteamKit2/Util/Win32Helpers.cs
```diff
@@ -1,258 +1,168 @@
...
+            if ( handle == null || handle.IsInvalid )
+            {
+                throw new FileNotFoundException( "Unable to open boot volume." );
+            }
+
+            var extents = new VOLUME_DISK_EXTENTS();
+
+            if ( !PInvoke.DeviceIoControl(
+                handle,
+                PInvoke.IOCTL_VOLUME_GET_VOLUME_DISK_EXTENTS,
```
yaakov-h, line 61: I would be tempted to `using static PInvoke;` and have this code look a little more like the C/C++ code it is emulating.

SteamKit2/SteamKit2/Util/Win32Helpers.cs
```diff
@@ -1,258 +1,168 @@
...
+                null,
+                0,
+                &extents,
+                ( uint )VOLUME_DISK_EXTENTS.SizeOf( 1 ),
+                null,
+                null
+            ) )
+            {
+                throw new InvalidOperationException( $"Failed to get volume disk extents: {Marshal.GetLastWin32Error()}" );
```
yaakov-h, line 70: Why not just Win32Exception? It already includes the last Win32 error and the stack trace should tell us where it went wrong.

xPaw: It does? I saw dotnet code has an internal code to throw win32 errors after invokes, but didn't check this exception.

yaakov-h: https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/ComponentModel/Win32Exception.cs,26

PR #1414, opened 2024-08-31
Fix typo

Citrinate:
I haven't been able to get responses from messages sent to GameCoordinator since 3.0.0-Alpha.3.  In the 022_DotaMatchRequest sample, nothing currently happens after launching DOTA.  Narrowed the cause down to what looks to be just a typo.  Correcting it resolves my issue.


xPaw:
Thanks.

PR #1416, opened 2024-09-01
Implement getting mac address on linux

xPaw:
Mostly replicates what Windows version does, I can't verify if Steam's actual Linux version produces same result as we have no nethook for Linux.

### DIFF
--- a/SteamKit2/SteamKit2/Util/HardwareUtils.cs
+++ b/SteamKit2/SteamKit2/Util/HardwareUtils.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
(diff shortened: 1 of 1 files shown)


SteamKit2/SteamKit2/Util/HardwareUtils.cs
```diff
@@ -157,12 +164,8 @@ sealed class LinuxMachineInfoProvider : IMachineInfoProvider
         {
             string[] machineFiles =
             [
-                "/etc/machine-id", // present on at least some gentoo systems
                 "/var/lib/dbus/machine-id",
```
xPaw, line 172: Steam only reads '/var/lib/dbus/machine-id' and '~/.steam/machine-id'

PR #1420, opened 2024-09-01
Changing default Protocol from TCP to WebSockets. 

tpill90:
After Valve's maintenance on 2024/08/27 some servers in affected regions will no longer accept TCP connections, which can prevent clients from ever making a successful connection in some cases.  

For example CellId 63 (Washington DC) will endlessly retry the same servers without ever making a successful connection.  I have tested and confirmed that other regions are affected by this issue, however I don't remember which ones off the top of my head.
image

### DIFF
--- a/SteamKit2/SteamKit2/Steam/SteamClient/Configuration/SteamConfigurationBuilder.cs
+++ b/SteamKit2/SteamKit2/Steam/SteamClient/Configuration/SteamConfigurationBuilder.cs
@@ -35,7 +35,7 @@ namespace SteamKit2
 
                 MachineInfoProvider = MachineInfoProvider.GetDefaultProvider(),
 
-                ProtocolTypes = ProtocolTypes.Tcp,
+                ProtocolTypes = ProtocolTypes.WebSocket,
 
                 ServerListProvider = new MemoryServerListProvider(),
 
--- a/SteamKit2/SteamKit2/Steam/SteamClient/Configuration/SteamConfigurationBuilder.cs
+++ b/SteamKit2/SteamKit2/Steam/SteamClient/Configuration/SteamConfigurationBuilder.cs
@@ -35,7 +35,7 @@ namespace SteamKit2
 
                 MachineInfoProvider = MachineInfoProvider.GetDefaultProvider(),
 
-                ProtocolTypes = ProtocolTypes.WebSocket,
+                ProtocolTypes = ProtocolTypes.Tcp | ProtocolTypes.WebSocket,
 
                 ServerListProvider = new MemoryServerListProvider(),
 
--- a/SteamKit2/Tests/SteamConfigurationFacts.cs
+++ b/SteamKit2/Tests/SteamConfigurationFacts.cs
@@ -79,7 +79,7 @@ namespace Tests
         [Fact]
         public void DefaultProtocols()
         {
-            Assert.Equal(ProtocolTypes.Tcp, configuration.ProtocolTypes);
+            Assert.Equal(ProtocolTypes.Tcp | ProtocolTypes.WebSocket, configuration.ProtocolTypes);
         }
 
         [Fact]


yaakov-h:
That also suggests a secondary bug in the server selection logic, if we can't connect to one server we should be trying again on another.

SteamKit2/SteamKit2/Steam/SteamClient/Configuration/SteamConfigurationBuilder.cs
```diff
@@ -35,7 +35,7 @@ public static SteamConfigurationState CreateDefaultState()
 
                 MachineInfoProvider = MachineInfoProvider.GetDefaultProvider(),
 
-                ProtocolTypes = ProtocolTypes.Tcp,
+                ProtocolTypes = ProtocolTypes.WebSocket,
```
Lifeismana, line 38:
```suggestion
                ProtocolTypes = ProtocolTypes.Tcp | ProtocolTypes.WebSocket,
``` since not all the tcp servers are bad

**Comment by tpill90** [comment]
> That also suggests a secondary bug in the server selection logic, if we can't connect to one server we should be trying again on another.

The retry logic is definitely working as intended,  I can see that the servers are being iterated over as they fail.  Part of the issue here might be due to `GetCMList` almost always returning servers in the exact same order,  such that each time the client tries to connect it will have to traverse the list of always failing connections each time to eventually land on the remaining few good servers.

Taking a look at the logs from my Steam client it doesn't seem like the TCP servers are even considered, it just goes ahead and picks WebSockets : 
```
[2024-09-02 00:29:28] GetCMListForConnect -- got 4 Netfilter CMs and 76 WebSocket CMs
[2024-09-02 00:29:28] GetCMListForConnect -- DC 'iad1' count: 23
[2024-09-02 00:29:28] GetCMListForConnect -- DC 'atl3' count: 21
[2024-09-02 00:29:28] GetCMListForConnect -- DC 'ord1' count: 21
[2024-09-02 00:29:28] GetCMListForConnect -- DC 'sea1' count: 15
[2024-09-02 00:29:28] [Connecting, 4, 0] [U:1:9680370] CM Directory list says 95% of connections should be websockets, we rolled 58 - using WebSockets as default.
[2024-09-02 00:29:28] [Connecting, 4, 0] [U:1:9680370] PingWebSocketCM() (ext4-iad1.steamserver.net:27036) starting...
[2024-09-02 00:29:28] [Connecting, 4, 0] [U:1:9680370] PingWebSocketCM() (cmp1-iad1.steamserver.net:443) starting...
[2024-09-02 00:29:28] [Connecting, 4, 0] [U:1:9680370] PingWebSocketCM() (ext3-iad1.steamserver.net:27023) starting...
[2024-09-02 00:29:28] [Connecting, 4, 0] [U:1:9680370] PingWebSocketCM() (ext2-atl3.steamserver.net:443) starting...
[2024-09-02 00:29:29] [Connecting, 4, 0] [U:1:9680370] PingWebSocketCM() (ext4-iad1.steamserver.net:27036) results: 87.7416ms + load 29.0000
[2024-09-02 00:29:29] [Connecting, 4, 0] [U:1:9680370] PingWebSocketCM() stopping after 400.00ms timeout and using good enough host (ext4-iad1.steamserver.net:27036)
[2024-09-02 00:29:29] [Connecting, 4, 0] [U:1:9680370] PingWebSocketCM() (ext3-iad1.steamserver.net:27023) results: 87.9958ms + load 32.0000
[2024-09-02 00:29:29] [Connecting, 4, 0] [U:1:9680370] PingWebSocketCM() (cmp1-iad1.steamserver.net:443) results: 128.5093ms + load 31.0000
[2024-09-02 00:29:29] [Connecting, 4, 0] [U:1:9680370] PingWebSocketCM() (ext2-atl3.steamserver.net:443) results: 124.1386ms + load 39.0000
[2024-09-02 00:29:29] [Connecting, 4, 7] [U:1:9680370] Connect() starting connection (eNetQOSLevelHigh, ext4-iad1.steamserver.net:27036, WebSocket)
[2024-09-02 00:29:29] [Connecting, 4, 7] [U:1:9680370] ConnectionCompleted() (162.254.192.87:27036, WebSocket) local address (192.168.1.55:54600)
```

Is Valve trying to decommission their TCP servers and move over to WebSockets exclusively?  It would make sense that after server maintenance that some servers might no longer be accepting new TCP connections.

xPaw:
I thought the default was All instead of force TCP.

yaakov-h:
SteamClient has defaulted to TCP as far back as I remember and as far back as I can find code for, but I thought we changed that more recently. I suppose not.

And yes, I do believe that the long term plan is to migrate all of Steam over to WebSockets and dump the TCP and UDP implementations.

PR #1422, opened 2024-09-02
Switch to `GetCMListForConnect`

### LINKED ISSUE #1419: Use GetCMListForConnect in SteamDirectory
Steam client uses this I believe, and it returns websocket servers along with tcp in same list, so the order is better.

xPaw:
Fixes #1419

### DIFF
--- a/SteamKit2/SteamKit2/Steam/WebAPI/SteamDirectory.cs
+++ b/SteamKit2/SteamKit2/Steam/WebAPI/SteamDirectory.cs
@@ -57,39 +57,40 @@ namespace SteamKit2
 
             cancellationToken.ThrowIfCancellationRequested();
 
-            var response = await directory.CallAsync( HttpMethod.Get, "GetCMList", version: 1, args: args ).ConfigureAwait( false );
+            var response = await directory.CallAsync( HttpMethod.Get, "GetCMListForConnect", version: 1, args: args ).ConfigureAwait( false );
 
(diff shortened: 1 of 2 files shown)


SteamKit2/SteamKit2/Steam/WebAPI/SteamDirectory.cs
```diff
@@ -57,39 +57,40 @@ static async Task<IReadOnlyCollection<ServerRecord>> LoadCoreAsync( SteamConfigu
...
                     continue;
                 }
 
-                serverRecords.Add( ServerRecord.CreateWebSocketServer( child.Value ) );
-            }
+                var record = child[ "type" ].Value switch
+                {
+                    "websockets" => ServerRecord.CreateWebSocketServer( endpoint ),
+                    "netfilter" => ServerRecord.CreateDnsSocketServer( endpoint ),
+                    // TODO: There's also "china" type which uses websocket
```
yaakov-h, line 86: we might need to figure this out to keep DepotDownloader operational there

yaakov-h (approved):

SteamKit2/SteamKit2/Steam/WebAPI/SteamDirectory.cs
xPaw: I suspect these servers only come if you specify china realm (but I just tried and didn't get that anyway).

PR #1423, opened 2024-09-04
Adding Lancache support to the CDN Client.

tpill90:
Added automatic Lancache detection to the CDN client.  If a Lancache is on the local network and the trigger domain is poisoned, then SteamKit will direct all traffic through the cache instead of Valve's CDNs.

This behavior should match what the current behavior in Steam is.


SteamKit2/SteamKit2/Steam/CDN/Client.cs
```diff
@@ -227,10 +232,31 @@ public async Task<int> DownloadDepotChunkAsync( uint depotId, DepotManifest.Chun
...
+
             var chunkID = Utils.EncodeHexString( chunk.ChunkID );
             var url = $"depot/{depotId}/chunk/{chunkID}";
 
-            using var request = new HttpRequestMessage( HttpMethod.Get, BuildCommand( server, url, cdnAuthToken, proxyServer ) );
+            var builtUrl = BuildCommand( server, url, cdnAuthToken, proxyServer );
+            using var request = new HttpRequestMessage( HttpMethod.Get, builtUrl );
+            if ( LancacheDetected )
+            {
+                request.Headers.Host = server.Host;
```
yaakov-h, line 256: This looks like a bit of a dodgy approach to me.

I think it would be better to override the [ConnectCallback](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.socketshttphandler.connectcallback?view=net-8.0) so that the HTTP request is exactly the same as it would be to the CDN, we just connect to the lancache server instead of the CDN.

That would also help minimize the amount of lancache-specific branching logic here.

tpill90: I'm not very familiar with ConnectCallback so I'm not exactly sure how to implement this change using it.  Maybe I can provide some more details on why I implemented it this way and you can direct me where to start. 

The lancache instance is ultimately just an nginx server with some game cdn specific config. For pretty much every other cdn we just poison the appropriate domain names, and nginx will automatically forward upstream requests to the correct server.  In this case the client applications have no idea that anything is running through the nginx cache. 
Steam however is different in that they provided us a special domain `lancache.steamcontent.com` which doesn't **technically** point to one of their cdns, it's expected to always resolve to a local ip if there is a cache present.  Because of there being no actual upstream at `lancache.steamcontent.com` Nginx will need to know which upstream server to use instead, which is specified in the required Host header.  The other piece of the puzzle is that the user agent must absolutely match the one that Steam uses, otherwise requests won't complete successfully. 

As long as we can make sure those two headers are set correctly then we should be good to go with the implementation you're suggesting

SteamKit2/SteamKit2/Steam/CDN/LancacheDetector.cs
```diff
@@ -0,0 +1,86 @@
...
+        /// <param name="toTest">The IP address that will be tested</param>
+        /// <returns>Returns true if the IP is a private address, false if it isn't private</returns>
+        private static bool IsPrivateAddress( this IPAddress toTest )
+        {
+            if ( IPAddress.IsLoopback( toTest ) )
+            {
+                return true;
+            }
+
+            byte[] bytes = toTest.GetAddressBytes();
```
xPaw, line 72: ipv6 support?

SteamKit2/SteamKit2/Steam/CDN/LancacheDetector.cs
```diff
@@ -0,0 +1,86 @@
...
+            }
+            return false;
+        }
+
+        /// <summary>
+        /// Determines if an IP address is a private address, as specified in RFC1918
+        /// </summary>
+        /// <param name="toTest">The IP address that will be tested</param>
+        /// <returns>Returns true if the IP is a private address, false if it isn't private</returns>
+        private static bool IsPrivateAddress( this IPAddress toTest )
```
xPaw, line 65: lets not make this an extension

SteamKit2/SteamKit2/Steam/CDN/Client.cs
xPaw: ConnectCallback should allow us to replace the IP it connects to (the local one in this case), without manually changing the host header.

SteamKit2/SteamKit2/Steam/CDN/LancacheDetector.cs
```diff
@@ -0,0 +1,86 @@
...
+            }
+
+            // DNS hostnames can possibly resolve to more than one IP address (one-to-many), so we must check each one for a Lancache server
+            foreach (var ip in ipAddresses)
+            {
+                try
+                {
+                    // If the IP resolves to a private subnet, then we want to query the Lancache server to see if it is actually there.
+                    // Requests that are served from the cache will have an additional header.
+                    var response = await httpClient.GetAsync(new Uri($"http://{ip}/lancache-heartbeat"));
```
xPaw, line 45: This seems like more extra logic that Steam wouldn't do.

SteamKit2/SteamKit2/Steam/CDN/Client.cs
```diff
@@ -227,10 +232,31 @@ public async Task<int> DownloadDepotChunkAsync( uint depotId, DepotManifest.Chun
                 }
             }
 
+            await _lancacheDetectionLock.WaitAsync();
```
xPaw, line 235: Putting this check inside of depot download wont perform for manifest downloads.

But my initial idea was to add this as a separate public method so consumers can chose to check/enable for lancache support, instead of it happening automagically.

SteamKit2/SteamKit2/Steam/CDN/Client.cs
```diff
@@ -227,10 +232,31 @@ public async Task<int> DownloadDepotChunkAsync( uint depotId, DepotManifest.Chun
...
             var url = $"depot/{depotId}/chunk/{chunkID}";
 
-            using var request = new HttpRequestMessage( HttpMethod.Get, BuildCommand( server, url, cdnAuthToken, proxyServer ) );
+            var builtUrl = BuildCommand( server, url, cdnAuthToken, proxyServer );
+            using var request = new HttpRequestMessage( HttpMethod.Get, builtUrl );
+            if ( LancacheDetected )
+            {
+                request.Headers.Host = server.Host;
+                // User agent must match the Steam client in order for Lancache to correctly identify and cache Valve's CDN content
+                request.Headers.Add( "User-Agent", "Valve/Steam HTTP Client 1.0" );
```
xPaw, line 258: Do you mind linking me code in lancache that checks for this? I can't find it.

Lifeismana: seems like it's nginx that handles that
although it's in 2 places :thinking: 
https://github.com/lancachenet/monolithic/blob/d01c26ae0d92ba643017791a213573d0f54e0758/overlay/etc/nginx/conf.d/30_maps.conf#L2
https://github.com/lancachenet/monolithic/blob/d01c26ae0d92ba643017791a213573d0f54e0758/overlay/hooks/entrypoint-pre.d/15_generate_maps.sh#L27

xPaw: Okay, so we should be able to at least append (SteamKit) to the user agent here

SteamKit2/SteamKit2/Steam/CDN/Client.cs
Dimensional: This is a good thing to be honest. Since manifests are downloaded with request codes appended to the URL, if they were to be passed through lancache, it could in theory make the same manifest be downloaded and saved multiple times, as lancache uses the complete url string to tell if it's already saved or a new file.
In short, new request code = new url = new file.

xPaw: Does Steam forward manifest requests to lancache?

SteamKit2/SteamKit2/Steam/CDN/LancacheDetector.cs
tpill90: Yeah, this is not something that Steam does.  This avoids the config error of where there is no actual cache server pointing at the IP that `lancache.steamcontent.com` resolves to, which causes Steam to fail on every request.

If we're shooting for 1:1 behavior match with Steam then I'll drop this.  Would simplify other things too.

SteamKit2/SteamKit2/Steam/CDN/Client.cs
tpill90: Yes it does.

Dimensional: It appears so. I checked my access logs in lancache, and it seems that even the Steamdeck, if it detects lancache, will send the request through there. I have snippet saved to an ods spreadsheet if you want to review it. Can send it to you via Discord.

SteamKit2/SteamKit2/Steam/CDN/LancacheDetector.cs
xPaw: Yes I don't think it's our problem. And it can be solved by making it optional (by having a call to enable lancache).

tpill90: So drop the whole class or just that piece of functionality

SteamKit2/SteamKit2/Steam/CDN/LancacheDetector.cs
tpill90: Good catch. Ipv6 is supported. I'll fix this.

SteamKit2/SteamKit2/Steam/CDN/Client.cs
tpill90: Tested appending earlier and it does not work. Instead an exact match is expected, which I believe is what the leading tilde signifies

SteamKit2/SteamKit2/Steam/CDN/Client.cs
tpill90: If we implement pointing to the lancache ip in ConnectCallback we will still need to pass the upstream cdn in the host header, otherwise the cache will just fail the request. Since we no longer have the original domain in the request (just the ip) nginx will need to be explicitly told what the upstream is.   I can show some example requests that demonstrate this behavior if you'd like.

SteamKit2/SteamKit2/Steam/CDN/Client.cs
```diff
@@ -227,10 +230,24 @@ public async Task<int> DownloadDepotChunkAsync( uint depotId, DepotManifest.Chun
                 }
             }
 
+            if ( !CheckedForLancacheServer )
```
xPaw, line 233: I still would prefer this to be an explicit separate call a consumer needs to call.

SteamKit2/SteamKit2/Steam/CDN/LancacheDetector.cs
```diff
@@ -0,0 +1,64 @@
...
+
+namespace SteamKit2.Steam.CDN
+{
+    /// <summary>
+    /// Attempts to automatically resolve the Lancache's IP address in the same manner that the Steam client does.
+    ///
+    /// Will automatically try to detect the Lancache through the poisoned DNS entries.
+    /// This is a modified version from the original source : https://github.com/tpill90/lancache-prefill-common/blob/main/dotnet/LancacheIpResolver.cs
+    /// </summary>
+    public static class LancacheDetector
```
xPaw, line 18: Make `CDN.Client` partial, and then extend it here, then you can put all the lancache code in the same file.

SteamKit2/SteamKit2/Steam/CDN/LancacheDetector.cs
xPaw: also would be nice to have a test unit for this function.

Dimensional:
This has been sitting here for a while, but I realized something that needs to be taken into consideration is that if the LanCache has a bad file saved, like a file got corrupted in storage or network issues resulted in a 0-length file, this setup would constantly request the file over and over again, constantly getting the same bad file and delaying downloads.

This is a flaw even with Steam when it grabs from LanCache, and requires reviewing the access logs, identify the files being requested repeatedly, and then have to either locate the cached files and delete them, or do what steam-lancache-prefill does and append `?nocache=1` to the end of the url string in order to force LanCache to discard what it has saved and grab from upstream again.

SteamKit2/SteamKit2/Steam/CDN/LancacheDetector.cs
tpill90: @xPaw I finally got back to looking at this.  Broke the lancache specific code out into this file, and made it an explicit flag that needs to be enabled in order to use the lancache.  Can you take another look and check if the current code is closer to what you were thinking?

SteamKit2/SteamKit2/Steam/CDN/Client.cs
```diff
@@ -230,18 +229,26 @@ public async Task<int> DownloadDepotChunkAsync( uint depotId, DepotManifest.Chun
...
+                request = BuildLancacheRequest( server, url, cdnAuthToken );
+            }
+
             using var cts = new CancellationTokenSource();
             cts.CancelAfter( RequestTimeout );
 
             try
             {
-                using var response = await httpClient.SendAsync( request, HttpCompletionOption.ResponseHeadersRead, cts.Token ).ConfigureAwait( false );
+                using var response = await httpClient
```
xPaw, line 244: undo the unrelated formatting changes pls

SteamKit2/SteamKit2/Steam/CDN/ClientLancache.cs
```diff
@@ -0,0 +1,40 @@
...
+            {
+                Scheme = "http",
+                Host = "lancache.steamcontent.com",
+                Port = 80,
+                Path = command,
+                Query = query ?? string.Empty
+            };
+
+            var request = new HttpRequestMessage( HttpMethod.Get, builder.Uri );
+            if ( UseLancacheServer )
```
xPaw, line 30: you can't get here without this being true

SteamKit2/SteamKit2/Steam/CDN/Client.cs
```diff
@@ -230,18 +229,26 @@ public async Task<int> DownloadDepotChunkAsync( uint depotId, DepotManifest.Chun
             var chunkID = Utils.EncodeHexString( chunk.ChunkID );
             var url = $"depot/{depotId}/chunk/{chunkID}";
 
-            using var request = new HttpRequestMessage( HttpMethod.Get, BuildCommand( server, url, cdnAuthToken, proxyServer ) );
-
+            var builtUrl = BuildCommand( server, url, cdnAuthToken, proxyServer );
+            var request = new HttpRequestMessage( HttpMethod.Get, builtUrl );
+            if ( UseLancacheServer )
```
xPaw, line 234: the two calls above shouldn't happen for lancache then, just being overriden.

xPaw:
I thought the method to detect lancache automatically was fine, by doing a look up. Just make it assign result into  UseLancacheServer if available (and leave calling the method to consumers).

SteamKit2/SteamKit2/Steam/CDN/Client.cs
tpill90: Split it out into an if-else

SteamKit2/SteamKit2/Steam/CDN/ClientLancache.cs
tpill90: Thanks, good catch.

SteamKit2/SteamKit2/Steam/CDN/Client.cs
tpill90: Sorry, didn't catch that my editor did that.  Reverted.

tpill90:
> I thought the method to detect lancache automatically was fine, by doing a look up. Just make it assign result into UseLancacheServer if available (and leave calling the method to consumers).

Put the lancache detection into a method for users to call, should match what you were expecting now.

SteamKit2/SteamKit2/Steam/CDN/ClientLancache.cs
```diff
@@ -0,0 +1,109 @@
...
+        public static bool UseLancacheServer { get; private set; }
+
+        private static string TriggerDomain = "lancache.steamcontent.com";
+
+        /// <summary>
+        /// Attempts to automatically resolve a Lancache on the local network.  If detected, SteamKit will route all downloads through the cache
+        /// rather than through Steam's CDN.
+        ///
+        /// Will try to detect the Lancache through the poisoned DNS entries, however if that is not possible it will then check
+        /// 'localhost' to see if the Lancache is available locally.  If the server is not available on 'localhost', then 172.17.0.1 will be checked to see if
```
xPaw, line 29: that's not happening here, is it?

tpill90:
@xPaw Should I squash this branch prior to merging now that it's approved?

xPaw:
We can squash on merge. Did you test this pr with Depot downloader?

SteamKit2/SteamKit2/Steam/CDN/Client.cs
```diff
@@ -16,7 +16,7 @@ namespace SteamKit2.CDN
     /// <summary>
     /// The <see cref="Client"/> class is used for downloading game content from the Steam servers.
     /// </summary>
-    public sealed class Client : IDisposable
+    public partial class Client : IDisposable
```
xPaw, line 19: keep sealed here

tpill90: Done.

tpill90:
Fixed the latest comment, rebased, and pushed again.

tpill90:
@yaakov-h You mind giving this a review when you have a moment?  We should be good to merge at this point.

PR #1425, opened 2024-09-05
Fix Protobufs path case-sensitivity

rastla:
The folder is named "Protobufs", so if you execute this script in linux, it will fail with "ProtoBufs"


voided:
Thanks!

PR #1426, opened 2024-09-05
Add Deadlock protobufs to generator csv

rastla:
Not sure if all of those will really be needed, but at least they compile.
Also wasn't sure about the naming scheme for all of them, I mostly tried to name it like the Dota 2 Protobufs, but since some are different obviously


xPaw:
I'm not sure we should be adding the game protos to the library anymore. People would probably be better served by separate libraries that target a specific game which can generate the protos and helper methods.

psychonic:
Superseded by #1539

PR #1428, opened 2024-09-10
Fix CallbackManager to be thread-safe

xPaw:
I think Add/Remove methods allocating new arrays is a worthwhile cost versus allocating in Handle.

To be extra safe I could add lock in add/remove.


yaakov-h:
See #1431 for my feedback.

PR #1432, opened 2024-09-15
Use System.TimeProvider to control time

yaakov-h:
This PR introduces use of TimeProvider (hooray for `net8.0`) to control time.

This can be configured via `SteamConfiguration` and allows for time to be slowed, sped up, paused, skipped, etc. to make some functions easier to test.

### DIFF
--- a/SteamKit2/SteamKit2/Steam/CMClient.cs
+++ b/SteamKit2/SteamKit2/Steam/CMClient.cs
@@ -142,10 +142,12 @@ namespace SteamKit2.Internal
 
             ID = identifier;
 
-            heartBeatFunc = new ScheduledFunction( () =>
-            {
-                Send( new ClientMsgProtobuf<CMsgClientHeartBeat>( EMsg.ClientHeartBeat ) );
-            } );
+            heartBeatFunc = new ScheduledFunction(
+                configuration.TimeProvider,
+                () =>
+                {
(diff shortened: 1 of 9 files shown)


yaakov-h:
System.Threading.Tasks.Dataflow doesn't seem to support this just yet so I can't easily hook it in to callback timeouts. I think.

SteamKit2/SteamKit2/Steam/CMClient.cs
```diff
@@ -142,10 +142,12 @@ public CMClient( SteamConfiguration configuration, string identifier )
 
             ID = identifier;
 
-            heartBeatFunc = new ScheduledFunction( () =>
-            {
-                Send( new ClientMsgProtobuf<CMsgClientHeartBeat>( EMsg.ClientHeartBeat ) );
-            } );
+            heartBeatFunc = new ScheduledFunction(
```
xPaw, line 145: Maybe this can be replaced with PeriodicTimer while we're at it

yaakov-h: I was hoping to not have to do any significant rewriting in this PR.

xPaw:
> System.Threading.Tasks.Dataflow doesn't seem to support this just yet

Is this for `WaitForCallback( TimeSpan timeout )`? Can it be worked around by testing `WaitForCallbackAsync( CancellationToken cancellationToken = default )` instead?

PR #1433, opened 2024-09-18
[RFC] Remove UnifiedMessages reflection usage

affederaffe:
This PR tries to remove the last directly existing reflection usage in SteamUnifiedMessages.
The protobuf code generator was updated instead to directly handle the messaging.
Obviously this is a breaking change, but, at least in my opinion, necessary in order to fully remove reflection.


xPaw:
I like the idea. If you enable the trimming warnings, what warnings are reported? In particular handling of incoming messages.

Resources/ProtobufGen/ProtobufGen/SteamKitCSharpCodeGenerator.cs
```diff
@@ -37,20 +32,54 @@ protected override void WriteExtensionsFooter( GeneratorContext ctx, FileDescrip
...
         protected override void WriteServiceMethod( GeneratorContext ctx, MethodDescriptorProto method, ref object state )
         {
-            var outputType = MakeRelativeName( ctx, method.OutputType );
-            var inputType = MakeRelativeName( ctx, method.InputType );
-
-            ctx.WriteLine( $"{Escape( outputType )} {Escape( method.Name )}({Escape( inputType )} request);" );
+            ctx.WriteLine( $"public AsyncJob<SteamUnifiedMessages.ServiceMsg<{Escape( method.OutputType[1..] )}>> {Escape( method.Name )}({Escape( MakeRelativeName( ctx, method.InputType ) )} request)" )
+                .WriteLine( "{" )
+                .Indent()
+                .WriteLine( $"return UnifiedMessages.SendMessage<{Escape( MakeRelativeName( ctx, method.InputType ) )}, {Escape( method.OutputType[1..] )}>( $\"{{SERVICE_NAME}}.{Escape( method.Name )}#1\", request );" )
```
xPaw, line 79: I'd remove {SERVICE_NAME} and just make it a const string.

affederaffe:
> I like the idea. If you enable the trimming warnings, what warnings are reported? In particular handling of incoming messages.
> 
> Do note that protobuf.net is heavily reflection based, so that doesn't help us with AOT/trimming atm.

There seem to be no more trimming warnings aside from protobuf.net usage. I'll try to see how to avoid these too soon :tm:

xPaw:
@yaakov-h was also looking at that, and arrived at the conclusion that switching to google.protobuf is probably the way to go.

xPaw:
I don't however like calling CanHandleMsg for every single message, maybe registration and a dictionary lookup would work better.

See 807fa510b9a9b8b8aa5b8d79d436acccce3d6fb7 for previous attempt at that.

affederaffe:
> I don't however like calling CanHandleMsg for every single interface, maybe registration and a dictionary lookup would work better.
> 
> See [807fa51](https://github.com/SteamRE/SteamKit/commit/807fa510b9a9b8b8aa5b8d79d436acccce3d6fb7) for previous attempt at that.

It's not really every interface, only for currently created services (until they are disposed that is).
A dictionary could actually be worse for memory usage I imagine, but that'd need testing

xPaw:
Right I see it only goes through the registered services. But it calls CanHandleMsg for every service so it does the parsing for every one of these.

The handlers list could be a dictionary of <service name, handler>, so that you can parse the method name, and directly check if a handler is registered for a particular service.

yaakov-h:
This is a very different approach to the one I was looking at, and I do like it.

Keep in mind that consumers should be able to generate and run additional services (e.g. from updated protobufs) so some of those `internal` methods would need to be `public`.

Regarding the discussion above, perhaps each service should expose its name as a public property, then we can parse the service name once and invoke any matching handlers?

However, yes, for full trimming support and AOT we will almost certainly need to switch to Google.Protobuf. I haven't spiked AOT w/ Google just yet, but even manually hand-writing serializers and deserializers with protobuf-net.Core triggered trim and AOT warnings because it is annoyingly dependent on dynamic coding APIs.

xPaw:
> perhaps each service should expose its name as a public property

Could add service name to `UnifiedService`, and then have `HandleMsg( method, version, packet )` and then only call HandleMsg if the service name matches.

xPaw:
FYI this PR needs to rebase the protobufs submodule because its using an older commit.

affederaffe:
> This is a very different approach to the one I was looking at, and I do like it.
> 
> Keep in mind that consumers should be able to generate and run additional services (e.g. from updated protobufs) so some of those `internal` methods would need to be `public`.
> 
> Regarding the discussion above, perhaps each service should expose its name as a public property, then we can parse the service name once and invoke any matching handlers?
> 
> However, yes, for full trimming support and AOT we will almost certainly need to switch to Google.Protobuf. I haven't spiked AOT w/ Google just yet, but even manually hand-writing serializers and deserializers with protobuf-net.Core triggered trim and AOT warnings because it is annoyingly dependent on dynamic coding APIs.

From what I could gather, there currently is no satisfying solution:
1. Google.Protobuf with protoc ignores services completely, maybe we could modify the c# generator plugin
2. Build an AOT-compatible generator on top of protobuf-net's `CommonCodeGenerator`/`CSharpCodeGenerator`
2.1. For Google.Protobuf: would be completely AOT-compatible, although slightly cursed
2.2. For protobuf-net: less cursed, though even the protobuf-net.Core library seems to use reflection to some extend
3. (A completely home-brew solution?)

This is, however, out of scope of what I currently have time for.
Per https://protobuf-net.github.io/protobuf-net/3_0, there seems to be some intention to create a source generator, but I wouldn't bet it releases any time soon

Resources/ProtobufGen/ProtobufGen/SteamKitCSharpCodeGenerator.cs
```diff
@@ -37,20 +32,50 @@ protected override void WriteExtensionsFooter( GeneratorContext ctx, FileDescrip
...
         protected override void WriteServiceMethod( GeneratorContext ctx, MethodDescriptorProto method, ref object state )
         {
-            var outputType = MakeRelativeName( ctx, method.OutputType );
-            var inputType = MakeRelativeName( ctx, method.InputType );
-
-            ctx.WriteLine( $"{Escape( outputType )} {Escape( method.Name )}({Escape( inputType )} request);" );
+            ctx.WriteLine( $"public AsyncJob<SteamUnifiedMessages.ServiceMsg<{Escape( method.OutputType[1..] )}>> {Escape( method.Name )}({Escape( MakeRelativeName( ctx, method.InputType ) )} request)" )
+                .WriteLine( "{" )
+                .Indent()
+                .WriteLine( $"return UnifiedMessages.SendMessage<{Escape( MakeRelativeName( ctx, method.InputType ) )}, {Escape( method.OutputType[1..] )}>( $\"{state as string}.{Escape( method.Name )}#1\", request );" )
```
xPaw, line 75:
```suggestion
                .WriteLine( $"return UnifiedMessages.SendMessage<{Escape( MakeRelativeName( ctx, method.InputType ) )}, {Escape( method.OutputType[1..] )}>( \"{state as string}.{Escape( method.Name )}#1\", request );" )
```

SteamKit2/SteamKit2/Steam/Handlers/SteamUnifiedMessages/SteamUnifiedMessages.cs
```diff
@@ -158,101 +71,65 @@ public AsyncJob<ServiceMethodResponse> SendMessage<TRequest>( string name, TRequ
...
-                    Client.PostCallback( callback );
-                }
+        internal void HandleServiceMsg<TService>( IPacketMsg packetMsg )
+            where TService : IExtensible, new()
+        {
+            var callback = new ServiceMsg<TService>( (packetMsg as PacketClientMsgProtobuf)! );
+            Client.PostCallback( callback );
+        }
+
+        public abstract class UnifiedService : IDisposable
```
xPaw, line 120: need xmldocs. I think these HandleMsg and ServiceName likely need to be public for consumers to be able to implement their own service handlers (e.g. try generating a service proto outside of SteamKit project)

xPaw:
Fix the build error on ci please.

xPaw:
Breaking change for consumers:

old:
```csharp
SteamUnifiedMessages.UnifiedService<ISomething> service = unifiedMessages.CreateService<ISomething>();
SteamUnifiedMessages.ServiceMethodResponse job = await service.SendMessage(x => x.DoThing(msg));
SomeResponse response = job.GetDeserializedResponse<SomeResponse>();
```

new:
```csharp
Something service = unifiedMessages.CreateService<Something>();
SteamUnifiedMessages.ServiceMsg<SomeResponse> job = await service.DoThing(msg);
SomeResponse response = job.PacketResult;
```

`PacketResult` is an odd name, and I think `ServiceMsg` also needs some better name for responses.

SteamKit2/SteamKit2/Steam/Handlers/SteamUnifiedMessages/SteamUnifiedMessages.cs
```diff
@@ -19,134 +14,52 @@ namespace SteamKit2
...
         /// The returned <see cref="AsyncJob{T}"/> can also be awaited to retrieve the callback result.
         /// </summary>
         /// <typeparam name="TRequest">The type of a protobuf object.</typeparam>
         /// <param name="name">Name of the RPC endpoint. Takes the format ServiceName.RpcName</param>
         /// <param name="message">The message to send.</param>
-        /// <returns>The JobID of the request. This can be used to find the appropriate <see cref="ServiceMethodResponse"/>.</returns>
-        public AsyncJob<ServiceMethodResponse> SendMessage<TRequest>( string name, TRequest message )
-            where TRequest : IExtensible, new()
+        /// <returns>The JobID of the request. This can be used to find the appropriate <see cref="ServiceMsg{TResult}"/>.</returns>
+        public AsyncJob<ServiceMsg<TResult>> SendMessage<TRequest, TResult>( string name, TRequest message )
```
xPaw, line 44: TResult needs xmldoc

xPaw:
I just noticed that `SendNotification` is not being generated, for example `RemoteClient.NotifyOnline#1`. I *think* we can do it when response type is `NoResponse`.

affederaffe:
> One slight worry I have is that services directly return the service, which is an extension of `UnifiedService` instead of being a wrapper like `UnifiedService<Something>`

I think this would be rather complicated to achieve since we'd need some form of mapping between the interface and implementation.
Since there will be another breaking change when switching to Google.Protobuf anyway and I don't think the public api of the generated classes will change until then, this is good enough.

SteamKit2/SteamKit2/Base/Generated/SteamMsgRemoteClient.cs
```diff
@@ -9,33 +9,196 @@
...
+        public override string ServiceName { get; } = "RemoteClient";
+
+        public AsyncJob<SteamUnifiedMessages.ServiceMsg<CRemoteClient_GetPairingInfo_Response>> GetPairingInfo(CRemoteClient_GetPairingInfo_Request request)
+        {
+            return UnifiedMessages.SendMessage<CRemoteClient_GetPairingInfo_Request, CRemoteClient_GetPairingInfo_Response>( "RemoteClient.GetPairingInfo#1", request );
+        }
+
+        public AsyncJob<SteamUnifiedMessages.ServiceMsg<NoResponse>> NotifyOnline(CRemoteClient_Online_Notification request)
+        {
+            return UnifiedMessages.SendMessage<CRemoteClient_Online_Notification, NoResponse>( "RemoteClient.NotifyOnline#1", request );
```
affederaffe, line 23: Correct me if I misunderstood you, but this request is generated no?

xPaw: It needs to call SendNotification

xPaw:
013_UnifiedMessages sample needs updating.

SteamKit2/SteamKit2/Steam/Handlers/SteamUnifiedMessages/Callbacks.cs
```diff
@@ -3,114 +3,33 @@
...
 namespace SteamKit2
 {
-    public partial class SteamUnifiedMessages : ClientMsgHandler
+    public partial class SteamUnifiedMessages
     {
         /// <summary>
         /// This callback is returned in response to a service method sent through <see cref="SteamUnifiedMessages"/>.
         /// </summary>
-        public class ServiceMethodResponse : CallbackMsg
+        public class ServiceMsg<TResult> : CallbackMsg where TResult : IExtensible, new()
```
xPaw, line 15:
```suggestion
        public class ServiceMethodResponse<TResult> : CallbackMsg where TResult : IExtensible, new()
```

Let's keep the same class name.

SteamKit2/SteamKit2/Steam/Handlers/SteamUnifiedMessages/Callbacks.cs
```diff
@@ -3,114 +3,33 @@
...
-                get { return MethodName[ ( ServiceName.Length + 1 ).. ].Split( '#' )[ 0 ]; }
-            }
-
-            /// <summary>
-            /// Gets the full name of the service method.
-            /// </summary>
-            public string MethodName { get; private set; }
-
-            private PacketClientMsgProtobuf PacketMsg;
+            public TResult MessageBody { get; private set; }
```
xPaw, line 25:
```suggestion
            public TResult Body { get; private set; }
```

we use `Body` in `ClientMsgProtobuf<TBody>`

xPaw (approved):
Good stuff.

EDIT: The sample needs fixing after the rename.

LossyDragon:
I do have a question about how this PR will work on Unified Service methods that we wouldn't necessarily post a request first. 

Such as FriendMessagesClient.IncomingMessage (NotifyAckMessageEcho, Etc). When the client is logged in using the NewSteamChat, what would I subcribe to to listen to any unified service methods that are send to the client?

xPaw:
`CreateService<FriendMessagesClient>` calling would do that (but you have to keep a ref so it doesn't dispose), maybe we need a create-and-forget register method that doesn't return a disposable? but then that also requires an unregister call.

LossyDragon:
My C# skills aren't the best, so I wouldn't know of the best outcome for that. I maintain the JavaSteam port and the Unified Messaging implementation there roughly follows the same concept. 

Currently, SK can handle the two 'ServiceMethods' using the same callback func style as other handlers.
```c#
// ...
manager.Subscribe<SteamUnifiedMessages.ServiceMethodResponse>( OnMethodResponse );
manager.Subscribe<SteamUnifiedMessages.ServiceMethodNotification>( OnMethodNotification );
//... 

// Exmple get msg history, this results in OnMethodResponse
// var friendMsgs = steamUnifiedMessages.CreateService<IFriendMessages>();
// var req = new CFriendMessages_GetRecentMessages_Request
// {
//     steamid1 = steamClient.SteamID.ConvertToUInt64(),
//     steamid2 = friendID,
//     count = 10,
// };
// friendMsgs.SendMessage( x => x.GetRecentMessages( req ) );

void OnMethodResponse( SteamUnifiedMessages.ServiceMethodResponse callback )
{
    Console.WriteLine( "ServiceMethodResponse: " + callback.MethodName );
    if( callback.RpcName == "GetRecentMessages" )
    {
        var incomingRecentMsgs = callback.GetDeserializedResponse<CFriendMessages_GetRecentMessages_Response>();
        Console.WriteLine("Msg Count: " + incomingRecentMsgs.messages.Count );
        // etc.
    }
}

void OnMethodNotification( SteamUnifiedMessages.ServiceMethodNotification callback )
{
    Console.WriteLine( "OnMethodNotification: " + callback.MethodName);
    if(callback.RpcName == "IncomingMessage" )
    {
    // LogonDetails.ChatMode = SteamUser.ChatMode.NewSteamChat will allow this method to be sent if a friend starts typing or send a msg.
        var incomingMsg = ( CFriendMessages_IncomingMessage_Notification )callback.Body;
        Console.WriteLine("Entry type" + incomingMsg.chat_entry_type );
        // etc
    }
}
```

With this PR, this is where I seem to get lost. Either by me confusing myself, or there's something missing to handle the messages above. 

It seems now ServiceMethodResponse and ServiceMethod (EMsg headers) are combined to just ServiceMethodResponse<TResult>. But I cant easily subscribe to that without it erroring for a Type argument.

SteamKit2/SteamKit2/Steam/Handlers/SteamUnifiedMessages/SteamUnifiedMessages.cs
```diff
@@ -158,101 +72,79 @@ public AsyncJob<ServiceMethodResponse> SendMessage<TRequest>( string name, TRequ
...
-
-
-        /// <summary>
-        /// Handles a client message. This should not be called directly.
-        /// </summary>
-        /// <param name="packetMsg">The packet message that contains the data.</param>
+        /// <inheritdoc />
         public override void HandleMsg( IPacketMsg packetMsg )
         {
-            switch ( packetMsg.MsgType )
```
xPaw, line 192: This EMsg handling needs to stay, and not solely rely on target_job_name.

`ServiceMethodNotification` needs to stay, because Steam server can send clients notifications too.

xPaw (changes requested):
See comment.

We probably should add (Un)RegisterService that doesn't return a disposable service, because consumers may want to handle incoming notifications, and this is a breaking change which requires a handler to be registered.

FriendMessagesClient.IncomingMessage in new chat mode would be a good test for this.

SteamKit2/SteamKit2/Steam/SteamClient/CallbackMgr/CallbackMsg.cs
```diff
@@ -6,22 +6,10 @@
 
 namespace SteamKit2
 {
-    /// <summary>
-    /// A callback message
-    /// </summary>
-    public interface ICallbackMsg
```
xPaw, line 12: I don't see why this interface was removed, unrelated to the changes in this PR?

LossyDragon:
I removed my previous post as I was wrong about how the proposed implementation works and as I was writing up a new response,  I saw xPaw pointed out the issues to be addressed. Thanks 😄 

This brings up another question I have. Since ServiceMethodResponse and ServiceMethod(Notification), are technically consumers right now. Would the ServiceMethodResponse<T> be eliminated as a consumer; since the example shows we await for that resp,  and ServiceMethodNotification stays as one?

xPaw:
No it should stay if people want to directly handle incoming messages (I am not entirely sure if Steam can send you a response not a notification without asking for one), this still matches the callbackmanager / asyncjobs design.

xPaw:
@affederaffe did you see my comment in regards to the incoming messages/notifications?

affederaffe:
> @affederaffe did you see my comment in regards to the incoming messages/notifications?

Yes, sorry, time is currently sparse.
I'll implement a few ideas and see what works best.

SteamKit2/SteamKit2/Steam/SteamClient/CallbackMgr/CallbackMsg.cs
affederaffe: This is opinionated, but either the interface or the abstract class should be removed.
CallbackMsg is the sole inheritor of ICallbackMsg and just adds further complexity as a user won't know which one to use as they have the same purpose.

SteamKit2/SteamKit2/Steam/Handlers/SteamUnifiedMessages/SteamUnifiedMessages.cs
```diff
@@ -103,12 +105,27 @@ public override void HandleMsg( IPacketMsg packetMsg )
...
+                    var methodName = jobName[ ( dot + 1 )..hash ].ToString();
+                    handler.HandleMsg( methodName, packetMsg );
+                    break;
+                }
+                case EMsg.ServiceMethod:
+                {
+                    if ( !int.TryParse( jobName[ ( hash + 1 ).. ], NumberFormatInfo.InvariantInfo, out var version ) )
+                        return;
+                    var methodName = jobName[ ( dot + 1 )..hash ].ToString();
+                    var notification = new ServiceMethodNotification( packetMsgProto, serviceName, methodName, version );
```
xPaw, line 124: So i'm thinking we can still generate typed messages with HandleMsg, but it would have to construct from the request not the response body (since it's NoResponse).

SteamKit2/SteamKit2/Steam/SteamClient/CallbackMgr/CallbackMsg.cs
xPaw: I understand that, but that seems outside of the scope of this PR. Want to make a separate one just for this change?

SteamKit2/SteamKit2/Steam/Handlers/SteamUnifiedMessages/SteamUnifiedMessages.cs
xPaw: We should avoid trying to parse the job name before verifying the EMsg imo.

xPaw:
Just a thought I had, calling `CreateService` multiple times like in d9f6b1b1d38d51823ca77f8e51648cafb2ce264a might be problematic, as it returns a disposable service, and the handler is being added by the name.

Wouldn't it break if you called CreateService multiple times?

JustArchi:
This piece of code is not working (callback function handler not being called despite event arriving):

```csharp	CallbackManager.Subscribe<SteamUnifiedMessages.ServiceMethodNotification<CChatRoom_IncomingChatMessage_Notification>>(OnIncomingChatMessage);
		CallbackManager.Subscribe<SteamUnifiedMessages.ServiceMethodNotification<CFriendMessages_IncomingMessage_Notification>>(OnIncomingMessage);
```

Previously I had this and it worked properly:

```csharp
CallbackManager.Subscribe<SteamUnifiedMessages.ServiceMethodNotification>(OnServiceMethod);

private async void OnServiceMethod(SteamUnifiedMessages.ServiceMethodNotification notification) {
		ArgumentNullException.ThrowIfNull(notification);

		switch (notification.MethodName) {
			case "ChatRoomClient.NotifyIncomingChatMessage#1":
				await OnIncomingChatMessage((CChatRoom_IncomingChatMessage_Notification) notification.Body).ConfigureAwait(false);

				break;
			case "FriendMessagesClient.IncomingMessage#1":
				await OnIncomingMessage((CFriendMessages_IncomingMessage_Notification) notification.Body).ConfigureAwait(false);

				break;
		}
	}

```

My experimental code lies in https://github.com/JustArchiNET/ArchiSteamFarm/tree/unified-experiments, the upgrade commit itself is https://github.com/JustArchiNET/ArchiSteamFarm/commit/75fd1b867511419fa11d224042f5bc761571ac8d - if I did something wrong, let me know and I can give it another try.

affederaffe:
> Just a thought I had, calling `CreateService` multiple times like in [d9f6b1b](https://github.com/SteamRE/SteamKit/commit/d9f6b1b1d38d51823ca77f8e51648cafb2ce264a) might be problematic, as it returns a disposable service, and the handler is being added by the name.
> 
> Wouldn't it break if you called CreateService multiple times? Maybe to simplify services should be non disposable and it should return an existing one (handlers.GetOrAdd basically).

I agree that using a single instance is probably the way to go, but it should still be disposable in order to be removed from `_handlers`.

xPaw:
> but it should still be disposable

This could be achieved by adding a separate call to remove it. I think that's more explicit than leaving it disposable (because if you call CreateService for some temporary call, disposing it would be dispose it for everything).

affederaffe:
> This piece of code is not working (callback function handler not being called despite event arriving):
> 
> ```cs
> CallbackManager.Subscribe<SteamUnifiedMessages.ServiceMethodNotification<CChatRoom_IncomingChatMessage_Notification>>(OnIncomingChatMessage);
> 
> CallbackManager.Subscribe<SteamUnifiedMessages.ServiceMethodNotification<CFriendMessages_IncomingMessage_Notification>>(OnIncomingMessage);
> ```
> 
> Previously I had this and it worked properly:
> 
> ```cs
> CallbackManager.Subscribe<SteamUnifiedMessages.ServiceMethodNotification>(OnServiceMethod);
> 
> private async void OnServiceMethod(SteamUnifiedMessages.ServiceMethodNotification notification) {
> 		ArgumentNullException.ThrowIfNull(notification);
> 
> 		switch (notification.MethodName) {
> 			case "ChatRoomClient.NotifyIncomingChatMessage#1":
> 				await OnIncomingChatMessage((CChatRoom_IncomingChatMessage_Notification) notification.Body).ConfigureAwait(false);
> 
> 				break;
> 			case "FriendMessagesClient.IncomingMessage#1":
> 				await OnIncomingMessage((CFriendMessages_IncomingMessage_Notification) notification.Body).ConfigureAwait(false);
> 
> 				break;
> 		}
> 	}
> ```
> 
> My experimental code lies in https://github.com/JustArchiNET/ArchiSteamFarm/tree/unified-experiments, the diff with main is [JustArchiNET/ArchiSteamFarm@main...unified-experiments](https://github.com/JustArchiNET/ArchiSteamFarm/compare/main...unified-experiments). You can also get artifacts for testing in https://github.com/JustArchiNET/ArchiSteamFarm/actions/workflows/publish.yml?query=branch%3Aunified-experiments. If I did something wrong, let me know and I can give it another try.

You need to create the service the notification is coming from first with `SteamUnifiedMessages.CreateService` as it is used to map the notification method name to the result type internally.

xPaw:
> You need to create the service the notification is coming from first 

He does in ArchiHandler.

SteamKit2/SteamKit2/Steam/Handlers/SteamUnifiedMessages/SteamUnifiedMessages.cs
```diff
@@ -16,22 +14,31 @@ namespace SteamKit2
...
+            return (_handlers.GetOrAdd( service.ServiceName, service ) as TService)!;
+        }
+
+        /// <summary>
+        /// Removes a service so it no longer can be used to send messages or receive notifications.
+        /// </summary>
+        /// <typeparam name="TService">The type of the service to remove.</typeparam>
+        public void RemoveService<TService>() where TService : UnifiedService, new()
+        {
+            var serviceName = new TService().ServiceName;
```
xPaw, line 40: Can ServiceName be made static? I don't know if `TService.ServiceName` is possible.

affederaffe:
> > You need to create the service the notification is coming from first
> 
> He does in ArchiHandler.

`ChatRoomClient` and `FriendMessagesClient` are missing, which is why the message cannot be dispatched; there's no handler registered for it.

xPaw:
My bad, missed the "Client" part.

This is outside of the scope of this PR, maybe a future improvement: Subscribing to incoming notifications could be done through `CreateService` so that you couldn't mess this up.

Resources/ProtobufGen/ProtobufGen/SteamKitCSharpCodeGenerator.cs
```diff
@@ -37,20 +34,106 @@ protected override void WriteExtensionsFooter( GeneratorContext ctx, FileDescrip
...
 
         protected override void WriteServiceMethod( GeneratorContext ctx, MethodDescriptorProto method, ref object state )
         {
-            var outputType = MakeRelativeName( ctx, method.OutputType );
-            var inputType = MakeRelativeName( ctx, method.InputType );
-
-            ctx.WriteLine( $"{Escape( outputType )} {Escape( method.Name )}({Escape( inputType )} request);" );
+            if ( method.OutputType.AsSpan()[ 1.. ] is "NoResponse" )
+            {
+                ctx.WriteLine( $"public void {Escape( method.Name )}({Escape( MakeRelativeName( ctx, method.InputType ) )} request)" )
```
xPaw, line 119:
```suggestion
                ctx.WriteLine( $"public void {Escape( method.Name )}({Escape( MakeRelativeName( ctx, method.InputType ) )} request )" )
```

Resources/ProtobufGen/ProtobufGen/SteamKitCSharpCodeGenerator.cs
```diff
@@ -37,20 +34,106 @@ protected override void WriteExtensionsFooter( GeneratorContext ctx, FileDescrip
...
+                    .WriteLine( "{" )
+                    .Indent()
+                    .WriteLine( $"UnifiedMessages.SendNotification<{Escape( MakeRelativeName( ctx, method.InputType ) )}>( \"{Escape( state as string )}.{Escape( method.Name )}#1\", request );" )
+                    .Outdent()
+                    .WriteLine( "}" )
+                    .WriteLine();
+            }
+            else
+            {
+                ctx.WriteLine( $"public AsyncJob<SteamUnifiedMessages.ServiceMethodResponse<{Escape( method.OutputType[ 1.. ] )}>> {Escape( method.Name )}({Escape( MakeRelativeName( ctx, method.InputType ) )} request)" )
```
xPaw, line 129:
```suggestion
                ctx.WriteLine( $"public AsyncJob<SteamUnifiedMessages.ServiceMethodResponse<{Escape( method.OutputType[ 1.. ] )}>> {Escape( method.Name )}( {Escape( MakeRelativeName( ctx, method.InputType ) )} request )" )
```

xPaw:
One solution I've come up to solve the handler not being registered is this:
1. `mgr.Subscribe<ChatRoomClient, CChatRoom_IncomingChatMessage_Notification>(func)`
2. `ChatRoomClient.Subscribe<CChatRoom_IncomingChatMessage_Notification>(mgr, func)`

The first option makes sense because the callbacks still go through the manager. The second option makes sense because you are subscribing on a specific service (but you still have to give it a manager).

However this still doesn't prevent users from shooting themselves in the foot because you still register <WrongService, DifferentMethod>, as that would have to be validated somehow (generate another method to check? but that seems overkill, maybe at that point its not our problem anymore).

And `mgr.Subscribe` would have to disallow subscribing directly to ServiceMethodNotification<T>.

Thoughts? Any different ideas?

JustArchi:
> > > You need to create the service the notification is coming from first
> > 
> > 
> > He does in ArchiHandler.
> 
> `ChatRoomClient` and `FriendMessagesClient` are missing, which is why the message cannot be dispatched; there's no handler registered for it.

Thanks, I can confirm that it works after adding required clients.

The only outstanding issue I see with this PR is that right now, it's possible to register for callback but skip registration of the service.

```
    mgr.Subscribe<ChatRoomClient, CChatRoom_IncomingChatMessage_Notification>(func)
    ChatRoomClient.Subscribe<CChatRoom_IncomingChatMessage_Notification>(mgr, func)
```

As @xPaw suggested above, any of those two would fix it. If you asked me, I like first one slightly more.

affederaffe:
> One solution I've come up to solve the handler not being registered is this:
> 
>     1. `mgr.Subscribe<ChatRoomClient, CChatRoom_IncomingChatMessage_Notification>(func)`
> 
>     2. `ChatRoomClient.Subscribe<CChatRoom_IncomingChatMessage_Notification>(mgr, func)`
> 
> 
> The first option makes sense because the callbacks still go through the manager, and probably preferrable to keep the logic in same place. It probably needs a separate method name, like SubscribeServiceMethodResponse and ServiceMethodNotification. The second option makes sense because you are subscribing on a specific service (but you still have to give it a manager).
> 
> However this still doesn't prevent users from shooting themselves in the foot because you still can register <WrongService, DifferentMethod>, as that would have to be validated somehow (generate another method to check? but that seems overkill, maybe at that point its not our problem anymore).
> 
> And `mgr.Subscribe` would have to disallow subscribing directly to ServiceMethodNotification.
> 
> Thoughts? Any different ideas?
> 
> P.S. See my comment above about adding a sample for receiving notifications.

The approach to subscribe to the manager directly seems more in line with what users currently do and therefore what is probably expected.
Additionally mapping types to method return types seems overkill to me, imo just documenting the behaviour is enough.
Disallowing `ServiceMethodNotification<T>` or `ServiceMethodResponse<T>` should be doable with type constraints.

I'll try to implement this soon and add/update the samples

voided (commented):
This looks really good!

I agree with having that one-line approach of subscribing to notifications via the callback manager - infinitely less foot-guns than needing to remember to create a service.

I'll defer to @xPaw for the final approval, feel free to consider my comments nit-picky. :)

SteamKit2/SteamKit2/Base/Generated/SteamMsgWorkshop.cs
```diff
@@ -77,9 +77,31 @@ public bool needs_action
...
     }
 
-    public interface IWorkshop
+    public class Workshop : SteamUnifiedMessages.IUnifiedService
     {
-        CWorkshop_GetEULAStatus_Response GetEULAStatus(CWorkshop_GetEULAStatus_Request request);
+        public static string ServiceName { get; } = "Workshop";
+
+        /// <inheritdoc />
+        public SteamUnifiedMessages UnifiedMessages { get; init; }
```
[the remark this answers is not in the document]
xPaw: What if a consumer wants to generate and use their own service?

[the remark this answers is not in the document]
voided: Good point. I think an explicit interface implementation can solve for that: https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/interfaces/explicit-interface-implementation

It would hide access to the property on the implementation types and only allow access if you cast to a `IUnifiedService`, which the `SteamUnifiedMessages` handler can easily do.

A consumer could end up doing the same, but it would at least make it more unlikely.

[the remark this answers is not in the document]
xPaw: Yeah it would be good to get rid of this property from the generated classes, and only leave it on the parent.

JustArchi (approved):
Works for me, summary of all changes I did: <https://github.com/JustArchiNET/ArchiSteamFarm/compare/main...unified-experiments>

xPaw:
@voided's comments were caused by making servicename a static, I reverted that. Squashed all the commits with minor cleanup and merged in 2ba0a55f71c688f1f778b2dd4f1f8229506c1fa4

Thanks!

PR #1441, opened 2024-09-30
NHA2: Decode CS2 store data

yaakov-h:
- Adds support for LZMA-decoded Binary KV to NHA2
- Adds support for 0x0B (AlternateEnd) to KeyValue
- Allow decoding of KV, KV-LZMA, and Protobuf data even above the 400 byte limit. (Performance of the application is not guaranteed.)

This needs tests but PR is up for people to build themselves or use GHA artefacts.

Addresses the issue described in discussion #1440.

The payload is a binary blob with the four-byte magic `LZMA`. This doe not seem to be a standard but is possibly a Valve creation, as seen elsewhere in ValveResourceFormat.

It also uses a different type of KeyValues binary serialisation, using `AlternateEnd`from VBKV as seen here: https://github.com/ValveResourceFormat/ValveKeyValue/issues/9

These changes probably need some new tests for the modified KV logic, and ideally also for the new LzmaUtil class.


SteamKit2/SteamKit2/Util/LzmaUtil.cs
```diff
@@ -0,0 +1,125 @@
...
+using System.Buffers.Binary;
+using System.Diagnostics.CodeAnalysis;
+using System.IO;
+
+namespace SteamKit2.Internal
+{
+    /// <summary>
+    /// Internal helper to decode data blobs with the 'LZMA' magic header.
+    /// </summary>
+    public static class LzmaUtil
```
xPaw, line 11: I would avoid exposing this publicly.

yaakov-h: I'd need InternalsVisibleTo for NHA2 then, or to move the whole thing into NHA2.

If GC are using it at the protocol level somehow then it would be needed by SK2 clients.
If they're not and its game-specific then I'd be on the fence about including it.

xPaw: Just put it in NH.

yaakov-h: Then I need a second copy of the LZMA SDK to get at the `Decoder`. 😞

xPaw: That's annoying. InternalsVisibleTo then? :(

yaakov-h: yeah, seems like the least evil option for now

xPaw (approved):

PR #1447, opened 2024-10-12
Fix `GetCDNAuthToken` after Steam change

oureveryday:
Move the GetCDNAuthToken from SteamApps to SteamContent

### DIFF
--- a/SteamKit2/SteamKit2/Steam/Handlers/SteamApps/Callbacks.cs
+++ b/SteamKit2/SteamKit2/Steam/Handlers/SteamApps/Callbacks.cs
@@ -1,4 +1,4 @@
-﻿/*
+/*
  * This file is subject to the terms and conditions defined in
  * file 'license.txt', which is part of this source code package.
  */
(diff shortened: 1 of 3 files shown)


xPaw:
Result code still needs to be checked/available

oureveryday:
I didn't see result code in CContentServerDirectory_GetCDNAuthToken_Response

xPaw:
Service response itself should have a result. You should try testing it on invalid hostname / depotid, I would expect that to give a bad eresult.

SteamKit2/SteamKit2/Steam/Handlers/SteamApps/Callbacks.cs
```diff
@@ -1,4 +1,4 @@
-﻿/*
+/*
```
xPaw, line 1: can you undo these changes please (utf BOM probably)

oureveryday: Fixed that

oureveryday:
> Service response itself should have a result. You should try testing it on invalid hostname / depotid, I would expect that to give a bad eresult.

If a invalid hostname / depotid passed, it will return null result.

xPaw:
> it will return null result.

Yeah but what's the `message.Result`?

oureveryday:
> Yeah but what's the `message.Result`?

Don't know where it is...

oureveryday:
Fixed that.

xPaw:
xmldoc in cdn client needs updating, see the failing build

PR #1452, opened 2024-10-15
Build functionality for the server list to refresh itself

xPaw:
Server list providers now return the last time they were written (I had to add a bool to ReplaceList to avoid rewriting the file if its using provider list).

By default it will consider server list newer than 7 days as fresh. If it fails to refresh the list from API, it will use the previous logic of using the provider.

cc @justarchi review the logic please

### DIFF
--- a/SteamKit2/SteamKit2/Steam/Discovery/IServerListProvider.cs
+++ b/SteamKit2/SteamKit2/Steam/Discovery/IServerListProvider.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Threading.Tasks;
 
 namespace SteamKit2.Discovery
@@ -8,6 +9,14 @@ namespace SteamKit2.Discovery
     /// </summary>
     public interface IServerListProvider
     {
+        /// <summary>
+        /// When the server list was last refreshed, used to determine if the server list should be refreshed from the Steam Directory
+        /// </summary>
+        /// <remarks>
+        /// This should return DateTime with the UTC kind
+        /// </remarks>
+        DateTime LastServerListRefresh { get; }
+
         /// <summary>
         /// Ask a provider to fetch any servers that it has available
         /// </summary>
--- a/SteamKit2/SteamKit2/Steam/Discovery/SmartCMServerList.cs
+++ b/SteamKit2/SteamKit2/Steam/Discovery/SmartCMServerList.cs
@@ -4,6 +4,7 @@ using System.Collections.ObjectModel;
 using System.Diagnostics;
 using System.Linq;
 using System.Net;
+using System.Net.Sockets;
 using System.Threading.Tasks;
 
 namespace SteamKit2.Discovery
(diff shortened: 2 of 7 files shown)


SteamKit2/SteamKit2/Steam/Discovery/SmartCMServerList.cs
```diff
@@ -91,45 +96,98 @@ private bool WaitForServersFetched()
...
+                    return;
+                }
             }
 
-            DebugWrite( "Resolved {0} servers", endpointList.Count );
-            ReplaceList( endpointList );
+            // This is a last effort to attempt any valid connection to Steam,
+            // however Steam client does not appear to do this anymore,
+            // and this will not work when Steam moves to only websocket servers
+            DebugWrite( $"Server list provider had no entries, {nameof( SteamDirectory )} failed, falling back to cm0" );
```
xPaw, line 176: I didn't check if this server still works, but I'm thinking this should be removed at some point.

SteamKit2/SteamKit2/Steam/Discovery/IServerListProvider.cs
```diff
@@ -8,6 +9,14 @@ namespace SteamKit2.Discovery
     /// </summary>
     public interface IServerListProvider
     {
+        /// <summary>
+        /// When the server list was last refreshed, used to determine if the server list should be refreshed from the Steam Directory
+        /// </summary>
+        /// <remarks>
+        /// This should return DateTime with the UTC kind
+        /// </remarks>
+        DateTime LastServerListRefresh { get; }
```
JustArchi, line 18: Breaking change, can't we make use of interfaces default implementation and make it `=> DateTime.MinValue` or something by default? I like how we force it from the customers, that's the way it should be done tbh, but we have alternative if needed.

SteamKit2/SteamKit2/Steam/Discovery/SmartCMServerList.cs
JustArchi: I'd consider hardcoding some general servers that *seem* like their name couldn't change.

It's better than nothing, even if the list gets outdated. It'd be fallback used only if fetching up-to-date info has failed. I'd rather thave 5 hardcoded servers than zero in such case. You could return them at random, never save, never do anything else.

JustArchi (commented):
Other than that LGTM

SteamKit2/SteamKit2/Steam/Discovery/IServerListProvider.cs
xPaw: Returning MinValue would change the behaviour (it would always try to refresh before using the provided list)

SteamKit2/SteamKit2/Steam/Discovery/SmartCMServerList.cs
xPaw: I added a public static string for default servers (had to make separate ones for WS / tcp)

JustArchi (approved):

SteamKit2/SteamKit2/Steam/Discovery/SmartCMServerList.cs
```diff
@@ -53,6 +54,22 @@ public SmartCMServerList( SteamConfiguration configuration )
             this.configuration = configuration ?? throw new ArgumentNullException( nameof( configuration ) );
         }
 
+        /// <summary>
+        /// The default fallback Websockets server to attempt connecting to if fetching server list through other means fails.
+        /// </summary>
+        /// <remarks>
+        /// If the default server set here no longer works, please create a pull request to update it.
+        /// </remarks>
+        public static string DefaultServerWebsocket { get; set; } = "cmp1-ord1.steamserver.net";
```
yaakov-h, line 63: Can this config be moved to `SteamConfiguration` or does it need to be accessed w/o any configuration?

xPaw: server list has access to the config. I can move it but I will keep it static, the only reason I'm exposing it is for specific cases when someone wants to test it without rebuilding SK or something.

SteamKit2/SteamKit2/Steam/Discovery/SmartCMServerList.cs
xPaw: And now after last Steam update they changed the servers:

cmp1-dfw1.steamserver.net:27018
cmp1-dfw1.steamserver.net:443
cmp1-iad1.steamserver.net:27018
cmp1-iad1.steamserver.net:443
cmp1-lax1.steamserver.net:27018
cmp1-lax1.steamserver.net:443
cmp1-sea1.steamserver.net:27018
cmp1-sea1.steamserver.net:443

PR #1454, opened 2024-10-18
HardwareUtils.cs: Fix Android support

TheKingFireS:
I ran DepotDownloader in Android using Termux and proot, after 2.7.3 update alongside with SteamKit2 update, it break Android support due to probably improper MAC address detection (Trying to get access to /sys/class/net but Android restricted it). I added nested to it. It passed without problem.

This is my first pull request ever, let me know if I'm doing it wrong.

### DIFF
--- a/SteamKit2/SteamKit2/Util/HardwareUtils.cs
+++ b/SteamKit2/SteamKit2/Util/HardwareUtils.cs
@@ -300,12 +300,19 @@ namespace SteamKit2
                         continue;
                     }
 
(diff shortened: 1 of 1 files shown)


xPaw:
You added a try catch in a try catch

TheKingFireS:
> You added a try catch in a try catch, it was fixed already

Oh, I didn't realize fix was already made after DepotDownloader 2.7.3, my bad. Good night :sweat_smile:

PR #1459, opened 2024-10-18
Make DepotManifest setters public

NicknineTheEagle:
This makes constructors and setters for DepotManifest and its children public. This allows programs using SteamKit to manually build and edit depot manifest instances.

### DIFF
--- a/SteamKit2/SteamKit2/Types/DepotManifest.cs
+++ b/SteamKit2/SteamKit2/Types/DepotManifest.cs
@@ -64,7 +64,10 @@ namespace SteamKit2
             {
             }
 
-            internal ChunkData( byte[] id, uint checksum, ulong offset, uint comp_length, uint uncomp_length )
+            /// <summary>
+            /// Initializes a new instance of the <see cref="ChunkData"/> class with specified values.
+            /// </summary>
+            public ChunkData( byte[] id, uint checksum, ulong offset, uint comp_length, uint uncomp_length )
             {
                 this.ChunkID = id;
                 this.Checksum = checksum;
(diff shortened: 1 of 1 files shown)


SteamKit2/SteamKit2/Types/DepotManifest.cs
```diff
@@ -83,36 +86,50 @@ public class FileData
...
+            public string LinkTarget { get; set; }
 
+            /// <summary>
+            /// Initializes a new instance of the <see cref="FileData"/> class.
+            /// </summary>
+            public FileData()
+            {
+                FileName = "";
+                FileNameHash = [];
+                Chunks = new List<ChunkData>();
```
xPaw, line 124:
```suggestion
                Chunks = [];
```

SteamKit2/SteamKit2/Types/DepotManifest.cs
```diff
@@ -83,36 +86,50 @@ public class FileData
...
             /// </summary>
-            public string LinkTarget { get; private set; }
+            public string LinkTarget { get; set; }
 
+            /// <summary>
+            /// Initializes a new instance of the <see cref="FileData"/> class.
+            /// </summary>
+            public FileData()
+            {
+                FileName = "";
```
xPaw, line 122:
```suggestion
                FileName = Use string.Empty;
```

SteamKit2/SteamKit2/Types/DepotManifest.cs
```diff
@@ -83,36 +86,50 @@ public class FileData
...
+                FileNameHash = [];
+                Chunks = new List<ChunkData>();
+                FileHash = [];
+                LinkTarget = "";
```
xPaw, line 126: Isn't it null by default in the protos?

NicknineTheEagle: Technically, all fields are null by default in the protos. This particular field is optional if that's what you mean, it's usually not set in official manifests. `DepotManifest::Serialize` checks IsNullOrWhiteSpace so it's safe to use empty string here.

xPaw: I would just avoid setting it

PR #1465, opened 2024-10-28
Cleanup callback manager internals

xPaw:
`Subscription` is just a pointless wrapper around `CallbackBase` which is disposable by itself, so I don't see why it can't be returned directly (the subscribes only return IDisposable).


xPaw:
CallbackBase also unregisters, the logic should have remained the same

yaakov-h (approved):

PR #1471, opened 2024-11-15
Add functions to return raw CDN manifests and chunks

NicknineTheEagle:
This PR splits DownloadManifestAsync and DownloadDepotChunkAsync functions into two with HTTP download process itself now being its own function. This lets apps using SteamKit to get raw CDN output buffer and do their own processing on it. The signature and output of existing functions remains unchanged so this shouldn't break compatibility.

### DIFF
--- a/SteamKit2/SteamKit2/Steam/CDN/Client.cs
+++ b/SteamKit2/SteamKit2/Steam/CDN/Client.cs
@@ -52,23 +52,20 @@ namespace SteamKit2.CDN
         }
 
         /// <summary>
-        /// Downloads the depot manifest specified by the given manifest ID, and optionally decrypts the manifest's filenames if the depot decryption key has been provided.
+        /// Downloads raw depot manifest specified by the given manifest ID.
         /// </summary>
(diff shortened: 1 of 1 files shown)


xPaw:
I'm not sure there's any reason to have this, if you want raw contents just download the the url using any http library yourself. That's effectively what splitting it into a separate method does.

Dimensional:
Probably, but it still means effectively having a whole separate http call from the one built in, from what I can tell. Then again, I'm an amateur, but having it split up at least provides extra options, if someone wants to download the files and save directly or have them processed and then saved, without having to copy the client.cs file and repurpose it.

xPaw:
There's nothing to copy besides constructing the url which is a couple of lines at most.

NicknineTheEagle:
I've looked everything over and you appear to be correct, when download function is called, the application already knows everything it needs to build CDN request URL. Having everything handled in SteamKit would make things easier on app side but if you believe this PR is redundant, I won't insist.

xPaw:
Gonna close then.

PR #1484, opened 2024-12-30
chore: GitHub should be spelled with a capital H

nolddor:
This is such a tiny minor issue I hesitate to even bother with it, but folks at GitHub tend to care about it, so let's use the correct naming.


nolddor:
Additionally, they also tend to capitalize `Discussions` as well.

PR #1486, opened 2025-01-04
Add support for subscribing to callbacks that return Task

### LINKED ISSUE #1464: CallbackMgr: Allow subscribing to callbacks that return Task
It would allow using `async Task` methods instead of having to use `async void`.

xPaw:
@yaakov-h what do you think will this work?

Fixes #1464


yaakov-h:
I don't like blocking waiting for the task, but since we want thread affinity.. 🤔 

Might be worth converting all callbacks to `ValueTask`.

xPaw:
> I don't like blocking waiting for the task

Non task callbacks already block, so not sure what difference that makes here.

> Might be worth converting all callbacks to ValueTask.

How would that look?

xPaw:
I don't see how ValueTask can be applied here, so I would stick with Task.

JustArchi (approved):
Tests done via https://github.com/JustArchiNET/ArchiSteamFarm/compare/sk2-test

No regressions observed, LGTM

PR #1501, opened 2025-02-02
Fix macOS incompatibility related to IOKit

JustArchi:
See https://github.com/Homebrew/homebrew-core/issues/206331#issuecomment-2629541596 for detailed reasoning and analysis.

If you could schedule a bugfix release with it (3.0.2), I'd appreciate that. Thanks in advance.

References:
- https://github.com/JustArchiNET/ArchiSteamFarm/issues/3381
- https://github.com/Homebrew/homebrew-core/issues/206331

### DIFF
--- a/SteamKit2/SteamKit2/Util/MacHelpers.cs
+++ b/SteamKit2/SteamKit2/Util/MacHelpers.cs
@@ -132,7 +132,7 @@ namespace SteamKit2.Util.MacHelpers
     [SupportedOSPlatform( "macos" )]
     static partial class IOKit
     {
-        const string LibraryName = "IOKit.framework/IOKit";
+        const string LibraryName = "/System/Library/Frameworks/IOKit.framework/IOKit";
 
         public const uint kIOMasterPortDefault = 0;
         public const string kIOPlatformSerialNumberKey = "IOPlatformSerialNumber";


JustArchi:
Added unit test and fixed another occurances of the same issue. Unit test was failing before, so it definitely proves that there was an issue.

Thanks in advance for merging.

yaakov-h:
By reverting the changes to the library paths and keeping the tests, I cannot get this to fail on my own MBP with macOS Sequoia 15.3 (24D60) with either .NET 8 or .NET 9 SDKs. 🤔

JustArchi:
I'm not an expert in macOS so I don't know why you can't reproduce it locally, but it definitely did fail in CI:
- https://github.com/SteamRE/SteamKit/actions/runs/13103559845/job/36554805608
- https://github.com/SteamRE/SteamKit/actions/runs/13103538944/job/36554764315

yaakov-h:
Yeah, I saw that. I still have Questions.

Will have a think about this - seems harmless atm but not sure what the wider impact is exactly.

yaakov-h:
ah, got it.

This only seems to affect .NET 9.0.1 / SDK 9.0.102 or higher. Iiiiiinteresting.

yaakov-h:
If we have global.json for 8.0.100 latestMinor then why is dlopen on GitHub Actions looking in `/Users/runner/.dotnet/shared/Microsoft.NETCore.App/9.0.1`?

xPaw:
https://github.com/SteamRE/SteamKit/commit/ad531d957e124038b4b056d56b3fd581d336f5fe

PR #1515, opened 2025-03-03
Trim password to 64 characters.

Alstruit:
Steam will only accept password length up to 64 characters.

Reference:
SteamRE/DepotDownloader#601
SteamRE/DepotDownloader#597

### DIFF
--- a/SteamKit2/SteamKit2/Steam/Authentication/SteamAuthentication.cs
+++ b/SteamKit2/SteamKit2/Steam/Authentication/SteamAuthentication.cs
@@ -139,6 +139,14 @@ namespace SteamKit2.Authentication
 
             // Encrypt the password
             var publicKey = await GetPasswordRSAPublicKeyAsync( details.Username! ).ConfigureAwait( false );
+
+            // Password limit is 64. If it's longer, trim it.
+            if (!string.IsNullOrEmpty(details.Password) && details.Password.Length > 64)
+            {
+                DebugLog.WriteLine("SteamAuthentication", "Notice: password is longer than 64 characters and will be trimmed.");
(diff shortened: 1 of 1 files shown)


SteamKit2/SteamKit2/Steam/Authentication/SteamAuthentication.cs
```diff
@@ -139,6 +139,14 @@ public async Task<CredentialsAuthSession> BeginAuthSessionViaCredentialsAsync( A
 
             // Encrypt the password
             var publicKey = await GetPasswordRSAPublicKeyAsync( details.Username! ).ConfigureAwait( false );
+
+            // Password limit is 64. If it's longer, trim it.
+            if (!string.IsNullOrEmpty(details.Password) && details.Password.Length > 64)
+            {
+                DebugLog.WriteLine("SteamAuthentication", "Notice: password is longer than 64 characters and will be trimmed.");
```
xPaw, line 146:
```suggestion
                DebugLog.WriteLine(nameof(SteamAuthentication), "Notice: password is longer than 64 characters and will be trimmed.");
```

Make a const variable for the "64".

Also, you didn't need to make a new pull request, you could have just pushed to the same branch.

xPaw:
I'm not really a fan of doing this in the library (this also misses this in SteamUser.LogOn btw), especially because without replacing non-ascii characters like Valve does this logic isn't entirely correct.

yaakov-h (commented):
I think we should just log the warning but not trim.

Any consumer can trim at any point, but if Valve change their behaviour I don't want consumers to need to upgrade before they can log back in.

Alstruit:
On the Unicode part. Steam does display an error if the password is not standard ASCII.

PR #1539, opened 2025-05-10
Add Deadlock protos

psychonic:
* Added Deadlock GC protos to protos.csv
* Generated code for Deadlock GC protos
* Added NHA specialization for Deadlock GC


xPaw:
See also #1426

psychonic:
That's a fair point. I didn't realize that other PR existed. (It also includes many protos that don't involve Steam traffic).

I would argue that there's no harm in including more until that potential future point in time when all are dropped.

psychonic:
Disregard that latest commit. I meant to include it in a separate branch. Removing shortly
'

xPaw:
I won't block this. I assume we can't really use shared generic Steam protos because they can be out of sync between games?

psychonic:
> I won't block this. I assume we can't really use shared generic Steam protos because they can be out of sync between games?

I'm not certain, but I think those base ones (steammessages_*) may be able to be shared. While they differ between games, I believe it's only due to them being synced from the steam code to the game's gcsdk at different times. In at least most cases, I would think the newer ones are backward compatible with old.

@voided may remember why that wasn't originally unified

psychonic:
Oh, I think I remember now. At the network level, they're compatible. But the game-specific proto classes reference the steam ones in some cases, and there can be compatibility issues there when using a different version, or if the generated class path isn't what it's expecting (although some hackery may be able to deal with the latter)

PR #1542, opened 2025-05-19
Post client disconnection callback before cancelling async jobs

xPaw:
After #1532, if you have some jobs running in a loop that gets cancelled on a disconnected callback, it is now possible to get into a loop where it will keep trying to send jobs, and instantly fail them.

Posting early should give consumers a better control over this (besides always checking IsConnected).


yaakov-h:
Is this because we cancel the job before giving consumers any indication as to why (i.e. that we are disconnected)?

PR #1544, opened 2025-05-31
Use `HttpClientFactory` for websocket connections, add `HttpClientPurpose`

xPaw:
This allows consumers to provide custom http clients for connecting to websockets.


JustArchi (approved):
LGTM

yaakov-h:
Do we always take ownership of and dispose of the HttpClient from the factory?

xPaw:
That seems to be the case, yes

PR #1573, opened 2025-08-30
Fix LinkTarget not being decrypted in DepotManifest

yretenai:
LinkTarget is not decrypted, despite it being an encrypted filename.

current:
Fields for File Data, showing that the Link Target is base64 encoded data

with fix:
Fields for File Data, showing that the Link Target is Versions/Current/HttpClient

### DIFF
--- a/SteamKit2/SteamKit2/Types/DepotManifest.cs
+++ b/SteamKit2/SteamKit2/Types/DepotManifest.cs
@@ -6,6 +6,7 @@
 using System;
 using System.Buffers;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.IO;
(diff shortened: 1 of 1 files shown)


yretenai:
The function is local method and uses the buffer allocated in the upper function. The variables are shared across the entire method. In other terms, it's essentially putting the entire stack as ref variables. The only repeated allocation is the stackalloc for the IV since stackalloced functions cannot move across function boundaries where the lifetime is undetermined.

xPaw:
Ah right I didn't notice it's local, will take a closer look.

yretenai:
Not sure what the actual performance cost is for such a small stackalloc, but I'll make the change just to reduce the function size.

PR #1580, opened 2025-09-17
Add vectorized Adler32

xPaw:
ImageSharp's license: https://github.com/SixLabors/ImageSharp/blob/3948fa86f5d77a2bd9d80292ad558404f13cc2f1/LICENSE
We should be falling under their Apache License, Version 2.0.

I added a public method in DepotChunk because DepotDownloader also uses adler32.

BenchmarkDotNet v0.15.2, Windows 11 (10.0.26100.6584/24H2/2024Update/HudsonValley)
AMD Ryzen 7 5800X3D 3.40GHz, 1 CPU, 16 logical and 8 physical cores
DefaultJob : .NET 10.0.0 (10.0.25.45207), X64 RyuJIT AVX2

| Method         | Mean        | Error     | StdDev    |
|--------------- |------------:|----------:|----------:|
| Avx2           |    17.00 us |  0.058 us |  0.054 us |
| Sse            |    28.22 us |  0.109 us |  0.091 us |
| Scalar         |   237.29 us |  0.150 us |  0.133 us |
| ScalarSteamKit | 2,396.87 us | 13.190 us | 11.014 us |

```csharp
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Running;
using SteamKit2;

namespace MyBenchmarks;

public class Adler32Bench
{
    private const int N = 1024 * 1024;
    private readonly byte[] data;

    public Adler32Bench()
    {
        data = new byte[ N ];
        new Random( 42 ).NextBytes( data );
    }

    [Benchmark]
    public uint Sse() => Adler32.CalculateSse( 0, data );

    [Benchmark]
    public uint Avx2() => Adler32.CalculateAvx2( 0, data );

    [Benchmark]
    public uint Scalar() => Adler32.CalculateScalar( 0, data );

    [Benchmark]
    public uint ScalarSteamKit() => Utils.AdlerHashOld( data );
}

public class Program
{
    public static void Main( string[] args )
    {
        var summary = BenchmarkRunner.Run<Adler32Bench>();
    }
}
```

### DIFF
--- a/SteamKit2/SteamKit2/Util/Adler32.cs
+++ b/SteamKit2/SteamKit2/Util/Adler32.cs
@@ -1,71 +1,440 @@
-/*
- * This file is subject to the terms and conditions defined in
- * file 'license.txt', which is part of this source code package.
- */
+// Copyright (c) Six Labors.
+// Licensed under the Six Labors Split License.
+// SteamKit copied from https://github.com/SixLabors/ImageSharp/blob/3948fa86f5d77a2bd9d80292ad558404f13cc2f1/src/ImageSharp/Compression/Zlib/Adler32.cs
 
 using System;
 using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using System.Runtime.Intrinsics;
+using System.Runtime.Intrinsics.Arm;
+using System.Runtime.Intrinsics.X86;
 
+#pragma warning disable IDE0007 // Use implicit type
 namespace SteamKit2;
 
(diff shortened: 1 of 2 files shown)


voided:
Does the Six Labors Split License have any downstream effects for end users using SteamKit commercially?

xPaw:
Good question, I'm not sure to be honest.

xPaw:
Internals are only exposed to tests in debug.

PR #1581, opened 2025-09-18
Set rollForward to major

xPaw:
Replaces #1274

Ref: https://learn.microsoft.com/en-us/dotnet/core/tools/global-json#rollforward

> Uses the latest patch level for the specified major, minor, and feature band.

if not found, it will try rolling forward. For CI builds it should use the specified version since that's what will be installed I think.

It's really annoying to just keep deleting global.json locally.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
     "version": "8.0.100",
-    "rollForward": "latestMinor"
+    "rollForward": "major"
   }
 }


yaakov-h:
why are you deleting the file? if you want to override the behaviour I'm pretty sure there's an env var for it.

xPaw:
Guess I will bump it to 10 when it's out.

PR #1591, opened 2025-10-15
fix: handle disposed socket on close attempt

### LINKED ISSUE #524: ObjectDisposedException for websocket
websocket

Sadly I don't have any more info regarding this, neither I can reproduce it myself. Perhaps stacktrace will be enough to guess what is wrong.

ezhevita:
Fixes #524.

### Some investigation
Originally I had the following stacktrace for ArchiSteamFarm:
```
System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'System.Net.WebSockets.ClientWebSocket'.
   at System.Net.WebSockets.ClientWebSocket.get_ConnectedWebSocket()
   at SteamKit2.WebSocketConnection.WebSocketContext.RunCore(HttpMessageInvoker invoker, TimeSpan connectionTimeout, CancellationToken cancellationToken)
   at SteamKit2.TaskExtensions.IgnoringCancellation(Task task, CancellationToken token)
```

`ConnectedWebSocket` is a private property and there is no mention of it in SteamKit codebase whatsoever, so there is definitely some method that got inlined and its stackframe was omitted.
`ClientWebSocket` accesses that property in 7 methods:
- `SendAsync` (+2 overloads),
- `ReceiveAsync` (+1 overload),
- `CloseAsync`,
- `CloseOutputAsync`.

`WebSocketContext` calls these methods in these cases:
- `SendAsync` is called from `WebSocketContext.SendAsync` <- `WebSocketConnection.Send`, which doesn't match our stacktrace,
- `ReceiveAsync` is called from `ReadMessageAsync`, but the `ObjectDisposedException` is explicitly handled there,
- `CloseAsync` seems to be the only suitable candidate.

### DIFF
--- a/SteamKit2/SteamKit2/Networking/Steam3/WebSocketContext.cs
+++ b/SteamKit2/SteamKit2/Networking/Steam3/WebSocketContext.cs
@@ -87,18 +87,14 @@ namespace SteamKit2
                     }
                 }
 
-                if (socket.State == WebSocketState.Open)
+                if (socket.State == WebSocketState.Open && disposed == 0)
                 {
                     connection.log.LogDebug( nameof(WebSocketContext), "Closing connection...");
                     try
                     {
                         await socket.CloseAsync(WebSocketCloseStatus.NormalClosure, null, default).ConfigureAwait(false);
                     }
-                    catch ( ObjectDisposedException ex )
-                    {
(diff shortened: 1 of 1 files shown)


JustArchi (approved):

JustArchi (commented):
Should be sufficient, you don't need to release semaphore to dispose it.

